PR #1, opened 2015-08-28
Update import path case.

shurcooL:
Followup to 91eb6c3e and d13eb8d7.

~~Also update go -> Go in fenced code block highlighting (both work, but this is the correct language name case).~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # go-autorest
 
-[![GoDoc](https://godoc.org/github.com/azure/go-autorest/autorest?status.png)](https://godoc.org/github.com/azure/go-autorest/autorest) [![Build Status](https://travis-ci.org/Azure/go-autorest.svg?branch=master)](https://travis-ci.org/Azure/go-autorest) [![Go Report Card](http://goreportcard.com/badge/Azure/go-autorest)](http://goreportcard.com/report/Azure/go-autorest)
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
Omg omg omg first pull request.

ahmetalpbalkan:
I am not sure about Go vs go. I've always used "go" and it served me well. Can you point us to a resource? Does github use pygments?

shurcooL:
> I am not sure about Go vs go. I've always used "go" and it served me well. Can you point us to a resource? Does github use pygments?

If you prefer to stay with `go`, that's not a problem, I will revert that change.

I don't have a source, I myself also started using `go` initially but then over time ended up using `Go` everywhere. It seemed nicer since the language is officially called Go. But I don't want to change it for you if you already use `go` everywhere else.

ahmetalpbalkan:
LGTM thanks.

shurcooL:
No problem! I've updated the commit to remove the go -> Go change and updated the commit message accordingly.

> Omg omg omg first pull request.

I know the feeling. Sorry it's just a tiny one, nothing substantial. But first PRs always tend to be small and grow from there. :)

shurcooL:
By the way, you'll also want to remember to update the repository description. It's currently still the lowercase version:

image

brendandixon:
@shurcooL Thanks.

shurcooL:
Happy to help!

PR #3, opened 2015-09-30
-- Added Logging inspectors to trace http.Request / Response

brendandixon:
- Added support for User-Agent header
- Changed WithHeader PrepareDecorator to use set vs. add
- Added JSON to error when unmarshalling fails
-Added Client#Send method
- Corrected case of "Azure" in package paths
- Added convert helpers, Azure helpers, and improved ease-of-use
- Corrected golint issues

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -1,6 +1,10 @@
 package autorest
 
 import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"log"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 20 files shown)


brendandixon:
@ahmetalpbalkan @paulmey I don't know if you two are interested, but this PR contains the final changes to go-autorest. Once I'ver merged it into master, I will tag it v1.0.0.

Is there a standard for embedding a Version within Go code (e.g., a version file)?

ahmetalpbalkan:
@brendandixon there's no standard. you can create a const and update before tagging each release.

autorest/client.go
```diff
@@ -28,16 +32,69 @@ const (
...
+// body is restored after being emitted.
+//
+// Note: Since it reads the entire Body, this decorator should not be used where body streaming is
+// important. It is best used to trace JSON or similar body values.
+func (li LoggingInspector) WithInspection() PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			defer r.Body.Close()
+
+			body := &bytes.Buffer{}
```
ahmetalpbalkan: you can potentially do `var body, b bytes.Buffer` and use them with `&b` below.

brendandixon:
@ahmetalpbalkan Ok. I'll do that. :)

autorest/mocks/helpers.go
```diff
@@ -59,3 +76,43 @@ func SetResponseHeader(resp *http.Response, h string, v string) {
...
+}
+
+// SetRetryHeader adds the Retry-After header.
+func SetRetryHeader(resp *http.Response, delay time.Duration) {
+	SetResponseHeader(resp, http.CanonicalHeaderKey(headerRetryAfter), fmt.Sprintf("%v", delay.Seconds()))
+}
+
+// IsSameSlice performs as a trivial equality check for two slices. The slices are equal if they
+// are the same length and contain the same values in the same order.
+func IsSameSlice(s1 []string, s2 []string) bool {
```
ahmetalpbalkan: this is called `reflect.DeepEqual`

autorest/responder.go
```diff
@@ -135,3 +142,22 @@ func WithErrorUnlessStatusCode(codes ...int) RespondDecorator {
 func WithErrorUnlessOK() RespondDecorator {
 	return WithErrorUnlessStatusCode(http.StatusOK)
 }
+
+// ExtractHeader extracts all values of the specified header from the http.Response. It returns an
+// empty string slice if the passed http.Response is nil or the header does not exist.
+func ExtractHeader(header string, resp *http.Response) []string {
```
ahmetalpbalkan: you can potentially return a nil slice as well, semantically sounds a bit more appropriate.

autorest/mocks/helpers.go
brendandixon: @ahmetalpbalkan I think you're right. Should've remembered that...

autorest/responder.go
brendandixon: @ahmetalpbalkan I could, but then the caller would have to check for nil, correct? I was leaning towards a model where the caller cannot make an error.

ahmetalpbalkan: umm, not really. nil slice behaves exactly like empty slice if you're trying to append, len() and such.

ahmetalpbalkan:
LGTM :shipit:

autorest/responder.go
brendandixon: @ahmetalpbalkan Ah, I'll change as I suspect nil is more idiomatic.

PR #4, opened 2015-11-09
Implement certificate based auth for service principal tokens

colemickens:
This PR enables the use of certificates to retrieve a service principal auth token.

It replaces `clientSecret` on SPT with a field `secret` which is a struct implementing `ServicePrincipalSecret`. In `Refresh()`, rather than just adding the "client_secret" to the form, we call `ServicePrincipalSecret`'s `SetAuthValues` to modify the form as needed.

`ServicePrincipalTokenSecret` will add `client_secret` to the request.

`ServicePrincipalCertSecret` will add `client_assertion` and `client_assertion_type` to the request. 

It does introduce two new dependencies to `go-autorest`:
 
* `github.com/dgrijalva/jwt-go`
* `golang.org/x/crypto/pkcs12` (the one Paul just pushed back into x/crypto)

It *could* be possible to avoid those by:

1. Reimplementing the few things needed from `jwt-go`.
2. Requiring the user to specify a PKCS#1-formatted or PKCS#8-formatted certificate instead of a PKCS#12 certificate. But, most Microsoft stuff seems to prefer PKCS#12 (most relevantly, Azure KeyVault).

I've tested this manually, and also included an example in the PR that starts from scratch and results in listing the subscription's resource groups, after retrieving an OAuth token via JWT signed with the cert's private key.

### DIFF
--- a/autorest/azure/token.go
+++ b/autorest/azure/token.go
@@ -1,18 +1,27 @@
 package azure
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha1"
+	"encoding/base64"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
 	"time"
 
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/dgrijalva/jwt-go"
+	"golang.org/x/crypto/pkcs12"
 )
 
 const (
 	defaultRefresh = 5 * time.Minute
 	oauthURL       = "https://login.microsoftonline.com/{tenantID}/oauth2/{requestType}?api-version=1.0"
 	tokenBaseDate  = "1970-01-01T00:00:00Z"
+
+	jwtAudienceTemplate = "https://login.microsoftonline.com/%s/oauth2/token"
 
 	// AzureResourceManagerScope is the OAuth scope for the Azure Resource Manager.
 	AzureResourceManagerScope = "https://management.azure.com/"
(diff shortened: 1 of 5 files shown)


autorest/azure/token.go
```diff
@@ -1,12 +1,19 @@
...
+	"encoding/base64"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
 	"time"
 
 	"github.com/Azure/go-autorest/autorest"
+	jwt "github.com/dgrijalva/jwt-go"
+	pkcs12 "golang.org/x/crypto/pkcs12"
```
brendandixon: I'm not a huge fan of unnecessarily naming imports. It's not a pattern I've seen broadly used. It is intended as an escape value when there are unavoidable conflicts.

Also, is "jwt" necessary? One example (see https://github.com/someone1/gcp-jwt-go) uses "jwt" directly without renaming the import. Also, the package name is "jwt".

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
+	ClientSecret string
+}
+
+// ServicePrincipalCertSecret implements ServicePrincipalSecret for cert auth with signed JWTs
+type ServicePrincipalCertSecret struct {
+	Pkcs12Bytes []byte
+	Password    string
+}
+
+// NewServicePrincipalTokenWithSecret create a ServicePrincipalToken using the supplied ServicePrincipalSecret impl
```
brendandixon: I strongly prefer proper English in comments. Could we please spell out "implementation" vs. "impl". :)

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
 	refreshWithin time.Duration
 	sender        autorest.Sender
 }
 
-// NewServicePrincipalToken creates a ServicePrincipalToken from the supplied Service Principal
-// credentials scoped to the named resource.
-func NewServicePrincipalToken(id string, secret string, tenantID string, resource string) (*ServicePrincipalToken, error) {
+// ServicePrincipalSecret is an interface that allows various secret mechanism to fill the form
+// that is submitted when acquiring an oAuth token
+type ServicePrincipalSecret interface {
```
brendandixon: I prefer a top-down patten. Please move the types upon which ServicePrincipalToken depends before its definition.

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
+	SetAuthValues(spt *ServicePrincipalToken, values *url.Values) error
+}
+
+// ServicePrincipalTokenSecret implements ServicePrincipalSecret for client_secret type authorization
+type ServicePrincipalTokenSecret struct {
+	ClientSecret string
+}
+
+// ServicePrincipalCertSecret implements ServicePrincipalSecret for cert auth with signed JWTs
```
brendandixon: Could we please spell out all abbreviations (e.g,. certificate, authentication). JWT is fine since it is an industry term.

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
-func NewServicePrincipalToken(id string, secret string, tenantID string, resource string) (*ServicePrincipalToken, error) {
+// ServicePrincipalSecret is an interface that allows various secret mechanism to fill the form
+// that is submitted when acquiring an oAuth token
+type ServicePrincipalSecret interface {
+	SetAuthValues(spt *ServicePrincipalToken, values *url.Values) error
+}
+
+// ServicePrincipalTokenSecret implements ServicePrincipalSecret for client_secret type authorization
+type ServicePrincipalTokenSecret struct {
+	ClientSecret string
```
brendandixon: Could we rename "ClientSecret" to just "Secret". I don't see "Client" adding any semantic value. It is a "Secret" that is passed. It has no binding to a client.

colemickens: I chose that name because the form field name that is utilized is "client_secret" on the form. I'm fine either way though.

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
+}
+
+// ServicePrincipalCertSecret implements ServicePrincipalSecret for cert auth with signed JWTs
+type ServicePrincipalCertSecret struct {
+	Pkcs12Bytes []byte
```
brendandixon: I don't know that the name "Pkcs12Bytes" is very go-ish. Go encourages minimal names that do not repeat otherwise available information. "Pkcs12" may be sufficient.

autorest/azure/token.go
```diff
@@ -70,29 +77,72 @@ func (t *Token) WithAuthorization() autorest.PrepareDecorator {
...
+		tenantID,
+		resource,
+		&ServicePrincipalTokenSecret{
+			ClientSecret: secret,
+		},
+	)
+}
+
+// NewServicePrincipalTokenFromCert create a ServicePrincipalToken from the supplied certData
+func NewServicePrincipalTokenFromCert(id string, pkcs12bytes []byte, password string, tenantID string, resource string) (*ServicePrincipalToken, error) {
```
brendandixon: Same comment on "pkcs12bytes". Likely just use "pkcs12".

autorest/azure/token.go
colemickens: Yes. I assume "oauth" is an okay abbreviation as well?

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
 
 // Refresh obtains a fresh token for the Service Principal.
 func (spt *ServicePrincipalToken) Refresh() error {
+	newToken, err := spt.GetOauthToken()
```
brendandixon: It's not clear to me why you moved the logic from Refresh into a separate method. I am guessing you needed the method in some other context? Still, the original design was for Refresh to handle all cases. I prefer that we move the logic back into Refresh.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+		return err
+	}
+	spt.Token = *newToken
+
+	return nil
+}
+
+// SetAuthValues is a method of the interface ServicePrincipalTokenSecret
+// It will populate the form submitted during oAuth Token Acquisition using a static client_secret
+func (tokenSecret *ServicePrincipalTokenSecret) SetAuthValues(spt *ServicePrincipalToken, v *url.Values) error {
```
brendandixon: I strongly prefer methods defined on a struct to be kept adjacent to the struct itself.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+		return "", err
+	}
+
+	signedString, err := token.SignedString(rsaPrivKey)
+	return signedString, nil
+}
+
+// SetAuthValues is a method of the interface ServicePrincipalTokenSecret
+// It will populate the form submitted during oAuth Token Acquisition using a JWT signed with a certificate
+func (secret *ServicePrincipalCertSecret) SetAuthValues(spt *ServicePrincipalToken, v *url.Values) error {
```
brendandixon: Please move this adjacent to the struct on which it is defined.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+
+// SetAuthValues is a method of the interface ServicePrincipalTokenSecret
+// It will populate the form submitted during oAuth Token Acquisition using a static client_secret
+func (tokenSecret *ServicePrincipalTokenSecret) SetAuthValues(spt *ServicePrincipalToken, v *url.Values) error {
+	v.Set("client_secret", tokenSecret.ClientSecret)
+	return nil
+}
+
+// SignedJwt returns the JWT signed with the secret's certData
+func (secret *ServicePrincipalCertSecret) SignedJwt(spt *ServicePrincipalToken) (string, error) {
```
brendandixon: Please move this adjacent to the defining struct.

brendandixon: Also, I prefer verbs for names. That is, I believe this is better called "SignJwt" vs. "SignedJwt". SignedJwt is a noun that defines a thing. SignJwt is an action that describes the method.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+	privateKey, cert, err := pkcs12.Decode(secret.Pkcs12Bytes, secret.Password)
+	if err != nil {
+		return "", err
+	}
+
+	rsaPrivKey, isRsaKey := privateKey.(*rsa.PrivateKey)
```
brendandixon: Can we please rename "rsaPrivKey" to "rsaPrivateKey."

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+	if !isRsaKey {
+		return "", fmt.Errorf("PKCS12 certificate must contain an RSA private key")
+	}
+
+	hasher := sha1.New()
```
brendandixon: Could these two statements be combined (e.g., sha1.New().Write(cert.Raw)). Since sha1.New does not return an error, introducing an unnecessary variable, to me, clutters things.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+	_, err = hasher.Write(cert.Raw)
+	if err != nil {
+		return "", err
+	}
+
+	thumbprint := base64.URLEncoding.EncodeToString(hasher.Sum(nil))
+
+	jti := make([]byte, 20)
```
brendandixon: What is "jti"? I do not find this variable name, especially with no surrounding comments, helpful.

autorest/azure/token.go
```diff
@@ -104,6 +154,79 @@ func (spt *ServicePrincipalToken) EnsureFresh() error {
...
+	_, err = rand.Read(jti)
+	if err != nil {
+		return "", err
+	}
+
+	token := jwt.New(jwt.SigningMethodRS256)
+	token.Header["x5t"] = thumbprint
+	token.Claims["aud"] = fmt.Sprintf("https://login.microsoftonline.com/%s/oauth2/token", spt.tenantID)
```
brendandixon: It may be cleaner to set token.Claims to a map (e.g., token.Claims = map[string]interface{}{ xxx: yyyy, zzzz: xxxx, ... }).

brendandixon: Please define a constant to hold the format string. Burying these in code makes finding / fixing / changing them very hard.

autorest/azure/token.go
colemickens: Yes. This is just an artifcat of how I got here. I agree it should be back in Refresh()

autorest/azure/token.go
colemickens: It's a JWT standard. I added a comment.

autorest/azure/token.go
brendandixon: Ah, got it. Makes sense.

autorest/azure/token.go
brendandixon: I believe so. I am a bit odd on this. I react strongly against the "hacker" culture at this point. I have had to maintain code with poor variable names and no comments. Those are not happy times. Hence, I push for good variable names, fully spelled out (without becoming the horrifically long Java-esque style). Ok?

colemickens: Yep, makes sense.

autorest/azure/token.go
brendandixon: Yah! I try to balance two goals: Get what I need done with a few lines /  methods as possible and don't make any one method too long (e.g., more than, say, 50-75 lines).

brendandixon:
Looks good to me with one tiny addition: Could you please update the semantic version in the version.go file? I believe this bumps us to 1.0.2. With that I'll gladly accept!

colemickens:
Should it be 1.1.0? Assuming semver, this is a backward-compatible feature addition, rather than just a bugfix. http://semver.org/

colemickens:
I squashed it all and bumped the version to 1.1.0 per semver.

PR #9, opened 2015-11-30
return pointer from to.StringMapPtr

ahmetalpbalkan:
All Tags fields actually accept `*map[string]*string` in Azure SDK. (`ack --go Tags | grep -v '^//' | grep map`)

cc: @brendandixon PTAL

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v2.0.0
+
+- Changed `to.StringMapPtr` method signature to return a pointer
+
 ## v1.0.0
 
 - Added Logging inspectors to trace http.Request / Response
--- a/autorest/to/convert.go
+++ b/autorest/to/convert.go
@@ -31,13 +31,13 @@ func StringMap(msp map[string]*string) map[string]string {
 	return ms
 }
 
-// StringMapPtr returns a map of string pointers built from the passed map of strings.
-func StringMapPtr(ms map[string]string) map[string]*string {
+// StringMapPtr returns a pointer to a map of string pointers built from the passed map of strings.
+func StringMapPtr(ms map[string]string) *map[string]*string {
 	msp := make(map[string]*string, len(ms))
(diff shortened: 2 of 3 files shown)


brendandixon:
@ahmetalpbalkan Could you, please, update the version as well. This change, as it alters the interface of a method, requires incrementing the minor version number. Thanks!

ahmetalpbalkan:
@brendandixon you mean major, right?

brendandixon:
@ahmetalpbalkan Yes, sadly for such a small change.

brendandixon:
@ahmetalpbalkan If, however, you were to add a new method alongside the old, then a minor increase. But that might be confusing...

ahmetalpbalkan:
Yeah this is definitely the major. I'll update changelog for now, however normally we should just keep accepting PRs until we decide to make a release and then update the changelog ourselves.

brendandixon:
@ahmetalpbalkan We could, but other code pulls from master and this will introduce a change...no?

ahmetalpbalkan:
> other code pulls from master

hmm, it shouldn't. Azure SDK has a copy of this package. We just need to update with godep over there.

brendandixon:
@ahmetalpbalkan That assumes the only consumer is the Azure SDK. Likely true, but still an unproven assumption.

ahmetalpbalkan:
@brendandixon nothing we can do about breaking the master branch. :) Maybe we should add a notice to readme and suggest using tags/releases of the source instead.

This LGTM now?

brendandixon:
@ahmetalpbalkan Lame. And, yes, I try to push tags.

ahmetalpbalkan:
Tagged v2.0.0

brendandixon:
@ahmetalpbalkan Peachy. Though I thought we were going to collect and then tag. No worries.

ahmetalpbalkan:
@brendandixon I just tagged anyway since I changed the CHANGELOG already.

PR #10, opened 2015-11-30
remove pkcs12 support, add generic cert support

colemickens:
ServicePrincipalCertificateSecret was overly pkcs12-specific. As it turns out, KeyVault requires a PKCS12 but places them on box as PKCS8.

This change makes ServicePrincipalCertificateSecret more generic. It also updates the example to decode the pkcs12 certificate and then use ServicePrincipalCertificateSecret as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v2.0.0
 
 - Changed `to.StringMapPtr` method signature to return a pointer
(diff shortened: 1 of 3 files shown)


colemickens:
@brendandixon @ahmetalpbalkan , can you take a look? Going to update this commit with a changelog entry right now as well.

brendandixon:
@colemickens @ahmetalpbalkan LGTM. Merging.

PR #12, opened 2016-01-05
Add XML unmarshaling support

ahmetalpbalkan:


colemickens:
Do some of the Azure APIs return XML by default? Just curious, or is this for something else?

PR #13, opened 2016-01-18
Add responder for Azure-specific errors

ahmetalpbalkan:
@brendandixon/@colemickens,

I took a stab at handling the Azure error responses specifically to allow them to be deserialized and surfaced to the caller in a nice, type-safe way. An Azure service error response looks like this:

```json
{
    "error": {
        "code": "LocationRequired",
        "message": "The location property is required for this definition."
    }
}
```

Logic is as follows:

* if statuscode=4xx or 5xx:
    * read the response body (and close) and unmarshal into `*AzureError` type.
    * if `*AzureError` looks alright: return it
    * otherwise: let the next decorator run

Downsides:
* the next decorator won't have the resp.Body anymore (such as WithErrorUnlessStatusCode will be returning a closed resp.Body)
* will work if Azure always returns this kind of a response.

Please let me know what you think.

We'll probably make this `WithAzureErrorCheckUnlessStatusCode(...)` if we end up using, but I wanted to stop here and get feedback.

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -6,6 +6,8 @@ See the included examples for more detail.
 package azure
 
 import (
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 
@@ -24,6 +26,22 @@ const (
 	// in the response.
 	HeaderRequestID = "x-ms-request-id"
 )
+
+// Error describes an error response returned by Azure service.
+type Error struct {
+	ServiceError *struct {
+		Code    string `json:"code"`
+		Message string `json:"message"`
+	} `json:"error"` // JSON object returned in service response body
+	StatusCode int    // HTTP response status code
+	RequestID  string // x-ms-request-id-header
+}
+
+// Error returns a human-friendly error message from service error.
+func (e *Error) Error() string {
+	return fmt.Sprintf("azure: Service returned an error. Code=%q Message=%q Status=%d",
+		e.ServiceError.Code, e.ServiceError.Message, e.StatusCode)
+}
 
 // WithReturningClientID returns a PrepareDecorator that adds an HTTP extension header of
 // x-ms-client-request-id whose value is the passed, undecorated UUID (e.g.,
@@ -69,4 +87,38 @@ func ExtractClientID(resp *http.Response) string {
 // x-ms-request-id header.
 func ExtractRequestID(resp *http.Response) string {
 	return autorest.ExtractHeaderValue(HeaderRequestID, resp)
+}
+
+// WithErrorUnlessStatusCode returns a RespondDecorator that emits an
+// azure.Error by reading the response body unless the response HTTP status code
+// is among the set passed.
+//
+// If there is a chance service may return responses other than the Azure error
+// format and the response cannot be parsed into an error, a decoding error will
+// be returned containing the response body. In any case, the Responder will
+// return an error if the status code is not satisfied.
+//
+// If this Responder returns an error, the response body will be replaced with
+// an in-memory reader, which needs no further closing.
+func WithErrorUnlessStatusCode(codes ...int) autorest.RespondDecorator {
+	return func(r autorest.Responder) autorest.Responder {
+		return autorest.ResponderFunc(func(resp *http.Response) error {
+			err := r.Respond(resp)
+			if err == nil && !autorest.ResponseHasStatusCode(resp, codes...) {
+				var e Error
+				defer resp.Body.Close()
+
+				b, decodeErr := autorest.CopyAndDecode(autorest.EncodedAsJSON, resp.Body, &e)
+				resp.Body = ioutil.NopCloser(&b) // replace body with in-memory reader
+				if decodeErr != nil || e.ServiceError == nil {
+					return fmt.Errorf("autorest/azure: error response cannot be parsed: %q error: %v", b.String(), err)
+				}
+
+				e.RequestID = ExtractRequestID(resp)
+				e.StatusCode = resp.StatusCode
+				err = &e
+			}
+			return err
+		})
+	}
 }
--- a/autorest/azure/azure_test.go
+++ b/autorest/azure/azure_test.go
@@ -2,22 +2,23 @@ package azure
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 	"testing"
 
-	. "github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/mocks"
 )
 
 // Use a Client Inspector to set the request identifier.
 func ExampleWithClientID() {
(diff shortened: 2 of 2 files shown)


autorest/azure/azure.go
```diff
@@ -70,3 +92,33 @@ func ExtractClientID(resp *http.Response) string {
...
+				// the Error field of AzureError anyway and in case of not
+				// returning any error here will be handled by the next responder.
+
+				if e != nil && e.ServiceError != nil {
+					e.RequestID = ExtractRequestID(resp)
+					e.StatusCode = resp.StatusCode
+					err = e
+				}
+			}
+			return err
```
brendandixon: @ahmetalpbalkan I like the general idea, but believe we should tweak this a bit.

First, returning nil and closing the Body will cause problems for subsequent Responders. I suggest that either the code read / replace the Body (so subsequent Responders have something to work with) or that it always return an error.

Second, since this code is within the "azure" package, I see no need to include Azure in any of the names (e.g., azure.WithErrrorUnlessStatusCode, azure.ServiceError, and azure.Error seems sufficiently clear and stutter free). Also, I feel it best to move this code into azure/error.go and azure/error_test.go files.

Further, it may be better if the code followed the pattern of autorest.WithErrorUnlessStatusCode (or similar). Services document their error codes in the corresponding Swagger (and documentation), so the check can be precise. If you are uncomfortable with that, provide a default set of codes or default predicate to catch the error codes.

Lastly, I'd feel better the code manufactured an error if the json.Decode fails. The error could be generic and possibly populated with details from the Decode error which may be important under different debugging scenarios.

Ok?

ahmetalpbalkan: @brendandixon 

> First, returning nil and closing the Body will cause problems for subsequent Responders. I suggest that either the code read / replace the Body (so subsequent Responders have something to work with) or that it always return an error.

How does this sound: create a `io.NopCloser(strings.NewReader(str))` and assign it to Body?



> Second, since this code is within the "azure" package, I see no need to include Azure in any of the names (e.g., azure.WithErrrorUnlessStatusCode, azure.ServiceError, and azure.Error seems sufficiently clear and stutter free). Also, I feel it best to move this code into azure/error.go and azure/error_test.go files.

ok



> Further, it may be better if the code followed the pattern of autorest.WithErrorUnlessStatusCode (or similar). Services document their error codes in the corresponding Swagger (and documentation), so the check can be precise. If you are uncomfortable with that, provide a default set of codes or default predicate to catch the error codes.

yes we'll definitely need to refactor this `...UnlessStatusCode` (mentioned in the PR description, just wanted to get quick feedback).



> Lastly, I'd feel better the code manufactured an error if the json.Decode fails. The error could be generic and possibly populated with details from the Decode error which may be important under different debugging scenarios.

What are we gonna do with the error surfaced from decode? This is for the cases such as if we get an IIS Error Page (text/html). We can't surface decode json failed. I think the appropriate thing would be have this and let autorest surface its own error(+body) with:

```go
  ...,
  azure.WithUnlessStatusCode()
  autorest.WithErrorUnlessStatusCode(...)
)
```

in Azure SDK?

ahmetalpbalkan:
@brendandixon per our offline sync and changes committed today, I came up with `azure.WithErrorUnlessStatusCode` and updated tests to reflect behavior regarding keeping the response body intact, handling of various good/malformed responses.

ahmetalpbalkan:
Oh and it appears like some API endpoints _may_ return a field like `target` in addition to `message` and `code` from what I read in azure-rest-api-specs repo. The specs in that repo are not particularly in a good shape: some don't even define an error, some define error but only for certain operations or operation types and some just define `message`/`target`/`code` triplet however it's clear from my experience that target may not always be returned.

So for now I'm planning to leave `"target"` out until we see that `message`/`code` are inconclusive.

ahmetalpbalkan:
@brendandixon another thing is I found some piece of generated code that makes impossible this function to run:

Pay attention to duplication status codes below (quoting some random piece of gen'd code)

* client.Send
* azure.WithErrorUnlessStatusCode

```go
func (client EndpointsClient) CreateSender(req *http.Request) (*http.Response, error) {
	return client.Send(req, http.StatusOK, http.StatusCreated, http.StatusAccepted)
}


func (client EndpointsClient) CreateResponder(resp *http.Response) (result TrackedResource, err error) {
	err = autorest.Respond(
		resp,
		client.ByInspecting(),
		azure.WithErrorUnlessStatusCode(http.StatusOK, http.StatusCreated, http.StatusAccepted),
		autorest.ByUnmarshallingJSON(&result),
		autorest.ByClosing())
	result.Response = autorest.Response{Response: resp}
	return
}

```

I think `client.Send()` should not take any status codes, or actually we should remove DoErrorUnlessStatusCode check in `Send()`:

```go
	resp, err := SendWithSender(c, req,
		DoErrorUnlessStatusCode(codes...))
```

this is taking precedence over azure error check. I'm hotfixing all these locally but let's remove this decorator next time we regenerate the code.

ahmetalpbalkan:
Hmm I now realized that in an actual usage my error gets wrapped in autorest.baseError (I don't know why this doesn't happen in the test cases I submitted)

I get an error like:

```
network/VirtualNetworksClient:CreateOrUpdate 400 Failure responding to request -- Original Error: azure: Service returned an error. Code="InvalidAddressPrefixFormat" Message="Address prefix  of resource /subscriptions/27b750cd-ed43-42fd-9044-8d75e124ae55/resourceGroups/docker-machine/providers/Microsoft.Network/virtualNetworks/docker-machine is not formatted correctly. It should follow CIDR notation, for example 10.0.0.0/24." Status=400
```

but I actually want caller to have the original error (`*azure.Error`) I am surfacing. @brendandixon what do you think?

brendandixon:
@ahmetalpbalkan autorest.NewErrorWithError returns the original error if it is already an instance of autorest.Error (see https://github.com/Azure/go-autorest/blob/master/autorest/error.go#L62). I put it there to save all callers having to make the check. Perhaps a similar approach would work for azure.Error? You could have a azure.NewError(xxxx) that returns the original error *if* it is an instance of azure.Error.

Also, I agree with removing the autorest.DoErrorUnlessStatusCode decorator from the default autorest.Client.Send method.

ahmetalpbalkan:
@brendandixon let's discuss this. Inner errors are often not a thing in Go and not commonly used. Even I'm not hoping to use fields of azure.Error. The message often should be just fine. But in this case we're unnecessarily wrapping an already meaningful error. Also I miss the point of always wrapping everything in a `autorest.baseError`, I'm not sure how will it be useful to anyone.

Actually the only thing I care is, just return me an `*azure.Error` if `err` is of that type, don't ever wrap it. `*azure.Error` is not meant for having InnerError, it is a standalone error type returned when status code is not met (unlike autorest.baseError).

brendandixon:
@ahmetalpbalkan Agreed. Let's chat.

ahmetalpbalkan:
@brendandixon we're green here too.

PR #14, opened 2016-01-19
Add StringSlice/StringSlicePtr

ahmetalpbalkan:
Azure Go SDK makes use of `*[]string` type in various models.
[`arm/network.AddressSpace`][1] is an example. Adding helper methods to
handle conversion of string slices to pointers and vice versa.

Signed-off-by: Ahmet Alp Balkan
cc: @brendandixon 

[1]: https://github.com/Azure/azure-sdk-for-go/blob/master/arm/network/models.go#L440

### DIFF
--- a/autorest/to/convert.go
+++ b/autorest/to/convert.go
@@ -17,6 +17,20 @@ func StringPtr(s string) *string {
 	return &s
 }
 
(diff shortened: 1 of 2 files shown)


brendandixon:
@ahmetalpbalkan LGTM. Merging.

ahmetalpbalkan:
thanks

PR #15, opened 2016-01-19
Created NewDecoder and CopyAndDecode helpers

brendandixon:
@ahmetalpbalkan @colemickens 

I factored out code previously used by ByUnmarshallingJSON and ByUnmarshallingXML into helpers. Doing so replaces the magic strings with constants and provides a CopyAndDecode helper useful in other situations.

### DIFF
--- a/autorest/responder.go
+++ b/autorest/responder.go
@@ -1,11 +1,7 @@
 package autorest
 
 import (
-	"bytes"
-	"encoding/json"
-	"encoding/xml"
 	"fmt"
-	"io"
 	"net/http"
 )
 
@@ -28,11 +24,6 @@ func (rf ResponderFunc) Respond(r *http.Response) error {
 // RespondDecorator takes and possibly decorates, by wrapping, a Responder. Decorators may react to
 // the http.Response and pass it along or, first, pass the http.Response along then react.
 type RespondDecorator func(Responder) Responder
-
-// decoder is a common $interface for xml.Decoder and json.Decoder.
-type decoder interface {
-	Decode(v interface{}) error
-}
 
 // CreateResponder creates, decorates, and returns a Responder. Without decorators, the returned
 // Responder returns the passed http.Response unmodified. Responders may or may not be safe to share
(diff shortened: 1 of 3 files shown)


autorest/utility.go
```diff
@@ -1,10 +1,47 @@
...
+func NewDecoder(encodedAs EncodedAs, r io.Reader) Decoder {
+	if encodedAs == EncodedAsJSON {
+		return json.NewDecoder(r)
+	} else if encodedAs == EncodedAsXML {
+		return xml.NewDecoder(r)
+	}
+	return nil
+}
+
+func CopyAndDecode(encodedAs EncodedAs, r io.Reader, v interface{}) (bytes.Buffer, error) {
```
ahmetalpbalkan: I think exported method requires a comment.

autorest/responder.go
```diff
@@ -109,25 +100,26 @@ func ByClosingIfError() RespondDecorator {
...
 func ByUnmarshallingXML(v interface{}) RespondDecorator {
-	return byUnmarshalling(v, "XML", func(r io.Reader) decoder { return xml.NewDecoder(r) })
+	return ByUnmarshallingAs(EncodedAsXML, v)
 }
 
-func byUnmarshalling(v interface{}, format string, df func(io.Reader) decoder) RespondDecorator {
+// ByUnmarshallingAs returns a RespondDecorator that decodes the data returned in the response
+// Body, based on the passed encoding, into the value pointed to by v.
+// encodedAs specifies the type of encoding and v provides the destination.
+func ByUnmarshallingAs(encodedAs EncodedAs, v interface{}) RespondDecorator {
```
ahmetalpbalkan: do we need to export this?

autorest/utility_test.go
```diff
@@ -24,6 +27,64 @@ const (
...
+func TestCopyAndDecodeReturnsXMLDecodingErrors(t *testing.T) {
+	_, err := CopyAndDecode(EncodedAsXML, strings.NewReader(xmlT[0:len(xmlT)-2]), &mocks.T{})
+	if err == nil {
+		t.Errorf("autorest: CopyAndDecode failed to return an error with invalid XML")
+	}
+}
+
+func TestCopyAndDecodeAlwaysReturnsACopy(t *testing.T) {
+	b, _ := CopyAndDecode(EncodedAsJSON, strings.NewReader(jsonT), &mocks.T{})
+	if len(b.String()) != len(jsonT) || b.String() != jsonT {
```
ahmetalpbalkan: `len(b.String()) != len(jsonT)` is unnecessary?

autorest/responder.go
brendandixon: @ahmetalpbalkan Not strictly necessary, but I see no harm. Are you concerned about polluting the namespace?

autorest/utility_test.go
brendandixon: @ahmetalpbalkan Nope. I'll remove it.

autorest/responder.go
ahmetalpbalkan: yeah it may be confusing to others. API services usually don't respond in other formats nowadays so I can't see anyone extending the pkg beyond json/xml.

autorest/utility_test.go
```diff
@@ -24,6 +27,64 @@ const (
...
+}
+
+func TestCopyAndDecodeReturnsJSONDecodingErrors(t *testing.T) {
+	_, err := CopyAndDecode(EncodedAsJSON, strings.NewReader(jsonT[0:len(jsonT)-2]), &mocks.T{})
+	if err == nil {
+		t.Errorf("autorest: CopyAndDecode failed to return an error with invalid JSON")
+	}
+}
+
+func TestCopyAndDecodeReturnsXMLDecodingErrors(t *testing.T) {
```
ahmetalpbalkan: Now that we have the logic tested downstream, do we really need to keep the all tests I added for ByUnmarshallingXML and the tests you had for ByUnmarshallingJSON. I sense some duplication in test cases (such as expecting error for malformed xml/json) now.

brendandixon: @ahmetalpbalkan I believe in testing each logical case even if there is overlap. That prevents gaps in the future which simple code coverage analysis will not catch. It's cheap insurance.

autorest/utility.go
brendandixon: @ahmetalpbalkan Good catch. Will add.

autorest/responder.go
brendandixon: @ahmetalpbalkan Ok. I'll mark it private.

autorest/utility_test.go
ahmetalpbalkan: :+1:

ahmetalpbalkan:
Looks good to me, let's merge. Thanks!

PR #18, opened 2016-01-20
Use http.DefaultClient instance

ahmetalpbalkan:
Little idiomatic change, should not have any semantic difference at all.


brendandixon:
@ahmetalpbalkan I'm fine with this. Merging / closing.

PR #19, opened 2016-01-20
Removing codes... from Send()

ahmetalpbalkan:
Breaking change. This will require changing the call to `Send()` in the auto-generated SDK.

image


ahmetalpbalkan:
@brendandixon the project still seems inactive on travis. could you please take a look?

ahmetalpbalkan:
CI green. @brendandixon

PR #20, opened 2016-01-20
fix token crash

colemickens:
token.go was updated to read the status code, but there isn't always a status code, since the client can return errors before the request even is sent.

### DIFF
--- a/autorest/azure/token.go
+++ b/autorest/azure/token.go
@@ -235,8 +235,12 @@ func (spt *ServicePrincipalToken) Refresh() error {
 
 	resp, err := autorest.SendWithSender(spt.sender, req)
 	if err != nil {
+		statusCode := autorest.UndefinedStatusCode
+		if resp != nil {
+			statusCode = resp.StatusCode
+		}
 		return autorest.NewErrorWithError(err,
-			"azure.ServicePrincipalToken", "Refresh", resp.StatusCode, "Failure sending request for Service Principal %s",
+			"azure.ServicePrincipalToken", "Refresh", statusCode, "Failure sending request for Service Principal %s",
 			spt.clientID)
 	}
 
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
I'm seeing the same error pattern in azure-sdk-for-go.

See `network/publicipaddresses.go`:

```go
func (client PublicIPAddressesClient) CreateOrUpdate(resourceGroupName string, publicIPAddressName string, parameters PublicIPAddress) (result PublicIPAddress, ae error) {
	req, err := client.CreateOrUpdatePreparer(resourceGroupName, publicIPAddressName, parameters)
	if err != nil {
		return result, autorest.NewErrorWithError(err, "network/PublicIPAddressesClient", "CreateOrUpdate", autorest.UndefinedStatusCode, "Failure preparing request")
	}

	resp, err := client.CreateOrUpdateSender(req)
	if err != nil {
		result.Response = autorest.Response{Response: resp}
		return result, autorest.NewErrorWithError(err, "network/PublicIPAddressesClient", "CreateOrUpdate", resp.StatusCode, "Failure sending request")
	}
    ...
```

Basically, in `resp, err := client.CreateOrUpdateSender(req)` if the `resp` comes back as nil (let's say low-level network issue?) it totally panics will nil dereference error in `resp.Status`.

I can easily induce this by disconnecting network, right before the call to the client method.

I'm thinking of opening an issue in azure-sdk-for-go, but the sensible option is probably making the `if resp==nil` check in go-autorest, because both the SDK and token.go call this `NewErrorWithError` method in go-autorest.

colemickens:
So `NewErrorWithError` would take the response pointer itself and would do the null check? I didn't consider changing what we were passing to NewErrorWithError, but this makes sense if the generated code can also hit this case.

(In my case I hadn't yet put my new mitmproxy in my cert store, so my app was blowing up in an x509 check).

brendandixon:
@colemickens @ahmetalpbalkan Let's change NewErrorWithError as suggested.

colemickens:
Agreed. Closing this.

PR #21, opened 2016-01-20
switch godep->glide

colemickens:
Remove everything vendored by godep.

Introduce a glide.yaml and glide.lock which respectively define our expectation of build dependencies and a description of exactly what version was resolved for each dependencies.

For now we `.gitignore` the `vendor/` directory, so we are not storing a copy of our dependencies in this repo. The glide.yaml or glide.lock file is sufficient for reconstituting dependencies, provided the repo doesn't vanish.


colemickens:
When this is in, I will send the same for `azure-sdk-for-go`.

ahmetalpbalkan:
travis is busted

colemickens:
? The build for this PR is busted or Travis itself is busted? I can't seem to view the build status for the PRs. I am told the repo isn't active?

ahmetalpbalkan:
our travis integration is also broken since yesterday.

ahmetalpbalkan:
I'd encourage updating arm/README as well.

colemickens:
Will do when I send the azure-sdk-for-go PR.

ahmetalpbalkan:
we should take a tag snapshot before merging this

colemickens:
Just rebased on master. cc: @brendandixon Are we proceeding with this?

.gitignore
```diff
@@ -0,0 +1 @@
+vendor/
```
ahmetalpbalkan: while you're at it can you adopt the standard Go.gitignore?

colemickens: Introduced it in a separate commit, but it's now in this PR.

brendandixon:
@colemickens @ahmetalpbalkan I want to proceed, but we **cannot** break the arm packages. Let's not de-vendor (wow! that's not even close to real English) those until we can regenerate and so forth. We should be fine for now since we've snapped a copy of go-autorest to that project.

brendandixon:
@colemickens @ahmetalpbalkan I don't see the removal of the Godeps paths to the various files (e.g., azure/token.go). Am I missing it?

colemickens:
@brendandixon It's at the bottom after all of the vendored file deletions.

ahmetalpbalkan:
@brendandixon hmm this change should not be breaking azure-sdk-for-go users as they already have a copy under `azure-sdk-for-go/Godeps/_workspace/src/github.com/Azure/go-autorest`. I don't see why not to merge.

brendandixon:
@colemickens @ahmetalpbalkan Merging...

PR #23, opened 2016-01-21
Fix nil dereference panic in NewErrorWithError when resp=nil

ahmetalpbalkan:
In cases where resp would be nil (such as net.Dial failures) code
in azure sdk et al. just panics with nil dereference error.

I am making NewErrorWithError take an *http.Response instead of
status code and handling nil of this argument internally.

Also removed NewErrorWithStatusCode and added NewErrorWithResponse
for cases where inner error does not exist.


brendandixon:
@ahmetalpbalkan LGTM.

ahmetalpbalkan:
@brendandixon let's release a tag.

brendandixon:
@ahmetalpbalkan Good idea! Let me know how it goes. :)

PR #25, opened 2016-01-26
OAuth2 Device Flow + Testing Coverage

colemickens:
1. Add support for the OAuth2 Device Flow

2. Adds tests for the Device Flow

3. Adds helpers and tests for persisting and restoring tokens

4. Adds support for manual "bring-your-own-tokens" in ServicePrincipalToken

5. Add tests for manual token and certificate based ServicePrincipalToken

6. Updates the example with support for device auth, and persisting and restoring the token between executions

The code coverage in `autorest/azure` is now:
* azure.go: `96.2%`
* devicetoken.go: `100%`
* persist.go: `92.0%`
* token.go: `96.2%`


colemickens:
[coverage.html in a zip](https://github.com/Azure/go-autorest/files/104695/coverage.zip) for those that are curious and/or lazy.

brendandixon:
@colemickens Any reason for not taking it to 100%

colemickens:
I can add a test for the Azure Error stuff that Ahmet added which will boost it a bit more. Otherwise the other error conditions seemed to me like things that would only occur due to transient IO errors. If you have suggestions on how I can hit those branches, I will implement them.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+	case "code_expired":
+		return nil, ErrDeviceCodeExpired
+	default:
+		return nil, ErrDeviceGeneric
+	}
+}
+
+// WaitForUserCompletion calls CheckForUserCompletion repeatedly until a token is granted or an error state occurs.
+// This prevents the user from looping and checking against 'ErrDeviceAuthorizationPending'.
+func WaitForUserCompletion(client *autorest.Client, clientID string, code *DeviceCode) (*Token, error) {
```
ahmetalpbalkan: Does this loop indefinitely? We'll probably need a timeout.

ahmetalpbalkan:
Awesome PR! I will review today.

brendandixon:
@colemickens Mocks, I use Mocks. And fake up error situations through them. Good testing covers error cases as well as normal path.

colemickens:
@brendandixon I've used mocks to cover a large number of error conditions. I'll stop by and we can chat about the few that remain.

@ahmetalpbalkan Azure will timeout after the ExpiresIn interval returned in the DeviceCode. I'll add code that ensures we stop after that time even if Azure doesn't time out.

ahmetalpbalkan:
@colemickens in that case that's not very necessary.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+// InitiateDeviceAuth initiates a device auth flow. It returns a DeviceCode
+// that can be used with CheckForUserCompletion or WaitForUserCompletion.
+func InitiateDeviceAuth(client *autorest.Client, clientID, resource string) (*DeviceCode, error) {
+	req, _ := autorest.Prepare(
+		&http.Request{},
+		autorest.AsPost(),
+		autorest.WithBaseURL(OAuthDeviceEndpoint),
+		autorest.WithFormData(url.Values{
+			"client_id": []string{clientID},
+			"resource":  []string{resource},
```
brendandixon: Any chance these values must be URL encoded?

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+	resp, err := client.Send(req)
+	if err != nil {
+		return nil, fmt.Errorf("Error occurred while requesting Device Authorization Code. %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf(
+			"Request for Device Authorization Code resulted in Status: %d: %s",
+			resp.StatusCode,
+			http.StatusText(resp.StatusCode))
+	}
```
brendandixon: When the Response body get closed if an error is returned? Might there be a case where the Body is present?

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+
+	var code DeviceCode
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&code)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to decode Device Authorization Code json")
+	}
```
brendandixon: Any reason for not using Respond and ByUnmarshallingJSON?

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+	resp, err := client.Send(req)
+	if err != nil {
+		return nil, fmt.Errorf("Error occurred while exchanging device code for token. %s", err)
+	}
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusBadRequest {
+		return nil, fmt.Errorf(
+			"Exchanging Device Code for OAuth token resulted in Status: %d: %s",
+			resp.StatusCode,
+			http.StatusText(resp.StatusCode))
+	}
```
brendandixon: Ditto as above regarding possibly open Response.Body.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+
+	var token DeviceToken
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&token)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to decode Device Token json")
+	}
+
```
brendandixon: Ditto as above regarding using Respond and ByUnmarshallingJSON.

autorest/azure/devicetoken.go
brendandixon: And, decoding does not close the Body.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+	// ErrDeviceAuthorizationPending represents the server waiting on the user to complete the device flow
+	ErrDeviceAuthorizationPending = fmt.Errorf("Error while retrieving OAuth token: Authorization Pending")
+
+	// ErrDeviceCodeExpired represents the server timing out and expiring the code during device flow
+	ErrDeviceCodeExpired = fmt.Errorf("Error while retrieving OAuth token: Code Expired")
+
+	// ErrDeviceSlowDown represents the service telling us we're polling too often during device flow
+	ErrDeviceSlowDown = fmt.Errorf("Error while retrieving OAuth token: Slow Down")
+)
```
brendandixon: Must these be exported?

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+		token, err := CheckForUserCompletion(client, clientID, code)
+
+		if err == nil {
+			return token, nil
+		}
+
+		switch err {
+		case ErrDeviceSlowDown:
+			waitDuration = waitDuration + (intervalDuration / 2)
+		case ErrDeviceAuthorizationPending:
```
brendandixon: Usually, waits get increasingly long (e.g., exponential back-off). These get longer, but they get longer by smaller amounts. Is that what we want?

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+func TestDeviceCodeIncludesResource(t *testing.T) {
+	sender := mocks.NewSender()
+	sender.EmitContent(MockDeviceCodeResponse)
+	sender.EmitStatus("OK", 200)
+	client := &autorest.Client{Sender: sender}
+
+	code, err := InitiateDeviceAuth(client, TestClientID, TestResource)
+	if err != nil {
+		t.Errorf("unexpected error initiating device auth")
+	}
```
brendandixon: We've been prepending all emitted errors with "autorest: " so they're clear in the output (e.g., see responder_test.go).

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+	sender := mocks.NewSender()
+	sender.EmitErrors(1)
+	sender.SetError(fmt.Errorf("this is an error"))
+	client := &autorest.Client{Sender: sender}
+
+	_, err := InitiateDeviceAuth(client, TestClientID, TestResource)
+	if err == nil {
+		t.Errorf("failed to get err") // (expecting Error occurred while requesting Device Authorization Code)
+	}
+}
```
brendandixon: Could you, please, add tests to ensure the Body is properly closed under all conditions?

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+func TestDeviceCodeReturnsErrorIfCannotDeserializeDeviceCode(t *testing.T) {
+	gibberishJSON := strings.Replace(MockDeviceCodeResponse, "expires_in", "\":, :gibberish", -1)
+	sender := mocks.NewSender()
+	sender.EmitContent(gibberishJSON)
+	client := &autorest.Client{Sender: sender}
+
+	_, err := InitiateDeviceAuth(client, TestClientID, TestResource)
+	if err == nil {
+		t.Errorf("failed to get error") // (expecting Failed to decode Device Authorization Code json)
+	}
```
brendandixon: Should we test to ensure we get the correct kind of error back?

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+func TestDeviceCodeReturnsErrorIfBadRequest(t *testing.T) {
+	sender := mocks.NewSender()
+	sender.EmitStatus("Bad Request", 400)
+	client := &autorest.Client{Sender: sender}
+
+	_, err := InitiateDeviceAuth(client, TestClientID, TestResource)
+	if err == nil {
+		t.Errorf("failed to get err") // (expecting Request for Device Authorization Code resulted in Status)
+	}
```
brendandixon: Should we test to ensure we get the correct kind of error back? Ditto throughout.

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+	sender := mocks.NewSender()
+	sender.EmitContent(errorDeviceTokenResponse("authorization_pending"))
+	sender.EmitStatus("Bad Request", 400)
+	client := &autorest.Client{Sender: sender}
+
+	go func() {
+		time.Sleep(1 * time.Second)
+		sender.EmitContent(errorDeviceTokenResponse("access_denied"))
+	}()
+
```
brendandixon: Go routines may not run w/o a sleep or some other break to give the Go scheduler a moment. You may need to add that here to ensure a reliable test.

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+	sender := mocks.NewSender()
+	sender.EmitContent(errorDeviceTokenResponse("slow_down"))
+	sender.EmitStatus("Bad Request", 400)
+	client := &autorest.Client{Sender: sender}
+
+	go func() {
+		time.Sleep(1 * time.Second)
+		sender.EmitContent(errorDeviceTokenResponse("access_denied"))
+	}()
+
```
brendandixon: Ditto as above.

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+	_, _ = WaitForUserCompletion(client, TestClientID, deviceCode())
+}
```
brendandixon: Where is the test in this routine? I see no line ever emitting a t.Errorf.

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+	sender.EmitStatus("Bad Request", 400)
+	client := &autorest.Client{Sender: sender}
+
+	_, err := WaitForUserCompletion(client, TestClientID, deviceCode())
+	if err == nil {
+		t.Errorf("failed to get error")
+	}
+	if err != ErrDeviceAccessDenied {
+		t.Errorf("got wrong error")
+	}
```
brendandixon: I believe it is enough to assert the correct error is returned. Nil will be, by definition, in that case "not a correct" error. Ditto throughout.

autorest/azure/example/main.go
```diff
@@ -17,83 +18,128 @@ import (
 
 const resourceGroupURLTemplate = "https://management.azure.com/subscriptions/{subscription-id}/resourcegroups"
 const apiVersion = "2015-01-01"
+const xplatClientID = "04b07795-8ddb-461a-bbee-02f9e1bf7b46"
```
brendandixon: Is this some magical, unchanging identity?

autorest/azure/persist_test.go
```diff
@@ -0,0 +1,132 @@
...
+	expectedToken := TestToken
+	actualToken, err := LoadToken(f.Name())
+	if err != nil {
+		t.Errorf("unexpected error loading token from file: %v", err)
+	}
+
+	if *actualToken != expectedToken {
+		t.Errorf("failed to decode properly actual(%v) expected(%v)", *actualToken, expectedToken)
+	}
+}
```
brendandixon: Pedantic comment: This test tests multiple things.

ahmetalpbalkan:
I feel like we should move devicetoken into a separate package. Mostly because it's an isolated feature. I'm not sure if it's technically possible as I'm not sure if it is relying on something from ./autorest/azure.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+		}),
+		autorest.WithQueryParameters(map[string]interface{}{
+			authAPIVersionQueryParamName: authAPIVersionQueryParamValue,
+		}))
+
+	resp, err := client.Send(req)
+	if err != nil {
+		return nil, fmt.Errorf("Error occurred while requesting Device Authorization Code. %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
```
ahmetalpbalkan: I think we can use responder decorators here to do status checks etc. (i.e. this package can have its own `withErrorUnlessStatusCode()` and you can close the body and deserialize json in that responder list easily as Brendan said.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+// It contains information to instruct the user to complete the auth flow
+type DeviceCode struct {
+	DeviceCode      *string `json:"device_code"`
+	UserCode        *string `json:"user_code"`
+	VerificationURL *string `json:"verification_url"`
+	ExpiresIn       *int64  `json:"expires_in,string"`
+	Interval        *int64  `json:"interval,string"`
+
+	Message  *string `json:"message"` // Azure specific
+	Resource *string // store this as well, needed during token exchange for azure
```
ahmetalpbalkan: maybe have a `json:",omitempty"` on this field just to be on the safe side?

autorest/azure/devicetoken.go
ahmetalpbalkan: yeah this status code check + body close + decoding is repeated in two methods, we can use response decorators here.

autorest/azure/devicetoken.go
ahmetalpbalkan: yeah, is this per oauth2 spec? actually I don't prefer increasing intervals.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,196 @@
...
+	OAuthTokenEndpoint = "https://login.microsoftonline.com/common/oauth2/token"
+
+	// authAPIVersionQueryParamName is the name
+	authAPIVersionQueryParamName  = "api-version"
+	authAPIVersionQueryParamValue = "1.0"
+)
+
+var (
+	// ErrDeviceGeneric represents an unknown error from the token endpoint when using device flow
+	ErrDeviceGeneric = fmt.Errorf("Error while retrieving OAuth token: Unknown Error")
```
ahmetalpbalkan: we need to prefix all errors with `autorest/azure/devicetoken:` or `devicetoken:`

autorest/azure/devicetoken_test.go
ahmetalpbalkan: +1

autorest/azure/devicetoken_test.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func TestDeviceTokenReturnsErrorIfAuthorizationPending(t *testing.T) {
+	sender := mocks.NewSender()
+	sender.EmitContent(errorDeviceTokenResponse("authorization_pending"))
+	sender.EmitStatus("Bad Request", 400)
+	client := &autorest.Client{Sender: sender}
+
+	go func() {
+		time.Sleep(1 * time.Second)
```
ahmetalpbalkan: I'm usually against sleeps in tests. Instead perhaps you can create a new mock that gets a list of responses (`["authorization_pending", "access_denied"]`) and on the first request, returns the first response, on the 2nd returns the second and so on.

autorest/azure/devicetoken.go
colemickens: Yes. If the user does not wish to use the automatic looping of WaitForUserCompletion then they need these errors to compare against to know how to proceed.

autorest/azure/example/main.go
```diff
@@ -104,27 +150,92 @@ func getResourceGroups(client *autorest.Client) (*string, error) {
...
+	if spt == nil {
+		log.Println("authenticating via 'mode'", mode)
+		switch mode {
+		case "device":
+			spt, err = getSptFromDeviceFlow(applicationID, tenantID, resource)
+		case "certificate":
+			spt, err = getSptFromCertificate(applicationID, tenantID, resource, certificatePath)
+		}
+		if err != nil {
+			log.Fatalln("failed to retrieve toke:", err)
```
ahmetalpbalkan: typo

autorest/azure/devicetoken_test.go
brendandixon: Agreed. Better idea. My main point is that go-routines are not reliable (in a sense).

autorest/azure/example/main.go
```diff
@@ -17,83 +18,128 @@ import (
...
+		clientID,
 		certificate,
 		rsaPrivateKey,
 		tenantID,
 		azure.AzureResourceManagerScope)
+
+	return spt, nil
+}
+
+func getSptFromDeviceFlow(clientID, tenantID, resource string) (*azure.ServicePrincipalToken, error) {
```
ahmetalpbalkan: Hmm the terminology is a bit confusing to me. Even though we do device flow, is what we get back a ServicePrincipalToken or just an oauth2.Token?

autorest/azure/devicetoken.go
colemickens: The spec only says: "The client is polling too quickly and should back off at a reasonable rate."

FWIW, even at an absurd rate, I was not able to get Azure to return this error.

Looking though, the wait duration grows by `(intervalDuration / 2)` and intervalDuration is constant as derived from the server's specified interval rate. So the wait period increasing at a constant rate.

With those in mind do we still want to change this or no?

autorest/azure/devicetoken_test.go
colemickens: String matching? Or export all the errors?

autorest/azure/persist_test.go
ahmetalpbalkan: @brendandixon that's idiomatic, I don't see a problem with that. t.Fatal/t.Error difference is exactly for this purpose.

autorest/azure/example/main.go
colemickens: Yes. It's the xplat client id. I debated whether or not to leave it in. It's makes the example easier to use, but I'm fine with removing it.

autorest/azure/devicetoken.go
ahmetalpbalkan: let's keep as is. my bar here is that after a minute, this should not be turning into something that polls every 10+ seconds. Anything in 5-10 secs range is reasonable to me.

autorest/azure/token.go
```diff
@@ -215,23 +256,25 @@ func (spt *ServicePrincipalToken) Refresh() error {
...
 	v := url.Values{}
 	v.Set("client_id", spt.clientID)
-	v.Set("grant_type", "client_credentials")
 	v.Set("resource", spt.resource)
 
-	err := spt.secret.SetAuthenticationValues(spt, &v)
-	if err != nil {
-		return err
+	if spt.RefreshToken != "" {
+		v.Set("grant_type", OAuthGrantTypeRefreshToken)
```
colemickens: On one hand, I could have a flag, or do a type assertion. On the other hand, I'd prefer to use refreshtoken anytime it's available and it will always be valid to do so. I lean towards leaving as-is.

autorest/azure/devicetoken_test.go
ahmetalpbalkan: these (a few tests above/below) will be expecting of those declared errors, right? if so `==` should do fine.

autorest/azure/token.go
```diff
@@ -168,12 +191,24 @@ func NewServicePrincipalTokenWithSecret(id string, tenantID string, resource str
 		refreshWithin: defaultRefresh,
 		sender:        &http.Client{},
 	}
-	return spt, nil
+	return spt
+}
+
+// NewServicePrincipalTokenFromManualToken creates a ServicePrincipalToken using the supplied token
+func NewServicePrincipalTokenFromManualToken(id string, tenantID string, resource string, token Token) *ServicePrincipalToken {
```
ahmetalpbalkan: I think this method can take the `refreshCallback` value and store it internally. I'm not sure the `SetRefreshCallback` setter will be called twice in any scenario. Also by making it settable only in the constructor you make it concurrency-safe (in other words, Go assignments are not atomic).

autorest/azure/devicetoken_test.go
colemickens: Yes, I much prefer this. Thanks!

autorest/azure/devicetoken_test.go
brendandixon: Agreed.

autorest/azure/example/main.go
colemickens: You get back an SPT. That's how you get autorefreshing, etc. Also, SPT implements the interface needed to be assigned to `autorest.Client`'s `Authorizer` field

colemickens: An SPT is just a convenience wrapper around a token. `getSptFromDeviceFlow` gets a token, and then creates the SPT wrapper and returns it.

autorest/azure/example/main.go
```diff
@@ -104,27 +150,92 @@ func getResourceGroups(client *autorest.Client) (*string, error) {
...
-	contentsString := string(contents)
+	var names = make([]string, len(value.ResourceGroups))
+	for i, name := range value.ResourceGroups {
+		names[i] = name.Name
+	}
+	return names, nil
+}
 
-	return &contentsString, nil
+func saveToken(spt azure.Token) {
```
ahmetalpbalkan: I think a refreshfunc should return an error. So I think you should define

    type TokenRefreshCallback func (azure.Token) error

and use that in the `SetRefreshCallback()`. I think the error returned from the callback (i.e. failure to save the token to a file for instance) can be handled by the `Refresh()` function in the autorest ––simply because when such a case happens, autorest is the caller and it is the package being executed.

Otherwise your only option in this callback is to a) panic b) log.Fatal/os.Exit(1). Neither are good nor idiomatic.

autorest/azure/token.go
colemickens: That means ~3 extra constructors then. Duplicates of the existing constructors with a suffix that indicates they take a refreshCallback. Or it means breaking the existing constructors.

I agree that the method either needs to be removed or locking added. I'm just not sure which.

autorest/azure/example/main.go
ahmetalpbalkan: ok I guess I'm not really familiar with SPT terminology at all. I hope our users are not as ignorant as I am. :)

autorest/azure/token.go
ahmetalpbalkan: hmm not sure why you would need new constructors? can't we just extend them with an extra nillable `refreshFunc` argument

autorest/azure/devicetoken_test.go
colemickens: Whoops, didn't realize it was an exported one. Those should certainly be tested. Sure, there are other error tests that do not refer to exported errors. Do we want assertions on substrings for those? Or just these since they're exported?

autorest/azure/devicetoken_test.go
colemickens: ignoring this in favor of Ahmet's solution and removing the sleep entirely.

autorest/azure/devicetoken_test.go
colemickens: (ditto as well) "ignoring this in favor of Ahmet's solution and removing the sleep entirely."

autorest/azure/token.go
colemickens: Those would be breaking. @brendandixon noted that they can be added in a backward compatible member if it is a variadic parameter.

autorest/azure/devicetoken_test.go
colemickens: Actually, I've already used the typed errors in places where they are returned. The only places that just check for non-nil would have to use sub string matching. I was doing it by hand last night, I don't mind writing them.

I just don't know if that's idiomatic or desirable?

brendandixon: At present the test could be pass and the code still incorrect (e.g., the code returns the wrong error). That's the hole I believe we should close.

colemickens: Agree. Looking for a push towards substr matching or exporting more errors.

brendandixon: There's no need to export the errors since test code resides in the same package. However, I'm fine with substr checks as well.

ahmetalpbalkan: @brendandixon I think exporting errors is meant for if someone else were to implement their own WaitForDeviceToken logic, they could use these exported errors in their state machine.

autorest/azure/devicetoken.go
colemickens: Per your other comment and writing a bit of it, I'm not going to implement this for now. One, Azure times out anyway. Two, I don't think it's undesirable to loop forever, if anything, there would be a user configurable timeout that would be passed in and respected. Otherwise the user can just call CheckFor... and do whatever they want.

autorest/azure/example/main.go
colemickens: What would you expect Refresh() to do with an error here? does it return it all the way out and make Refresh() error (which will kill off your in-flight request) ?

ahmetalpbalkan: sounds reasonable to me.

ahmetalpbalkan: especially because in the callback I can't handle the error fine because this is all single-threaded code and the code is currently executing the request. so if I won't be able to save the token, the request might as well fail.

colemickens:
I've now either replied to, or fixed all issues that have been brought up until now.

[Code coverage is higher](https://s3.amazonaws.com/colemickens-random/devicetoken-iter2.html) (still a few branches that I can't see an easy way to test)

RE: Body Closing: I added some `defer resp.Body.Close()` and then added tests for devicetoken that ensures the response is always closed. I also added another in token_test.go as an extra check that ensures that after `spt.Refresh()` is complete, we've closed the body.

Errors and test errors are prefixed now in accordance with how other packages are doing so.

Only outstanding issues:

1. Is it okay to leave xplat client id in example?

1. Do we split this into a separate package as @ahmetalpbalkan suggested? I think it's fine in here. There's basically no way to use it without the autorest package. While it's not explicitly tied to the autorest/azure package, I can't imagine any time it would *ever* be used without it.

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,200 @@
...
+	TraceID          *string `json:"trace_id,omitempty"`
+}
+
+// DeviceToken is the object return by the token exchange endpoint
+// It can either look like a Token or an ErrorToken, so put both here
+// and check for presence of "Error" to know if we are in error state
+type DeviceToken struct {
+	Token
+	TokenError
+}
```
brendandixon: @colemickens This union-that's-not-a-union bothers me. Is there a benefit in returning the TokenError within the DeviceToken struct? If not, why not define TokenError such that it implements the error interface, then return nil for DeviceToken and the new, improved TokenError as the error?

autorest/azure/devicetoken.go
```diff
@@ -0,0 +1,200 @@
...
+		}),
+		autorest.WithQueryParameters(map[string]interface{}{
+			authAPIVersionQueryParamName: authAPIVersionQueryParamValue,
+		}))
+
+	resp, err := client.Send(req)
+	if err != nil {
+		return nil, fmt.Errorf("%s %s: %s", logPrefix, errCodeSendingFails, err)
+	}
+	defer resp.Body.Close()
```
brendandixon: The defer rest.Body.Close() here, and anywhere followed by a Responder chain that includes ByClosing, is unnecessary. ByClosing *always* closes the Body. The code need only close the body  *if* it does not invoke a Responder chain that includes a ByClosing. Of course, then I'll ask why it's not using a Responder chain...

colemickens:
Removed the defers and made the union struct private.

I was going to add one more commit - adding `v3.1.0` to the Changelog... but I noticed that `autorest/version.go` still says `v1.1.1`. Should I go ahead and update `version.go` to 3.1.0 or is `version.go` for something else?

brendandixon:
@colemickens Yes, please update it. Thank you.

colemickens:
Version updated. (And fixed a small break that I'd introduced to a public api).

brendandixon:
LGTM.

ahmetalpbalkan:
LGMT too, we can iterate later. I still couldn't find time to integrate this, I will today.

colemickens:
I've added one more commit: https://github.com/colemickens/go-autorest/commit/86b3e80284bc12881a55facbbf65d837496b97b7

We back off with `waitDuration += waitDuration`. If the service tells us to slow down more than 2 times, we will fail and return an error.

@ahmetalpbalkan, @brendandixon, we okay with this?

brendandixon:
LGTM

PR #27, opened 2016-01-27
Ensured Azure service errors are unwrapped

brendandixon:
Restructured Azure and Autorest errors to ensure they are not unnecessarily wrapped.

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -27,20 +27,57 @@ const (
 	HeaderRequestID = "x-ms-request-id"
 )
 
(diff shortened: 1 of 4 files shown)


ahmetalpbalkan:
Now that we're changing the returned error type constantly we shall provide an IsAzureRequest() method (just like `os.IsNotExistError`) in which we tell the user, _yes, this was an azure response error_.

brendandixon:
Done.

Brendan Dixon
Microsoft Azure Linux, Principal SDE
425.922.8798 — @brendandixon

PR #29, opened 2016-01-29
azure: Fix issues with persist logic

ahmetalpbalkan:
- I noticed we're using math/rand without seeding to create a temp file.
  In case of multiple writers, they will create the same file path and
  will corrupt that tmp file, hence defeat the purpose. Replaced that
  with ioutil.TempFile, which is designed for this purpose.

- Added mode argument to SaveToken as tokens were saved under 0644
  mode which can be read by all users on the system. This lets setting
  perms such as 0600.

- TestSaveTokenFailsCantCreate was testing the wrong predicate. Fixed.

cc:@colemickens

### DIFF
--- a/autorest/azure/example/main.go
+++ b/autorest/azure/example/main.go
@@ -174,7 +174,7 @@ func getResourceGroups(client *autorest.Client) ([]string, error) {
 }
 
 func saveToken(spt azure.Token) {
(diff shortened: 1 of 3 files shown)


colemickens:
LGTM. Thanks for the fixes!

brendandixon:
LGTM

PR #30, opened 2016-01-29
Added test to ensure NewClientWithUserAgent never modifies DefaultClient

brendandixon:

### DIFF
--- a/autorest/client_test.go
+++ b/autorest/client_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"net/http"
 	"reflect"
 	"testing"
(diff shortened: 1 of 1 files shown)


autorest/client_test.go
```diff
@@ -595,3 +605,13 @@ func TestResponseGetPollingLocation(t *testing.T) {
 			mocks.TestURL, ar.GetPollingLocation())
 	}
 }
+
+func randomString(n int) string {
+	const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+	rand.Seed(time.Now().UTC().UnixNano())
```
ahmetalpbalkan: don't seed the global rand never ever. you're making other tests running in the same process which might be relying on the fixed seed of the global rand. if you want to construct a rand instance:

    r := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))

autorest/client_test.go
```diff
@@ -115,6 +116,15 @@ func TestNewClientWithUserAgent(t *testing.T) {
 	}
 }
 
+func TestNewclientWithUserAgentDoesNotModifyDefaultClient(t *testing.T) {
```
ahmetalpbalkan: Newclient vs NewClient

brendandixon:
@ahmetalpbalkan Fixed. Changed.

PR #32, opened 2016-02-04
Allow user to specify tenant id during device flow

colemickens:
This PR is meant to address issues affecting the reliability of device auth flow across a variety of scenarios (various combinations of apps, app owners, tenants, org/personal accounts).

This makes the device token auth flow a bit more generic. I was going to implement an alternative approach, but this is actually fine. The consumer can do what they like:

1. If they were, say interested in creating a really user friendly tool that didn't require knowledge of a subscriptionID ahead of time, they can get a `common` access tokens, list tenants, and refresh the `common` access token for a `tenant-specific` access token.

2. On the other hand, if the consumer knows `tenantID` they can go directly to the correct authority and skip the refresh orchestration (which is what was suggested to us by AAD, though I'm still not sure I agree. We document the common endpoint and we document the use of multiresource tokens granted by the common endpoint).

3. Or if the consumer has only the subscriptionID, they can attempt to get the subscription which will then contain a WWW-Authenticate header that includes a link to the correct authority and tenant specfic oauth endpoint. Then this tenantID can be used as in option 2.

I've tested this with: my personal MSA, my personal subscription's tenant's organizational user, my Microsoft.com organizational user.

@ahmetalpbalkan can you pull this branch and try it with my client id, and with your personal and then microsoft.com accounts? Thanks.

### DIFF
--- a/autorest/azure/devicetoken.go
+++ b/autorest/azure/devicetoken.go
@@ -13,16 +13,17 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/Azure/go-autorest/autorest"
 )
 
 const (
-	// OAuthDeviceEndpoint is Azure's OAuth2 Device Flow Endpoint
-	OAuthDeviceEndpoint = "https://login.microsoftonline.com/common/oauth2/devicecode"
-	// OAuthTokenEndpoint is Azure's OAuth2 Token Endpoint
-	OAuthTokenEndpoint = "https://login.microsoftonline.com/common/oauth2/token"
+	// OAuthDeviceEndpointTemplate is Azure's OAuth2 Device Flow Endpoint
+	OAuthDeviceEndpointTemplate = "https://login.microsoftonline.com/{tenantId}/oauth2/devicecode"
+	// OAuthTokenEndpointTemplate is Azure's OAuth2 Token Endpoint
+	OAuthTokenEndpointTemplate = "https://login.microsoftonline.com/{tenantId}/oauth2/token"
 
 	// authAPIVersionQueryParamName is the name
 	authAPIVersionQueryParamName  = "api-version"
@@ -63,7 +64,9 @@ type DeviceCode struct {
 	Interval        *int64  `json:"interval,string,omitempty"`
 
 	Message  *string `json:"message"` // Azure specific
-	Resource *string // store this as well, needed during token exchange for azure
+	Resource string  // store the following, stored when initiating, used when exchanging
+	ClientID string
+	TenantID string
 }
 
 // TokenError is the object returned by the token exchange endpoint
@@ -86,12 +89,14 @@ type deviceToken struct {
 
 // InitiateDeviceAuth initiates a device auth flow. It returns a DeviceCode
 // that can be used with CheckForUserCompletion or WaitForUserCompletion.
-func InitiateDeviceAuth(client *autorest.Client, clientID, resource string) (*DeviceCode, error) {
+func InitiateDeviceAuth(client *autorest.Client, clientID, tenantID, resource string) (*DeviceCode, error) {
+	oAuthDeviceEndpoint := strings.Replace(OAuthDeviceEndpointTemplate, "{tenantId}", tenantID, -1)
+
 	req, _ := autorest.Prepare(
 		&http.Request{},
 		autorest.AsPost(),
 		autorest.AsFormURLEncoded(),
-		autorest.WithBaseURL(OAuthDeviceEndpoint),
+		autorest.WithBaseURL(oAuthDeviceEndpoint),
 		autorest.WithFormData(url.Values{
 			"client_id": []string{clientID},
 			"resource":  []string{resource},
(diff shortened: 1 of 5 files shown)


autorest/azure/devicetoken.go
```diff
@@ -86,12 +89,14 @@ type deviceToken struct {
 
 // InitiateDeviceAuth initiates a device auth flow. It returns a DeviceCode
 // that can be used with CheckForUserCompletion or WaitForUserCompletion.
-func InitiateDeviceAuth(client *autorest.Client, clientID, resource string) (*DeviceCode, error) {
+func InitiateDeviceAuth(client *autorest.Client, clientID, tenantID, resource string) (*DeviceCode, error) {
+	oAuthDeviceEndpoint := strings.Replace(OAuthDeviceEndpointTemplate, "{tenantId}", tenantID, -1)
```
ahmetalpbalkan: I think we should either make it `%s` or text/template. string.Replace seems very hacky.

colemickens: Agreed. I'll fix this in a few places in another commit.

autorest/azure/example/main.go
```diff
@@ -41,29 +43,35 @@ func init() {
...
 	if mode == "device" && applicationID == "" {
+		log.Println("Using device mode auth. Will use `azkube` clientID since none was specified on the comand line.")
 		applicationID = nativeAppClientID
 	}
 
 	if mode == "certificate" && strings.Trim(certificatePath, " ") == "" {
 		log.Fatalln("Bad usage. Mode 'certificate' requires the 'certificatePath' argument.")
 	}
+
+	if strings.Trim(tenantID, " ") == "" || strings.Trim(subscriptionID, " ") == "" || strings.Trim(applicationID, " ") == "" {
```
ahmetalpbalkan: `strings.TrimSpace`.

autorest/azure/example/main.go
```diff
@@ -120,29 +128,37 @@ func getSptFromCertificate(clientID, tenantID, resource, certicatePath string, c
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to start device auth flow: %s", err)
 	}
 
 	fmt.Println(*deviceCode.Message)
 
-	token, err := azure.WaitForUserCompletion(oauthClient, clientID, deviceCode)
+	if impatient {
+		var i int64 = 1
+		deviceCode.Interval = &i
```
ahmetalpbalkan: you can do `to.Int64Ptr(1)`

autorest/azure/example/main.go
```diff
@@ -167,22 +183,26 @@ func getresourceGroups(client *autorest.Client) ([]string, error) {
...
 
-	var names = make([]string, len(value.ResourceGroups))
+	var groupNames = make([]string, len(value.ResourceGroups))
 	for i, name := range value.ResourceGroups {
-		names[i] = name.Name
+		groupNames[i] = name.Name
 	}
-	return names, nil
+
+	log.Println("Groups:", strings.Join(groupNames, ","))
```
ahmetalpbalkan: nit: comma+space would be more readable

autorest/azure/example/main.go
```diff
@@ -41,29 +43,35 @@ func init() {
 	flag.StringVar(&tenantID, "tenantId", "", "tenant id")
 	flag.StringVar(&subscriptionID, "subscriptionId", "", "subscription id")
 	flag.StringVar(&tokenCachePath, "tokenCachePath", "", "location of oauth token cache")
-	flag.BoolVar(&forceRefresh, "forceRefresh", false, "pass true to force a token refresh")
+	flag.BoolVar(&forceRefresh, "forceRefresh", true, "pass true to force a token refresh")
```
ahmetalpbalkan: is forceRefresh change intentional?

autorest/azure/example/main.go
colemickens: I'm going to remove this, this doesn't need to be encouraged in the example.

autorest/azure/example/main.go
colemickens: Not particularly. Reverted it.

autorest/azure/example/main.go
colemickens: Fixed.

autorest/azure/example/main.go
colemickens: Thanks

colemickens:
I've addressed the comments. Did you get a chance to try it?

ahmetalpbalkan:
@colemickens I started integrating. If you want to fix ahead of time, we should not hardcode the AD URLs, for instance China has `https://login.chinacloudapi.cn` and in docker-machine I can store a list of such base URLs and give one to initiate device auth.

Therefore

    OAuthTokenEndpointTemplate = "https://login.microsoftonline.com/{tenantId}/oauth2/token"

should be just like:

    OAuthTokenEndpointTemplate = "{tenantURL}/oauth2/token"

on your end because I get the tenantURL from `authorization_uri` in the header:

    WWW-Authenticate: Bearer authorization_uri="https://login.windows.net/996fe9d1-6171-40aa-945b-4c64b63bf655", error="invalid_token", error_description="The authentication failed because of missing 'Authorization' header."

ahmetalpbalkan:
...or we can have constants in the `go-autorest/azure` pkg such as:

```go
const (
    AzureCloudAuthDomain = `https://login.microsoftonline.com`
    AzureChinaCloudAuthDomain = `https://login.windows.cn`
    AzureUSGovernmentAuthDomain = `https://login.microsoftonline.com`
)
```

colemickens:
(Just on the top, I'm confused -- first comment says "login.chinacloudapi.cn", second comment says "login.windows.cn"? Is one auth and one token? Typo?)

I can model this the same way the oauth2 package does it. They're treated as pairs of endpoints generally. (The auth/token endpoints could be on different domains, for example) and the user passes a Config object that contains those two urls when they initiate the process.)

Because this is also going to need to be changed in `token.go` where the refresh code lives. Let me see what this change looks like, I'll do it in a separate commit at the very least.

ahmetalpbalkan:
it's https://login.chinacloudapi.cn/, I don't know where that came from lol. OK thanks, I'll wait.

ahmetalpbalkan:
LGTM it works. 

The only changes I needed to do was to give an actual tenant ID to `InitiateDeviceAuth` and remove `clientID` argument from `azure.WaitForUserCompletion`.

In a separate commit we should probably address taking the full `authorization_uri` to  `InitiateDeviceAuth`  (because China).

PR #33, opened 2016-02-08
Handle Azure long-running operations

brendandixon:

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -10,11 +10,16 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/Azure/go-autorest/autorest"
 )
 
 const (
+	// HeaderAsyncOperation is the Azure header containing the location to poll for long-running
+	// operations.
+	HeaderAsyncOperation = "Azure-AsyncOperation"
+
 	// HeaderClientID is the Azure extension header to set a user-specified request ID.
 	HeaderClientID = "x-ms-client-request-id"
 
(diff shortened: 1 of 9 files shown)


autorest/azure/azure.go
```diff
@@ -126,6 +131,58 @@ func ExtractRequestID(resp *http.Response) string {
...
 }
 
+// GetAsyncOperation retrieves the long-running URL to poll from the passed response.
+func GetAsyncOperation(resp *http.Response) string {
+	return resp.Header.Get(http.CanonicalHeaderKey(HeaderAsyncOperation))
+}
+
+// ResponseIsLongRunning returns true if the passed response is for an Azure long-running operation.
+func ResponseIsLongRunning(resp *http.Response) bool {
+	return autorest.ResponseRequiresPolling(resp, http.StatusCreated) && len(GetAsyncOperation(resp)) > 0
```
ahmetalpbalkan: I think we ought to handle both `http.StatusCreated` (201)  and `http.StatusAccepted` (202) for long-running operations.

`Azure-AsyncOperation` header is present in both response status codes (unless the operation is completed right away sometimes with HTTP 201, then we should not poll). 

Also let's use `== ""` comparison for strings.

autorest/azure/azure.go
```diff
@@ -126,6 +131,58 @@ func ExtractRequestID(resp *http.Response) string {
...
+}
+
+// NewAsyncPollingRequest allocates and returns a new http.Request to poll an Azure long-running
+// operation. If it successfully creates the request, it will also close the body of the passed
+// response, otherwise the body remains open.
+func NewAsyncPollingRequest(resp *http.Response, c autorest.Client) (*http.Request, error) {
```
ahmetalpbalkan: Where is this going to get called (sdk?)

autorest/azure/azure.go
brendandixon: 202 is handled by other code and, as per the current Azure REST specifications, relies on the Location header. If that's changed, then we can adapt this code.

autorest/azure/azure.go
brendandixon: Not to worry. I sent mail outlining the full scope of changes to Garima.

autorest/azure/azure.go
ahmetalpbalkan: @brendandixon hmm right. Although, in this case Azure APIs (specifically Storage RP) that return both `Location` and `Azure-AsyncOperation` header send different URLs in it. So there may be a behavioral difference. I'd say if the endpoint is specified as long running operation in Swagger, let's handle it in one place.

But before doing that, let's give it a try like this, I think even though these headers point to different URLs, they behaviorally do the same thing.

autorest/azure/azure.go
ahmetalpbalkan: alright, I appreciate if you can fwd.

autorest/azure/azure.go
brendandixon: We must follow the specification. The last time I read it, it outlined the implemented behavior. If that's changed, I want to see the update before I make code changes.

ahmetalpbalkan:
LGTM feel free to merge.

PR #35, opened 2016-02-18
Added RepondDecorators to unmarshall primitive types

brendandixon:
Added RepondDecorators to unmarshall primitive types
Corrected invocation of Prepare inspector / authorizer

### DIFF
--- a/autorest/mocks/helpers.go
+++ b/autorest/mocks/helpers.go
@@ -11,7 +11,7 @@ const (
 	TestAuthorizationHeader = "BEARER SECRETTOKEN"
 
 	// TestBadURL is a malformed URL
-	TestBadURL = ""
+	TestBadURL = "                               "
 
 	// TestDelay is the Retry-After delay used in tests.
 	TestDelay = 0 * time.Second
--- a/autorest/responder.go
+++ b/autorest/responder.go
@@ -97,6 +97,90 @@ func ByClosingIfError() RespondDecorator {
 	}
 }
 
+// ByUnmarshallingBool returns a RespondDecorator that decodes the http.Response Body into a bool
+// pointed to by b.
+func ByUnmarshallingBool(b *bool) RespondDecorator {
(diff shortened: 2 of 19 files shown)


autorest/responder.go
```diff
@@ -97,6 +97,90 @@ func ByClosingIfError() RespondDecorator {
 	}
 }
 
+// ByUnmarshallingBool returns a RespondDecorator that decodes the http.Response Body into a bool
+// pointed to by b.
+func ByUnmarshallingBool(b *bool) RespondDecorator {
```
ahmetalpbalkan: what's the use case for these responders?

autorest/mocks/helpers.go
```diff
@@ -11,7 +11,7 @@ const (
 	TestAuthorizationHeader = "BEARER SECRETTOKEN"
 
 	// TestBadURL is a malformed URL
-	TestBadURL = ""
+	TestBadURL = "                               "
```
ahmetalpbalkan: `strings.Repeat`

autorest/responder.go
brendandixon: Swagger expressly allows for primitive return types. Go's JSON decoder does not correctly handle those cases. We need to explicitly catch and convert using these responders.

autorest/mocks/helpers.go
brendandixon: Whatever. Blech. Ok.

brendandixon: Actually, no can do. TestBadURL is a constant (and should be). Go disallows constant initialization from code (that is, the call to strings.Repeat is disallowed).

autorest/responder.go
ahmetalpbalkan: Ah ok.

autorest/mocks/helpers.go
ahmetalpbalkan: that's fine then.

PR #36, opened 2016-02-23
Add Azure Environments; Use OAuthConfig throughout

colemickens:
new types: 
* OAuthConfig (represents the three oauth endpoints, tenant-specific) 
* Environment (represents the various endpoints for an azure cloud [public, usgov, china]).

After this change, SPT doesn't know about tenant-id (an Azure specific thing) and instead takes an "OAuthConfig" with urls that are already tenant-specific as the user requests. This makes our client more similar to other OAuth clients. See the `example.go` to see how this changes usage.

This is breaking change, but it's also needed to support anything other than the Azure public cloud.

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -1,0 +1,111 @@
+package azure
+
+import (
+	"fmt"
+	"net/url"
+)
+
+const (
+	activeDirectoryAPIVersion = "1.0"
+)
+
+// Environment represents a set of endpoints for each of Azure's Clouds.
+type Environment struct {
+	Name                      string
+	ManagementPortalURL       string
+	PublishSettingsURL        string
+	ServiceManagementEndpoint string
+	ResourceManagerEndpoint   string
+	ActiveDirectoryEndpoint   string
+	GalleryEndpoint           string
+	KeyVaultEndpoint          string
+	GraphEndpoint             string
+	StorageEndpointSuffix     string
+	SQLDatabaseDNSSuffix      string
+	TrafficManagerDNSSuffix   string
+	KeyVaultDNSSuffix         string
+	ServiceBusEndpointSuffix  string
+}
+
+var (
+	// PublicCloud is the default public Azure cloud environment
+	PublicCloud = Environment{
+		Name:                      "AzurePublicCloud",
+		ManagementPortalURL:       "https://manage.windowsazure.com/",
+		PublishSettingsURL:        "https://manage.windowsazure.com/publishsettings/index",
+		ServiceManagementEndpoint: "https://management.core.windows.net/",
+		ResourceManagerEndpoint:   "https://management.azure.com/",
+		ActiveDirectoryEndpoint:   "https://login.microsoftonline.com/",
+		GalleryEndpoint:           "https://gallery.azure.com/",
+		KeyVaultEndpoint:          "https://vault.azure.net/",
+		GraphEndpoint:             "https://graph.windows.net/",
+		StorageEndpointSuffix:     "core.windows.net",
+		SQLDatabaseDNSSuffix:      "database.windows.net",
+		TrafficManagerDNSSuffix:   "trafficmanager.net",
+		KeyVaultDNSSuffix:         "vault.azure.net",
+		ServiceBusEndpointSuffix:  "servicebus.azure.com",
+	}
+
+	// USGovernmentCloud is the cloud environment for the US Government
+	USGovernmentCloud = Environment{
+		Name:                      "AzureUSGovernmentCloud",
+		ManagementPortalURL:       "https://manage.windowsazure.us/",
+		PublishSettingsURL:        "https://manage.windowsazure.us/publishsettings/index",
+		ServiceManagementEndpoint: "https://management.core.usgovcloudapi.net/",
+		ResourceManagerEndpoint:   "https://management.usgovcloudapi.net",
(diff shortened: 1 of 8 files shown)


colemickens:
cc @ahmetalpbalkan

autorest/azure/environments.go
```diff
@@ -0,0 +1,99 @@
...
+
+var (
+	// PublicCloud is the default public Azure cloud environment
+	PublicCloud = Environment{
+		Name:                      "AzurePublicCloud",
+		ManagementPortalURL:       "https://manage.windowsazure.com/",
+		PublishSettingsURL:        "https://manage.windowsazure.com/publishsettings/index",
+		ServiceManagementEndpoint: "https://management.core.windows.net/",
+		ResourceManagerEndpoint:   "https://management.azure.com/",
+		ActiveDirectoryEndpoint:   "https://login.microsoftonline.com/?api-version=1.0",
```
ahmetalpbalkan: I think the `api-version` should not be in this string as we never actually query https://login.microsoftonline.com/?api-version=1.0 so it makes sense to extract api-version into a separate const.

autorest/azure/environments.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+// OAuthConfigForTenant returns an OAuthConfig wired for the given tenant
+func (env Environment) OAuthConfigForTenant(tenantID string) (*OAuthConfig, error) {
+	parsedURL, err := url.Parse(env.ActiveDirectoryEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to construct oauth configuration. Error parsing base URL: %q", err)
+	}
+	authorizeURL, tokenURL, deviceCodeURL := *parsedURL, *parsedURL, *parsedURL
+	authorizeURL.Path = path.Join(authorizeURL.Path, tenantID, "oauth2", "authorize")
```
ahmetalpbalkan: If you keep ?api-version=1.0 out of the base URL you can do this:

```go
u, _ := url.Parse(e.ActiveDirectoryEndpoint)
authorizeURL = u.Parse(fmt.Sprintf("%s/oauth2/authorize?api-version=%s", tenant, apiVersion)
tokenURL = ...
deviceURL = ...
```

also usage of `"path"` package here seemed a bit confusing, https://golang.org/pkg/net/url/#URL.Parse this method will do the relative URL path concating for you.

autorest/azure/environments.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"fmt"
+	"net/url"
+	"path"
+)
+
+// Environment represents a set of endpoints for each of Azure's Clouds.
+type Environment struct {
+	Name                      string
+	ManagementPortalURL       string
```
ahmetalpbalkan: I highly doubt we need ManagementPortalURL and PublishSettingsURL. they're browser endpoints. OTOH `StorageEndpointSuffix` now has duplication with `azure-sdk-for-go/storage` package, which actually doesn't use autorest, but assuming other people may use it, so we can keep that one.

autorest/azure/environments.go
colemickens: Agreed, fixed. I didn't realize that url.URL had it's own Parse method. Definitely is cleaner now.

autorest/azure/environments.go
colemickens: Fixed.

autorest/azure/environments.go
colemickens: I'd prefer to leave it. The .NET SDK specifies it. I can maybe imagine a scenario of wanting to link to a template vm deployment and giving the Portal URI for it. At the very least, I don't think it hurts anyone :)

autorest/azure/environments.go
```diff
@@ -0,0 +1,99 @@
...
+		StorageEndpointSuffix:     "core.chinacloudapi.cn",
+		SQLDatabaseDNSSuffix:      "database.chinacloudapi.cn",
+		TrafficManagerDNSSuffix:   "trafficmanager.cn",
+		KeyVaultDNSSuffix:         "vault.azure.net",
+		ServiceBusEndpointSuffix:  "servicebus.chinacloudapi.net",
+	}
+)
+
+// OAuthConfigForTenant returns an OAuthConfig with tenant specific urls
+func (env Environment) OAuthConfigForTenant(tenantID string) (*OAuthConfig, error) {
```
ahmetalpbalkan: this doesn't return error anymore

brendandixon:
@colemickens I like the change. Good general approach. I'll let you and @ahmetalpbalkan work out the details. But, other than that, LGTM.

autorest/azure/environments.go
```diff
@@ -0,0 +1,99 @@
...
+	template := "%s/oauth2/%s?api-version=%s"
+	u, _ := url.Parse(env.ActiveDirectoryEndpoint)
```
ahmetalpbalkan: I know it makes the code ugly but we should not ignore the errors.

brendandixon: I tend to agree.

colemickens: Agreed. Wasn't thinking about us taking user input (tenantID). Updated.

ahmetalpbalkan: url.Parse's error is still ignored.

colemickens:
Thanks for the catch @ahmetalpbalkan. Fixed in latest commit.

ahmetalpbalkan:
LGTM

PR #37, opened 2016-02-25
azure/token.go: Add RefreshExchange for Azure

colemickens:
This is needed to acquire access tokens to different `resource`s without re-initializing the auth flow each time. (Not so important for clientsecret/cert auth, but very important for device flow).

### DIFF
--- a/autorest/azure/token.go
+++ b/autorest/azure/token.go
@@ -263,9 +263,18 @@ func (spt *ServicePrincipalToken) InvokeRefreshCallbacks(token Token) error {
 
 // Refresh obtains a fresh token for the Service Principal.
 func (spt *ServicePrincipalToken) Refresh() error {
+	return spt.refreshInternal(spt.resource)
+}
+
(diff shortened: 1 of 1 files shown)


colemickens:
I'm already using this (using glide's aliasing) and it's working as expected.

ahmetalpbalkan:
lgtm

brendandixon:
@colemickens golint is failing. If you fix that, then I'll be Ok. :)

colemickens:
After yet another go around with AAD, I validated this for all of my scenarios. I'm starting to learn how to handle various AAD error states, so if you're ever hitting your head at an error that doesn't make any sense, or something works in one tenant and not another, I might be able to help.

In this case, device auth was "not working", because the application was not configured correctly. Reconfiguring it lead down one incorrect path with the obnoxious "Sorry, but we’re having trouble signing you in. We received a bad request." message. Finally got it configured correctly (it needed to ask for user consent to "Access Directory as Signed in User".

PR #38, opened 2016-02-26
Reworked handling asynchronous and polled requests

brendandixon:
This PR *significantly* changes how `go-autorest` handles polling. By using `SendDecorators` it can now address specific cases (such as handling Azure long-running operations).

### DIFF
--- a/autorest/autorest.go
+++ b/autorest/autorest.go
@@ -75,47 +75,14 @@ func ResponseHasStatusCode(resp *http.Response, codes ...int) bool {
 	return containsInt(codes, resp.StatusCode)
 }
 
-// ResponseRequiresPolling returns true if the passed http.Response requires polling follow-up
-// request (as determined by the status code being in the passed set, which defaults to HTTP 202
-// Accepted).
-func ResponseRequiresPolling(resp *http.Response, codes ...int) bool {
-	if resp.StatusCode == http.StatusOK {
-		return false
-	}
-
-	if len(codes) == 0 {
-		codes = []int{http.StatusAccepted}
-	}
-
-	return ResponseHasStatusCode(resp, codes...)
+// GetLocation retrieves the URL from the Location header of the passed response.
+func GetLocation(resp *http.Response) string {
+	return resp.Header.Get(HeaderLocation)
 }
 
-// NewPollingRequest allocates and returns a new http.Request to poll for the passed response. If
-// it successfully creates the request, it will also close the body of the passed response,
-// otherwise the body remains open.
-func NewPollingRequest(resp *http.Response, c Client) (*http.Request, error) {
-	location := GetPollingLocation(resp)
-	if location == "" {
-		return nil, NewErrorWithResponse("autorest", "NewPollingRequest", resp, "Location header missing from response that requires polling")
-	}
-
-	req, err := Prepare(&http.Request{},
-		AsGet(),
-		WithBaseURL(location))
-	if err != nil {
-		return nil, NewErrorWithError(err, "autorest", "NewPollingRequest", nil, "Failure creating poll request to %s", location)
-	}
-
-	Respond(resp,
-		c.ByInspecting(),
-		ByClosing())
-
-	return req, nil
-}
-
-// GetPollingDelay extracts the polling delay from the Retry-After header of the passed response. If
+// GetRetryAfter extracts the retry delay from the Retry-After header of the passed response. If
 // the header is absent or is malformed, it will return the supplied default delay time.Duration.
-func GetPollingDelay(resp *http.Response, defaultDelay time.Duration) time.Duration {
+func GetRetryAfter(resp *http.Response, defaultDelay time.Duration) time.Duration {
 	retry := resp.Header.Get(HeaderRetryAfter)
 	if retry == "" {
 		return defaultDelay
@@ -129,38 +96,19 @@ func GetPollingDelay(resp *http.Response, defaultDelay time.Duration) time.Durat
 	return d
 }
 
-// GetPollingLocation retrieves the polling URL from the Location header of the passed response.
-func GetPollingLocation(resp *http.Response) string {
-	return resp.Header.Get(HeaderLocation)
-}
-
-// PollForAttempts will retry the passed http.Request until it receives an HTTP status code outside
-// the passed set or has made the specified number of attempts. The set of status codes defaults to
-// HTTP 202 Accepted.
-func PollForAttempts(s Sender, req *http.Request, defaultDelay time.Duration, attempts int, codes ...int) (*http.Response, error) {
-	return SendWithSender(
-		decorateForPolling(s, defaultDelay, codes...),
-		req,
-		DoRetryForAttempts(attempts, time.Duration(0)))
-}
-
-// PollForDuration will retry the passed http.Request until it receives an HTTP status code outside
-// the passed set or the total time meets or exceeds the specified duration. The set of status codes
-// defaults to HTTP 202 Accepted.
-func PollForDuration(s Sender, req *http.Request, defaultDelay time.Duration, total time.Duration, codes ...int) (*http.Response, error) {
-	return SendWithSender(
-		decorateForPolling(s, defaultDelay, codes...),
-		req,
-		DoRetryForDuration(total, time.Duration(0)))
-}
-
-func decorateForPolling(s Sender, defaultDelay time.Duration, codes ...int) Sender {
-	if len(codes) == 0 {
-		codes = []int{http.StatusAccepted}
+// NewPollingRequest allocates and returns a new http.Request to poll for the passed response.
+func NewPollingRequest(resp *http.Response, cancel <-chan struct{}) (*http.Request, error) {
+	location := GetLocation(resp)
+	if location == "" {
+		return nil, NewErrorWithResponse("autorest", "NewPollingRequest", resp, "Location header missing from response that requires polling")
 	}
 
-	return DecorateSender(s,
-		AfterRetryDelay(defaultDelay),
-		DoErrorIfStatusCode(codes...),
-		DoCloseIfError())
+	req, err := Prepare(&http.Request{Cancel: cancel},
+		AsGet(),
+		WithBaseURL(location))
+	if err != nil {
+		return nil, NewErrorWithError(err, "autorest", "NewPollingRequest", nil, "Failure creating poll request to %s", location)
+	}
+
+	return req, nil
 }
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -1,0 +1,154 @@
+package azure
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/date"
+)
+
+const (
+	// HeaderAsyncOperation is the Azure header providing the URL from which to obtain the
+	// OperationResource for an asynchronous (aka long running) operation.
+	HeaderAsyncOperation = "Azure-AsyncOperation"
+)
+
+const (
+	// OperationCanceled says the underlying operation was canceled.
+	OperationCanceled string = "Canceled"
+	// OperationFailed says the underlying operation failed.
+	OperationFailed string = "Failed"
+	// OperationSucceeded says the underlying opertion succeeded.
+	OperationSucceeded string = "Succeeded"
+)
+
+// OperationError provides additional detail when an operation fails or is canceled.
+type OperationError struct {
+	// Code provides an invariant error code useful for troubleshooting, aggregration, and analysis.
+	Code string `json:"code"`
+	// Message indicates what error occurred and what can be done to address the issue.
+	Message string `json:"message"`
+}
+
+// Error implements the error interface returnin a string containing the code and message.
+func (oe OperationError) Error() string {
+	return fmt.Sprintf("Azure Operation Error: Code=%q Message=%q", oe.Code, oe.Message)
+}
+
+// OperationResource defines a resource describing the state of a long-running operation.
+type OperationResource struct {
+	// Id is the identifier used in a GET for the underlying resource.
+	ID string `json:"id"`
+	// Name matches the last segment of the Id field (typically system generated).
+	Name string `json:"name"`
+	// Status provides the state of the operation. Non-terminal states vary by resource;
+	Status string `json:"status"`
+	// Properties, on operation success, optionally contains per-service / per-resource values.
+	Properties map[string]interface{} `json:"properties"`
+	// Error provides additional detail if the operation is canceled or failed.
+	OperationError OperationError `json:"error"`
+	// StartTime optionally provides the time the operation started.
+	StartTime date.Time `json:"startTime"`
+	// EndTime optionally provides the time the operation completed.
+	EndTime date.Time `json:"endTime"`
+	// PercentComplete optionally provides the percent complete between 0 and 100.
+	PercentComplete float64 `json:"percentComplete"`
+}
+
+// HasTerminated returns true if the operation has terminated; false otherwise.
+func (or OperationResource) HasTerminated() bool {
+	switch or.Status {
+	case OperationCanceled, OperationFailed, OperationSucceeded:
+		return true
+	default:
+		return false
+	}
+}
+
+// Error returns an error if the operation was canceled or failed and nil otherwise.
+func (or OperationResource) Error() string {
+	switch or.Status {
+	case OperationCanceled, OperationFailed:
+		return or.OperationError.Error()
+	default:
+		return ""
+	}
+}
+
+// GetAsyncOperation retrieves the long-running URL from which to retrieve the OperationResource.
+func GetAsyncOperation(resp *http.Response) string {
+	return resp.Header.Get(http.CanonicalHeaderKey(HeaderAsyncOperation))
+}
+
+// IsAsynchronousResponse returns true if the passed response indicates that the request will
+// complete asynchronously. Such responses have either an http.StatusCreated or an
+// http.StatusAccepted status code and provide the Azure-AsyncOperation header.
+func IsAsynchronousResponse(resp *http.Response) bool {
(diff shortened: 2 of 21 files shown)


autorest/autorest.go
```diff
@@ -129,38 +96,19 @@ func GetPollingDelay(resp *http.Response, defaultDelay time.Duration) time.Durat
...
+func NewPollingRequest(resp *http.Response, cancel <-chan struct{}) (*http.Request, error) {
+	location := GetLocation(resp)
+	if location == "" {
+		return nil, NewErrorWithResponse("autorest", "NewPollingRequest", resp, "Location header missing from response that requires polling")
+	}
 
-func decorateForPolling(s Sender, defaultDelay time.Duration, codes ...int) Sender {
-	if len(codes) == 0 {
-		codes = []int{http.StatusAccepted}
+	req, err := Prepare(&http.Request{Cancel: cancel},
```
ahmetalpbalkan: nice addition.

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+// OperationError provides additional detail when an operation fails or is canceled.
+type OperationError struct {
+	// Code provides an invariant error code useful for troubleshooting, aggregration, and analysis.
+	Code string `json:"code,omitempty"`
+	// Message indicates what error occurred and what can be done to address the issue.
+	Message string `json:"message,omitempty"`
+}
+
+// ToError returns an error containing the code and message.
+func (oe OperationError) ToError() error {
```
ahmetalpbalkan: I think you should just implement `func (OE) Error() string` and this would be an error. fmt.Errorf is not necessary here.

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+	return or.Status == OperationCanceled
+}
+
+// HasFailed returns true if the operation has failed; false otherwise.
+func (or OperationResource) HasFailed() bool {
+	return or.Status == OperationFailed
+}
+
+// HasSucceeded returns true if the operation has succeeded; false otherwise.
+func (or OperationResource) HasSucceeded() bool {
```
ahmetalpbalkan: I feel like these 3 exported methods are not so much necessary as they're only used in `HasTerminated`. I think we can put all the logic into `HasTerminated` and we'd be fine.

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+
+// NewOperationResourceRequest allocates and returns a new http.Request to retrieve the
+// OperationResource for an asynchronous operation.
+func NewOperationResourceRequest(resp *http.Response, cancel <-chan struct{}) (*http.Request, error) {
+	location := GetAsyncOperation(resp)
+	if location == "" {
+		return nil, autorest.NewErrorWithResponse("azure", "NewOperationResourceRequest", resp, "Azure-AsyncOperation header missing from response that requires polling")
+	}
+
+	req, err := autorest.Prepare(&http.Request{Cancel: cancel},
```
ahmetalpbalkan: so this `req, err := ...` statement repeats with `NewPollingRequest`, perhaps extract this into an unexported method?

autorest/sender_test.go
```diff
@@ -171,102 +174,35 @@ func TestAfterDelay_Cancels(t *testing.T) {
...
+	start := time.Now()
 	r, _ := SendWithSender(client, mocks.NewRequest(),
-		AfterRetryDelay(d),
-		DoRetryForAttempts(2, time.Duration(0)))
-	s := time.Since(tt)
-	if s < d {
-		t.Error("autorest: AfterRetryDelay failed to wait for at least the specified duration")
+		AfterDelay(d))
+
+	if time.Now().Sub(start) > (5 * d) {
```
ahmetalpbalkan: `time.Since`

autorest/sender_test.go
```diff
@@ -554,25 +483,145 @@ func TestDelayForBackoff_Cancels(t *testing.T) {
...
 	}
 }
 
-// Disable test for TravisCI
-// func TestDelayForBackoffWithinReason(t *testing.T) {
+func TestDelayForBackoffWithinReason(t *testing.T) {
+	d := 5 * time.Millisecond
+	start := time.Now()
+	DelayForBackoff(d, 1, nil)
+	if time.Now().Sub(start) > (time.Duration(5.0) * d) {
```
ahmetalpbalkan: `time.Since` and you can just say `5*d` here I believe.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan I need to create a new Error object containing the code and message. How else would I do it?

autorest/azure/async.go
brendandixon: @ahmetalpbalkan It feels little overkill, but they capture distinct states. I am prone to leave them.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan Hmm, generally I agree with you. It's so few lines that it feels like an over DRY. I could be persuaded...

autorest/sender_test.go
brendandixon: @ahmetalpbalkan Yup, I'll change it.

autorest/sender_test.go
brendandixon: @ahmetalpbalkan Yup. Changed.

autorest/azure/async.go
ahmetalpbalkan: @brendandixon not sure what you mean. `error` is just an `interface { Error() string }`. If you implement `func Error()string` on this type, it'll already be an error and you won't need `ToError()`.

brendandixon: @ahmetalpbalkan Doh (he said sheepishly).

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+type OperationError struct {
+	// Code provides an invariant error code useful for troubleshooting, aggregration, and analysis.
+	Code string `json:"code,omitempty"`
+	// Message indicates what error occurred and what can be done to address the issue.
+	Message string `json:"message,omitempty"`
+}
+
+// Error implements the error interface returnin a string containing the code and message.
+func (oe OperationError) Error() string {
+	return fmt.Sprintf("Azure Operation Error -- code(%v), message(%v)", oe.Code, oe.Message)
```
ahmetalpbalkan: I suggest using %q for those. They'll 

a. add `"` quotes for you
b. Escape new lines etc properly so that you'll end up with a single line string. You can just say `Code=%q Message=%q`.

The communication language here with `--`s, `code(...)` etc seems a bit unusual to me. It's better to keep it simple:

    return fmt.Sprintf("Azure operation failed: Code=%q Message=%q")

is a ton more clearer and well formatted and consistent with rest of the error messages we have.

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+// HasTerminated returns true if the operation has terminated; false otherwise.
+func (or OperationResource) HasTerminated() bool {
+	return or.IsCanceled() || or.HasFailed() || or.HasSucceeded()
+}
+
+// GetError returns an error if the operation was canceled or failed and nil otherwise.
+func (or OperationResource) GetError() error {
+	switch {
+	case or.IsCanceled():
+		return fmt.Errorf("azure: Operation was canceled -- error(%s)", or.Error.Error())
```
ahmetalpbalkan: same comment as above, we should not do `--`s and `%q` is better than `error(%s)`.

autorest/azure/async_test.go
```diff
@@ -0,0 +1,516 @@
...
+
+	client := mocks.NewSender()
+	client.AppendResponse(newAsynchronousResponse())
+	client.AppendAndRepeatResponse(newOperationResourceResponse("Busy"), -1)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	start := time.Now()
+	go func() {
+		wg.Done()
```
ahmetalpbalkan: this should be `defer`?

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+	Message string `json:"message,omitempty"`
+}
+
+// Error implements the error interface returnin a string containing the code and message.
+func (oe OperationError) Error() string {
+	return fmt.Sprintf("Azure Operation Error -- code(%v), message(%v)", oe.Code, oe.Message)
+}
+
+// OperationResource defines a resource describing the state of a long-running operation.
+type OperationResource struct {
```
ahmetalpbalkan: I'd say `,omitempty`s in this type are unnecessary because we're not marshaling back. It's just for unmarshaling.

autorest/azure/async.go
ahmetalpbalkan: @brendandixon I can hardly think having this exported will be usable to anyone (in fact we are exporting a lot of symbols in this package generally). I just feel like having a single `HasTerminated` method would be easier to test and it's the only place this enum is actually used. Having all those separate methods seemed a bit overkill.

you can just do:

```go
func HasTerminated(){ 
  switch state {
    case OperationCanceled, OperationFailed, OperationSucceeded: return true
    default: return false
  }
}
```

which is also very much readable and you don't need separate methods and you know Go encourages you to keep things simple. The `HasFailed()`/`HasSucceeded()` reminds me of Java/C#.

autorest/azure/async.go
```diff
@@ -0,0 +1,165 @@
...
+	return or.IsCanceled() || or.HasFailed() || or.HasSucceeded()
+}
+
+// GetError returns an error if the operation was canceled or failed and nil otherwise.
+func (or OperationResource) GetError() error {
+	switch {
+	case or.IsCanceled():
+		return fmt.Errorf("azure: Operation was canceled -- error(%s)", or.Error.Error())
+	case or.HasFailed():
+		return fmt.Errorf("azure: Operation failed -- error(%s)", or.Error.Error())
```
ahmetalpbalkan: `...%s", or.Error.Error()` is overkill.
`azure: Operation failed: %v", or.Error` should be just enough.

ahmetalpbalkan: Also this will return:

    azure: Operation failed -- error(Azure Operation Error -- code(blah) message(blah))

it looks pretty ugly.

autorest/sender.go
```diff
@@ -168,7 +138,37 @@ func DoErrorUnlessStatusCode(codes ...int) SendDecorator {
...
 
-// DoRetryForAttempts returns a SendDecorator that retries the request for up to the specified
+// DoPollForStatusCodes returns a SendDecorator that polls if the http.Response contains one of the
+// passed status codes. It expects the http.Response to contain a Location header providing the
+// URL at which to poll (using GET) and will poll until the time passed is equal to or greater than
+// the supplied duration. It will delay between requests for the duration specified in the
+// RetryAfter header or, if the header is absent, the passed delay. Polling may be canceled by
+// closing the optional channel on the http.Request.
+func DoPollForStatusCodes(duration time.Duration, delay time.Duration, codes ...int) SendDecorator {
+	if len(codes) == 0 {
```
ahmetalpbalkan: this if check and assignment seemed a bit unnecessary. nil slice and empty slice often behaves the same.

autorest/sender.go
```diff
@@ -207,6 +207,19 @@ func DoRetryForDuration(d time.Duration, backoff time.Duration) SendDecorator {
...
 }
 
+// WithLogging returns a SendDecorator that implements simple before and after logging of the
+// request.
+func WithLogging(logger *log.Logger) SendDecorator {
+	return func(s Sender) Sender {
+		return SenderFunc(func(r *http.Request) (*http.Response, error) {
+			logger.Printf("Sending %s %s\n", r.Method, r.URL)
+			resp, err := s.Do(r)
+			logger.Printf("%s %s received %s\n", r.Method, r.URL, resp.Status)
```
ahmetalpbalkan: `resp.Status` may panic with nil dereference exception `if err != nil`

autorest/sender.go
```diff
@@ -207,6 +207,19 @@ func DoRetryForDuration(d time.Duration, backoff time.Duration) SendDecorator {
 	}
 }
 
+// WithLogging returns a SendDecorator that implements simple before and after logging of the
+// request.
+func WithLogging(logger *log.Logger) SendDecorator {
+	return func(s Sender) Sender {
+		return SenderFunc(func(r *http.Request) (*http.Response, error) {
+			logger.Printf("Sending %s %s\n", r.Method, r.URL)
```
ahmetalpbalkan: log.Printf appends the new line (`\n`) by default, you don't have to.

autorest/client.go
```diff
@@ -43,6 +28,12 @@ const (
 `
 )
 
+// Response serves as the base for all responses from generated clients. It provides access to the
+// last http.Response.
+type Response struct {
```
ahmetalpbalkan: This makes me think do we really need an `autorest.Response`? If you're defining certain methods on it, you can often just do:

    type Response *http.Response

instead of embedding it.

autorest/azure/async_test.go
brendandixon: @ahmetalpbalkan Nope. It needs to execute prior to the method leaving.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan Yah, whatever. There's a certain consistency to it all this way. :)

autorest/sender.go
brendandixon: @ahmetalpbalkan No, it does not. :)

autorest/azure/async_test.go
ahmetalpbalkan: Then I'm not clear what the intent is here. `wg.Wait()` can/will immediately unblock once this statement is executed. If you need to execute it before the goroutine ends, `defer` should be still doing what you need.

autorest/sender.go
ahmetalpbalkan: I can't repro, do you mind sending a play.golang.org snippet?

autorest/azure/async.go
brendandixon: @ahmetalpbalkan I'll clean-up the redundancy and manually add quotes -- the code and message are strings, %q works only for integers.

brendandixon: @ahmetalpbalkan Scratch that. I see that %q works for strings as well.

Regarding format: I've yet to see a clear pattern in Go. The pattern above I've used variously in the past. Happy to adjust if you feel the other is more idiomatic.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan Simplicity is good. I like the suggestion to use the switch. I'm less clear on removing the IsXxxx methods. How many symbols we're exporting does not seem to me a useful measure as there is no clear definition of what would be the right number. There are three distinct states: The two "error" states are quite different and need discrimination. Since we cannot easily discriminate with Go's builtin in error. Having the check serves a purpose. (It is better to hide string comparisons than to expose them.)

Regardless, since we're not (yet) exposing the object, I'll remove the IsXxxxx checks for now. I do not feel this is the best move.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan Agreed. Ugly. Will clean.

autorest/azure/async_test.go
brendandixon: @ahmetalpbalkan I think you may be missing the necessary execution flow.

The code tests that DoPollForAsynchronous will exit if canceled (that is, the http.Request.Cancel channel is closed). This requires that the code execute DoPollForAsynchronous on a "background thread" (as it were) while the "main thread" issues a cancel (that is, closes the channel). It uses a WaitGroup to pause the main thread so the goroutine has a chance to run. The goroutine must release the main thread *before* it enters its own "spin" loop; that is, wg.Done must called *before* SendWithSender.

I hope this clarifies the test.

autorest/azure/async.go
ahmetalpbalkan: @brendandixon I've been seeing error message patterns emerging in docker repo. It's usually like:

    package: Message field=%q..

autorest/azure/async_test.go
ahmetalpbalkan: Oh I see now, I guess I'd use channels to “pause the main thread so the goroutine has a chance to run” personally but waitgroup will do the same thing regardless.

autorest/sender.go
brendandixon: @ahmetalpbalkan I ran some tests and I (now) agree. Removing.

autorest/sender.go
brendandixon: @ahmetalpbalkan http://play.golang.org/p/waiPl1eh7T.

fmt.Printf does *not* append a newline. fmt.Println, however, does.

ahmetalpbalkan: @brendandixon you're not using `fmt` package here. logger adds new lines. https://play.golang.org/p/kLDSqpfUGk

autorest/azure/async_test.go
brendandixon: @ahmetalpbalkan That approach, in this case, requires internal knowledge of how DoPollForAsynchronous works. This approach will work in all cases.

autorest/sender.go
brendandixon: @ahmetalpbalkan Good point...and gross inconsistency in Go's documentation. The Logger documentation clearly states it behaves a fmt.Printf. Blech. I'll remove.

autorest/azure/async.go
brendandixon: @ahmetalpbalkan Removed.

autorest/sender.go
brendandixon: @ahmetalpbalkan Good catch. Fixed.

autorest/client.go
brendandixon: @ahmetalpbalkan At the time (months ago) it seemed we would need it. Now, not so much. Let's chat about the impact to the SDK.

ahmetalpbalkan:
Generally LGTM, feel free to merge once you sort out the issues as you see fit.

brendandixon:
@ahmetalpbalkan I appreciate your thoroughness.

ahmetalpbalkan:
Thanks, I'll wait @garimakhulbe to integrate this to SDK.

PR #40, opened 2016-03-03
Corrected DoPollForAsynchronous to not return an error when succeeding

brendandixon:

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -57,6 +57,11 @@ type OperationResource struct {
 	PercentComplete float64 `json:"percentComplete"`
 }
 
+// HasSucceeded returns true if the operation has succeeded; false otherwise.
+func (or OperationResource) HasSucceeded() bool {
+	return or.Status == OperationSucceeded
+}
+
 // HasTerminated returns true if the operation has terminated; false otherwise.
 func (or OperationResource) HasTerminated() bool {
 	switch or.Status {
@@ -67,13 +72,13 @@ func (or OperationResource) HasTerminated() bool {
 	}
 }
 
-// Error returns an error if the operation was canceled or failed and nil otherwise.
-func (or OperationResource) Error() string {
+// GetError returns an error if the operation was canceled or failed and nil otherwise.
+func (or OperationResource) GetError() error {
 	switch or.Status {
 	case OperationCanceled, OperationFailed:
-		return or.OperationError.Error()
+		return or.OperationError
 	default:
-		return ""
+		return nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
thanks for quick action!

ahmetalpbalkan:
LGTM. Will it return an error or `nil` when operation is Cancelled? (that said, I don't know what a cancelled operation status looks like, whether it contains an `"error"` field or not). I feel like if it doesn't have an `"error"` field but status=Cancelled, we're still returning an empty error msg?

I also think maybe this is right time to stop returning empty error message?

brendandixon:
It will return an error if the operation is canceled, as I believe it should. The error will be the OperationResource object containing whatever data it contains.

Brendan Dixon
Microsoft Azure Linux, Principal SDE
425.922.8798 — @brendandixon

ahmetalpbalkan:
@brendandixon yes, but what if Cancelled doesn't contain "error" field?

Your assumption below is, it will contain, but I think it won't. :smiley: Apparently we can't trigger this very easily so we gotta ask.

Even though it does, there may be other situations where operation was Status=Failed but API failed to provide a message. You just can't return "empty string" error for those, ever...

image

this logic should be like the following:

```go

if or.OperationError != nil { // you gotta make this field a pointer, first.
    return or.OperationError.Error()
} else {
    return fmt.Errorf("non-successful operation did not return an error. Status=%q ID=%q",  ...)
}
````

we should _never_ return an error message.

ahmetalpbalkan:
I can confirm this works, so I'm unblocked but I think we should address the emptystring-error issue.

brendandixon:
@ahmetalpbalkan If the protocol behaves, then an empty string will never be returned. I can the method to produce an error instead of being an error, but it loses a tad bit of elegance in doing so. Regardless, I'll change it.

ahmetalpbalkan:
@brendandixon even though server obeys the contract, tomorrow if some terminal state comes in that does not have an `"error"` field, we'll be returning callers an empty that looks like:

    network/SecurityGroupsClient:CreateOrUpdate 200 Failure sending request -- Original Error:


vs if we make the change:

    network/SecurityGroupsClient:CreateOrUpdate 200 Failure sending request -- Original Error: non-successful 
        operation did not return an error message. Status=InterruptedOrSomething ID=<guid>

brendandixon:
@ahmetalpbalkan The case you've identified will only occur in the code if the server returns an OperationResource object whose state is "canceled" or "failed" but lacks the "error" subfield. All other cases are correctly handled.

ahmetalpbalkan:
@brendandixon that's right, apparently error is returned if status == failed or status == canceled. regardless having an empty string error message is somewhat scary. thanks for fixing.

PR #44, opened 2016-03-17
Correct handling of malformed JSON and polling for Azure asynchronous operations

brendandixon:
Removed type-specific unmarshalling code; reverted to using ```json.Unmarshal```.
Added ```ByCopying``` responder with supporting ```TeeReadCloser```.
Rewrote Azure asynchronous handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # CHANGELOG
 
+## v7.0.0
+- Reverted to only unmarshalling JSON, corrected handling of RFC3339 time strings
+- Added ByCopying responder with supporting TeeReadCloser
+- Rewrote Azure asynchronous handling
+
+The `json.Decoder` does not catch bad data as thoroughly as `json.Unmarshal`. Since
+`encoding/json` successfully deserializes all core types, and extended types normally provide
+their custom JSON serialization handlers, the code has been reverted back to using
+`json.Unmarshal`. The original change to use `json.Decode` was made to reduce duplicate
+code; there is no loss of function, and there is a gain in accuracy, by reverting.
+
+Additionally, Azure services indicate requests to be polled by multiple means. The existing code
+only checked for one of those (that is, the presence of the `Azure-AsyncOperation` header).
+The new code correctly covers all cases and aligns with the other Azure SDKs.
+
 ## v6.1.0
 - Introduced `date.ByUnmarshallingJSONDate` and `date.ByUnmarshallingJSONTime` to enable JSON encoded values.
 
--- a/autorest/date/time.go
+++ b/autorest/date/time.go
@@ -4,6 +4,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"io/ioutil"
 	"net/http"
+	"strings"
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # CHANGELOG
 
+## v7.0.0
+- Reverted to only unmarshalling JSON, corrected handling of RFC3339 time strings
+- Added ByCopying responder with supporting TeeReadCloser
+- Rewrote Azure asynchronous handling
+
+The ```json.Decoder``` does not catch bad data as thoroughly as ```json.Unmarshal```. Since
```
ahmetalpbalkan: 1 backtick is enough here, no need for 3.

brendandixon: Habit from writing inline code blocks. :)

autorest/utility.go
```diff
@@ -48,53 +46,24 @@ func CopyAndDecode(encodedAs EncodedAs, r io.Reader, v interface{}) (bytes.Buffe
...
-func readBool(r io.Reader) (bool, error) {
-	s, err := readString(r)
-	if err == nil {
-		return strconv.ParseBool(s)
-	}
-	return false, err
+// TeeReadCloser returns a ReadCloser that writes to w what it reads from rc.
+// It utilizes io.TeeReader to copy the data read and has the same behavior when reading.
+// Further, when it is closed, it ensures that rc is closed as well.
+func TeeReadCloser(rc io.ReadCloser, w io.Writer) io.ReadCloser {
```
ahmetalpbalkan: I believe there's no need to export this identifier

ahmetalpbalkan:
generally LGTM. thanks for making PollingDelay/PollingDuration configurable.

autorest/utility.go
brendandixon: Unless you're strongly opposed, I'm going to leave it.

ahmetalpbalkan: we can leave it, it's just `newTeeReadCloser()` would be more idiomatic

brendandixon: @ahmetalpbalkan I modeled it on https://golang.org/pkg/io/#TeeReader.

autorest/date/time.go
```diff
@@ -19,7 +20,8 @@ func ParseTime(date string) (d Time, err error) {
 }
 
 func parseTime(date string, format string) (Time, error) {
-	d, err := time.Parse(format, date)
+	// Note: Convert the date first to uppercase since RFC3339 allows lower-case "t" and "z".
+	d, err := time.Parse(format, strings.ToUpper(date))
```
garimakhulbe: We need to move ToUpper() to UnmarsalJSON() because we are using that function now. 

```go
func (d *Time) UnmarshalJSON(data []byte) (err error) {
        return d.Time.UnmarshalJSON(bytes.ToUpper(data))
 }
```

brendandixon: @garimakhulbe You should move the code to the new types we discussed earlier. Ok?

garimakhulbe: Cool.

PR #46, opened 2016-03-28
Fix bugs in ByUnmarshallingJSON and GET method polling

garimakhulbe:
Bug fix in ByUnmarshallingJSON and GET method polling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v7.0.1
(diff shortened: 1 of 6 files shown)


brendandixon:
@garimakhulbe Looks good, but we should add a test to cover the case in ByUnmarshallingJSON (that is, a test which yields a Body of just blanks).

garimakhulbe:
Okay, let me add that test.

garimakhulbe:
Added test.

garimakhulbe:
Thanks.

PR #51, opened 2016-04-07
Update Error message

garimakhulbe:
Version 7.0.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## v7.0.4
+- Better error messages for long running operation failures
+
 ## v7.0.3
 - Corrected DoPollForAsynchronous to properly handle the initial response
 
--- a/autorest/azure/async_test.go
+++ b/autorest/azure/async_test.go
@@ -2,15 +2,14 @@ package azure
 
 import (
 	"fmt"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/mocks"
 	"io/ioutil"
 	"net/http"
 	"strings"
 	"sync"
 	"testing"
 	"time"
-
-	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/mocks"
 )
 
 func TestGetAsyncOperation_ReturnsAzureAsyncOperationHeader(t *testing.T) {
@@ -56,7 +55,7 @@ func TestHasTerminated_ReturnsFalseForUnknownStates(t *testing.T) {
 }
 
 func TestOperationError_ErrorReturnsAString(t *testing.T) {
-	s := (operationError{Code: "server code", Message: "server error"}).Error()
+	s := (ServiceError{Code: "server code", Message: "server error"}).Error()
 	if s == "" {
 		t.Errorf("azure: operationError#Error failed to return an error")
 	}
@@ -98,33 +97,33 @@ func TestOperationResource_HasTerminatedReturnsFalseForUnknownStates(t *testing.
 }
 
 func TestProvisioningStatus_StateReturnsState(t *testing.T) {
-	if (provisioningStatus{provisioningProperties{"state"}}).state() != "state" {
+	if (provisioningStatus{Properties: provisioningProperties{"state"}}).state() != "state" {
 		t.Errorf("azure: provisioningStatus#state failed to return the correct state")
 	}
 }
 
 func TestProvisioningStatus_HasSucceededReturnsFalseIfNotSuccess(t *testing.T) {
-	if (provisioningStatus{provisioningProperties{"not a success string"}}).hasSucceeded() {
+	if (provisioningStatus{Properties: provisioningProperties{"not a success string"}}).hasSucceeded() {
 		t.Errorf("azure: provisioningStatus#hasSucceeded failed to return false for a canceled operation")
 	}
 }
 
 func TestProvisioningStatus_HasSucceededReturnsTrueIfSuccessful(t *testing.T) {
-	if !(provisioningStatus{provisioningProperties{operationSucceeded}}).hasSucceeded() {
+	if !(provisioningStatus{Properties: provisioningProperties{operationSucceeded}}).hasSucceeded() {
 		t.Errorf("azure: provisioningStatus#hasSucceeded failed to return true for a successful operation")
 	}
 }
 
 func TestProvisioningStatus_HasTerminatedReturnsTrueForKnownStates(t *testing.T) {
 	for _, state := range []string{operationSucceeded, operationCanceled, operationFailed} {
-		if !(provisioningStatus{provisioningProperties{state}}).hasTerminated() {
+		if !(provisioningStatus{Properties: provisioningProperties{state}}).hasTerminated() {
 			t.Errorf("azure: provisioningStatus#hasTerminated failed to return true for the '%s' state", state)
 		}
 	}
 }
 
 func TestProvisioningStatus_HasTerminatedReturnsFalseForUnknownStates(t *testing.T) {
-	if (provisioningStatus{provisioningProperties{"not a known state"}}).hasTerminated() {
+	if (provisioningStatus{Properties: provisioningProperties{"not a known state"}}).hasTerminated() {
 		t.Errorf("azure: provisioningStatus#hasTerminated returned true for a non-terminal operation")
 	}
 }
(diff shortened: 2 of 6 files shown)


ahmetalpbalkan:
doesn't work:

```
Failure sending request: Status code 400 -- Original Error: Long running operation
terminated with status 'Failed': Code="" Message=""
```

@garimakhulbe Can you please write a test case mocking a proper error.code & error.message response and compare the returned error message to an expected one?

garimakhulbe:
Updated. Check now.

brendandixon:
@ahmetalpbalkan You'll only get an error of the underlying API uses Azure-AsyncOperation. Otherwise, the fields containing the error message (I believe) vary too much.

ahmetalpbalkan:
> Otherwise, the fields containing the error message (I believe) vary too much.

I don't think so, I remember reading error.message and error.code keys are common in all long running operation results yielding errors.

autorest/azure/async_test.go
```diff
@@ -812,6 +812,50 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+
+	if err == nil || (!strings.Contains(fmt.Sprintf("%v", err), "Unknown") && !strings.Contains(fmt.Sprintf("%v", err), "None")) {
+		t.Errorf("azure: DoPollForAsynchronous failed to return an appropriate error message for a failed Operations for unknown error")
+	}
+
+	autorest.Respond(r,
+		autorest.ByClosing())
+}
+
+func TestDoPollForAsynchronous_ReturnsAnKnownErrorForFailedOperations(t *testing.T) {
```
ahmetalpbalkan: AKnown?

ahmetalpbalkan:
@garimakhulbe issue still persists, the fix did not change anything

    network/SubnetsClient#CreateOrUpdate: Failure sending request: Status code 400 -- Original Error: Long running operation terminated with status 'Failed': Code="" Message=""

I think you really need to trigger a response with error.message, error.code and actually compare `err.Error()` string to what we are trying to get.

glide.lock
```diff
@@ -1,14 +1,15 @@
-hash: e20d5cdd14a2862b34f3bc112d77c3c3ad85e0c267101e88d791ec5a0e3ea4cb
```
ahmetalpbalkan: mistakenly updated?

autorest/azure/async_test.go
```diff
@@ -812,6 +812,50 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+	r3 := newOperationResourceErrorResponse(operationFailed)
+
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+	client.AppendAndRepeatResponse(r2, 2)
+	client.AppendAndRepeatResponse(r3, 1)
+
+	r, err := autorest.SendWithSender(client, mocks.NewRequest(),
+		DoPollForAsynchronous(time.Millisecond))
+	if err == nil || (!strings.Contains(fmt.Sprintf("%v", err), "BadArgument")) {
```
ahmetalpbalkan: redundant parens around Contains check

autorest/azure/async_test.go
```diff
@@ -812,6 +812,50 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+	client.AppendAndRepeatResponse(r2, 2)
+	client.AppendAndRepeatResponse(r3, 1)
+
+	r, err := autorest.SendWithSender(client, mocks.NewRequest(),
+		DoPollForAsynchronous(time.Millisecond))
+
+	if err == nil || (!strings.Contains(fmt.Sprintf("%v", err), "Unknown") && !strings.Contains(fmt.Sprintf("%v", err), "None")) {
```
ahmetalpbalkan: redundant parens around Contains check

ahmetalpbalkan: > fmt.Sprintf("%v", err)

this should be just `err.Error()`

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## v7.0.4
+-Error Message updated.
```
ahmetalpbalkan: “Better error messages for long running operation errors.”

autorest/azure/async_test.go
brendandixon: The extra parenthesis are fine. While they are technically redundant, in this case they add visual clarity of the precedence.

brendandixon: The `err == nil` can be deleted and the `&&` should be an `||`.

autorest/azure/async_test.go
brendandixon: The `err == nil` is unnecessary and can be removed.

autorest/azure/async_test.go
ahmetalpbalkan: > The extra parenthesis are fine.

I am kinda surprised golint/govet hasn't complained about those yet :) I can clearly see it is a simple  `!f()`  check, I don't think it provides extra visual clarity. :)

ahmetalpbalkan:
CI seems to be failing, it appears like `go vet` is now part of go.tool, if you remove the `go-get` statement for that in .travis.yml, I feel like the build will continue.

autorest/azure/async_test.go
brendandixon: It clarifies the associativity of the operators. Regardless, her changes will address this. BTW: `go vet` is an opinion only. :)

glide.lock
garimakhulbe: @ahmetalpbalkan  I am doing it in ```TestDoPollForAsynchronous_ReturnsAnKnownErrorForFailedOperations```, sending response with error message and code in operationResource and checking if it has an error code in error string. I also checked the other case where APl does not send OperationResource back for LRO.

Do you know if CreateOrUpdate sends error and code in coming in Operation Resource? How can I repro it? 

Followed below logic to fetch error:
```
if (operationResource present)
        take error message and code from there
else 
        "Unknown" and "None"
```

ahmetalpbalkan:
@garimakhulbe I realized the firstmost call is failing in this case. This is even before the polling starts, that's why we might be looking at the wrong place for the error.

How to repro:

1. Create a vnet
2. CreateOrUpdate  a subnet with addressPrefix: 10.0.0.0/24 in the vnet
3. CreateOrUpdate the same subnet with different (and invalid) addressPrefix: 10.0.0.0/3
4. That request returns HTTP400 Bad Request with:

```
{
  "error": {
    "code": "NetcfgInvalidSubnet",
    "message": "Subnet 'dm' is not valid in virtual network 'docker-machine-vnet'.",
    "details": []}}
```

If you get stuck let me know.

autorest/azure/async_test.go
garimakhulbe: Did not revert to err.Error() because it will give nil pointer if error is nil.

ahmetalpbalkan: why not:

```
if err == nil {
 t.Fatal("was expecting error; got nil")
else if err.Error() != expected {
  t.Fatalf("wrong error message: got=%q expected=%q...)
}
```

we should switch to gocheck for this repo as well, that'd dramatically shorten the test code.

brendandixon: Wonderful *follow-on* project. :)

garimakhulbe:
@ahmetalpbalkan This will only fix the blank code and message issue, ``` code=="" and message==""```. For above error, you will now get ``` code==Unknown and message==None```. 
Current approach only includes operation resource errors in error string, else unknown. Can you provide trace for error response like above scenario? Want to check field hierarchy and if these errors always come in fields as mentioned above.

garimakhulbe:
@ahmetalpbalkan @brendandixon Please review.

glide.lock
```diff
@@ -11,4 +11,4 @@ imports:
   vcs: git
   subpackages:
   - /pkcs12
-devImports: []
+devImports: []
```
ahmetalpbalkan: please undo

autorest/azure/async_test.go
```diff
@@ -812,6 +812,87 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+	client.AppendAndRepeatResponse(r2, 2)
+	client.AppendAndRepeatResponse(r3, 1)
+
+	r, err := autorest.SendWithSender(client, mocks.NewRequest(),
+		DoPollForAsynchronous(time.Millisecond))
+
+	if !strings.Contains(fmt.Sprintf("%v", err), getErrorString("BadArgument", "The provided database 'foo' has an invalid username.")) {
```
ahmetalpbalkan: please make this a `==` check; not Contains

autorest/azure/async_test.go
```diff
@@ -812,6 +812,87 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+func TestDoPollForAsynchronous_ReturnsErrorForFirstPutWithUnknownError(t *testing.T) {
+	// Return 400 bad response for unknown error
+	r1 := newAsynchronousErrorResponseWithUnknownError()
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+
+	r, err := autorest.SendWithSender(client, mocks.NewRequest(),
+		DoPollForAsynchronous(time.Millisecond))
+
+	if !strings.Contains(fmt.Sprintf("%v", err), getErrorString("Unknown", "None")) {
```
ahmetalpbalkan: please make this a `==` check; not `Contains`

garimakhulbe: Yes. My bad.

ahmetalpbalkan:
@garimakhulbe still doesn't work as expected I get error:

```
network/SubnetsClient#CreateOrUpdate: Failure sending request: Status code 400 -- Original Error: Error occurred unmarshalling JSON - Error = 'json: cannot unmarshal array into Go value of type string' JSON = '{
  "error": {
    "code": "NetcfgInvalidSubnet",
    "message": "Subnet 'dm' is not valid in virtual network 'docker-machine-vnet'.",
    "details": []
  }
}'
```

brendandixon:
@ahmetalpbalkan @garimakhulbe Hmm, odd. This gist (http://play.golang.org/p/DC_VKMWIuv) with that JSON works fine.

ahmetalpbalkan:
hmm maybe I applied the patch incorrectly.

ahmetalpbalkan:
nope I didn't. @brendandixon because you don't have `Details` field, which NRP returns an array, but our code has `string`:

image

@garimakhulbe I think we should omit the 'details' field for now.

brendandixon:
@garimakhulbe @ahmetalpbalkan Agreed.

garimakhulbe:
```Details``` removed. It should work fine now.

**Comment by ahmetalpbalkan** [comment]
@garimakhulbe works as expected now! :+1:

File: autorest/azure/async_test.go [hunk 9]
```diff
@@ -944,3 +1072,7 @@ func newProvisioningStatusResponse(status string) *http.Response {
 	r.Body = mocks.NewBody(fmt.Sprintf(pollingStateFormat, status))
 	return r
 }
+
+func getErrorString(code string, message string) string {
```
ahmetalpbalkan: `makeLongRunningOperationError`?

autorest/error.go
```diff
@@ -70,7 +71,7 @@ func NewErrorWithError(original error, packageType string, method string, resp *
 // StatusCode, Message, and original error (if any)).
 func (e DetailedError) Error() string {
 	if e.Original == nil {
-		return fmt.Sprintf("%s:%s %v %s", e.PackageType, e.Method, e.StatusCode, e.Message)
+		return fmt.Sprintf("%s#%s: %s: Status code %v", e.PackageType, e.Method, e.Message, e.StatusCode)
```
ahmetalpbalkan: `StatusCode=%d` is better

autorest/error.go
```diff
@@ -70,7 +71,7 @@ func NewErrorWithError(original error, packageType string, method string, resp *
...
 	}
-	return fmt.Sprintf("%s:%s %v %s -- Original Error: %v", e.PackageType, e.Method, e.StatusCode, e.Message, e.Original)
+	return fmt.Sprintf("%s#%s: %s: Status code %v -- Original Error: %v", e.PackageType, e.Method, e.Message, e.StatusCode, e.Original)
```
ahmetalpbalkan: `StatusCode=%d` is better

autorest/azure/async_test.go
```diff
@@ -812,6 +811,112 @@ func TestDoPollForAsynchronous_ReturnsAnErrorForFailedOperations(t *testing.T) {
...
+	// Return 400 bad response with error code and message in first put
+	r1 := newAsynchronousResponseWithError()
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+
+	r, err := autorest.SendWithSender(client, mocks.NewRequest(),
+		DoPollForAsynchronous(time.Millisecond))
+
+	if err.Error() != getErrorString("NetcfgInvalidSubnet", "Subnet is not valid in virtual network.") {
+		t.Errorf("azure: DoPollForAsynchronous failed to return an appropriate error message for a failed Operations")
```
ahmetalpbalkan: if you fail with `... expected=%q got=%q`, when it fails the person who's trying to understand the failure will be happier :)

ahmetalpbalkan:
generally LGTM, please squash into a single commit

ahmetalpbalkan:
Thanks a lot!

PR #53, opened 2016-04-13
Replace t.Error with t.Fatal

ahmetalpbalkan:
t.Error continues execution for the rest of the test and often
relies on miscomputed values etc, whereas t.Fatal returns from
the test case right away.

t.Fatal fits better in our use case as we compute-and-use
in our use case rather than exercising a set of test inputs.


brendandixon:
@ahmetalpbalkan LGTM.

PR #55, opened 2016-04-22
azure/persist: fallback to copy if rename fails

colemickens:
cc: @ahmetalpbalkan

### DIFF
--- a/autorest/azure/persist.go
+++ b/autorest/azure/persist.go
@@ -34,7 +34,7 @@ func SaveToken(path string, mode os.FileMode, token Token) error {
 		return fmt.Errorf("failed to create directory (%s) to store token in: %v", dir, err)
 	}
 
-	newFile, err := ioutil.TempFile(os.TempDir(), "token")
+	newFile, err := ioutil.TempFile(dir, "token")
 	if err != nil {
 		return fmt.Errorf("failed to create the temp file to write the token: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
hmm don't you think fallback logic just makes it harder to maintain/trace? If our tmpdir was `dirname(dest)` in the first place, we wouldn't even bother working with /tmp.

ahmetalpbalkan:
what I was trying to say is, this logic seems a lot complicated, if we can just do `os.TempFile(dirname(dest), "...")`, write the new file there and do Rename it would just work. I think the main issue was copying across filesystems, no? If we create a temp file next to the destination file, we won't have this problem?

colemickens:
I follow. I agree. Will change.

colemickens:
Changed per your feedback. Tested with the example in the repo and `-tokenCachePath="$HOME/temptmpfs/cache/token.json"` and `$HOME/temptmpfs` is a directory with `tmpfs` mounted in, so it's a different FS.

colemickens:
(I'll squash these, leaving them separate commits for reviewing purposes)

autorest/azure/persist.go
```diff
@@ -49,7 +49,7 @@ func SaveToken(path string, mode os.FileMode, token Token) error {
 
 	// Atomic replace to avoid multi-writer file corruptions
 	if err := os.Rename(tempPath, path); err != nil {
-		return fmt.Errorf("failed to move temporary token to desired output location. source=(%s). destination=(%s). error = %v", tempPath, path, err)
+		return fmt.Errorf("failed to move temporary token to desired output location. source=(%s). destination=(%s). error(rename)=%q", tempPath, path, err)
```
ahmetalpbalkan: the format can be better as:

    failed to move temporary token to desired output location: src=%q dsts=%q: %v

ahmetalpbalkan:
LGTM

ahmetalpbalkan:
thanks

PR #56, opened 2016-05-12
Add decorator DoRetryForStatusCodes for retry logic. Add CookieJar.

garimakhulbe:
- Add retry logic for 408, 500, 502, 503 and 504 status codes.
- Add CookieJar in Client.
- Change url path and query encoding logic.
- Fix DelayForBackoff for proper exponential delay.

### DIFF
--- a/autorest/preparer_test.go
+++ b/autorest/preparer_test.go
@@ -80,8 +80,8 @@ func ExampleCreatePreparer_multiple() {
 		"param2": "c",
 	}
 
-	p1 := CreatePreparer(WithBaseURL("https://microsoft.com/"), WithPath("/{param1}/b/{param2}/"))
-	p2 := CreatePreparer(WithPathParameters(params))
+	p1 := CreatePreparer(WithBaseURL("https://microsoft.com/"))
+	p2 := CreatePreparer(WithPathParameters("/{param1}/b/{param2}/", params))
 
 	r, err := p1.Prepare(&http.Request{})
 	if err != nil {
@@ -104,8 +104,8 @@ func ExampleCreatePreparer_chain() {
 		"param2": "c",
 	}
 
-	p := CreatePreparer(WithBaseURL("https://microsoft.com/"), WithPath("/{param1}/b/{param2}/"))
-	p = DecoratePreparer(p, WithPathParameters(params))
+	p := CreatePreparer(WithBaseURL("https://microsoft.com/"))
+	p = DecoratePreparer(p, WithPathParameters("/{param1}/b/{param2}/", params))
 
 	r, err := p.Prepare(&http.Request{})
 	if err != nil {
@@ -203,8 +203,7 @@ func ExampleWithEscapedPathParameters() {
 	}
 	r, err := Prepare(&http.Request{},
 		WithBaseURL("https://microsoft.com/"),
-		WithPath("/{param1}/b/{param2}/"),
-		WithEscapedPathParameters(params))
+		WithEscapedPathParameters("/{param1}/b/{param2}/", params))
 	if err != nil {
 		fmt.Printf("ERROR: %v\n", err)
 	} else {
@@ -221,8 +220,7 @@ func ExampleWithPathParameters() {
 	}
 	r, err := Prepare(&http.Request{},
 		WithBaseURL("https://microsoft.com/"),
-		WithPath("/{param1}/b/{param2}/"),
-		WithPathParameters(params))
+		WithPathParameters("/{param1}/b/{param2}/", params))
 	if err != nil {
 		fmt.Printf("ERROR: %v\n", err)
 	} else {
@@ -249,6 +247,31 @@ func ExampleWithQueryParameters() {
 	// Output: https://microsoft.com/a/b/c/?q1=value1&q2=value2
 }
 
+func ExampleWithBaseURL_InvalidURL() {
+	_, err := Prepare(&http.Request{}, WithBaseURL(":"))
+	fmt.Println(err)
+	// Output: parse :: missing protocol scheme
+}
+
+func TestWithPathWithInvalidPath(t *testing.T) {
+	p := "path%2*end"
+	if _, err := Prepare(&http.Request{}, WithBaseURL("https://microsoft.com/"), WithPath(p)); err == nil {
+		t.Fatalf("autorest: WithPath should fail for invalid URL escape error for path '%v' ", p)
+	}
+
+}
+
+func TestWithPathParametersWithInvalidPath(t *testing.T) {
+	p := "path%2*end"
+	m := map[string]interface{}{
+		"path1": p,
+	}
+	if _, err := Prepare(&http.Request{}, WithBaseURL("https://microsoft.com/"), WithPathParameters("/{path1}/", m)); err == nil {
+		t.Fatalf("autorest: WithPath should fail for invalid URL escape for path '%v' ", p)
+	}
+
+}
+
 func TestCreatePreparerDoesNotModify(t *testing.T) {
 	r1 := &http.Request{}
 	p := CreatePreparer()
@@ -267,7 +290,7 @@ func TestCreatePreparerRunsDecoratorsInOrder(t *testing.T) {
 	if err != nil {
 		t.Fatalf("autorest: CreatePreparer failed (%v)", err)
 	}
-	if r.URL.String() != "https://microsoft.com/1/2/3" {
+	if r.URL.String() != "https:/1/2/3" && r.URL.Host != "microsoft.com" {
 		t.Fatalf("autorest: CreatePreparer failed to run decorators in order")
 	}
 }
@@ -585,14 +608,14 @@ func TestWithPathCatchesNilURL(t *testing.T) {
 }
 
 func TestWithEscapedPathParametersCatchesNilURL(t *testing.T) {
-	_, err := Prepare(&http.Request{}, WithEscapedPathParameters(map[string]interface{}{"foo": "bar"}))
+	_, err := Prepare(&http.Request{}, WithEscapedPathParameters("", map[string]interface{}{"foo": "bar"}))
 	if err == nil {
 		t.Fatalf("autorest: WithEscapedPathParameters failed to catch a nil URL")
 	}
 }
 
 func TestWithPathParametersCatchesNilURL(t *testing.T) {
-	_, err := Prepare(&http.Request{}, WithPathParameters(map[string]interface{}{"foo": "bar"}))
+	_, err := Prepare(&http.Request{}, WithPathParameters("", map[string]interface{}{"foo": "bar"}))
 	if err == nil {
 		t.Fatalf("autorest: WithPathParameters failed to catch a nil URL")
 	}
@@ -610,7 +633,7 @@ func TestModifyingExistingRequest(t *testing.T) {
 	if err != nil {
 		t.Fatalf("autorest: Preparing an existing request returned an error (%v)", err)
 	}
-	if r.URL.String() != "https://bing.com/search?q=golang" {
+	if r.URL.String() != "https:/search?q=golang" && r.URL.Host != "bing.com" {
 		t.Fatalf("autorest: Preparing an existing request failed (%s)", r.URL)
 	}
 }
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -1,7 +1,9 @@
 package autorest
 
 import (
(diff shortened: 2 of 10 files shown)


garimakhulbe:
@ahmetalpbalkan @brendandixon @colemickens Please take a look.

ahmetalpbalkan:
Why are we adding a cookie jar?

garimakhulbe:
@ahmetalpbalkan Autorest code generator has some retry tests which use cookie.

autorest/preparer_test.go
```diff
@@ -70,7 +70,7 @@ func ExampleCreatePreparer() {
 	} else {
 		fmt.Println(r.URL)
 	}
-	// Output: https://microsoft.com/a/b/c
```
brendandixon: @garimakhulbe Why did you remove the "/microsoft.com"? The resulting URLs are not valid (they lack the required "//" after the HTTP scheme).

autorest/sender.go
```diff
@@ -222,13 +254,18 @@ func WithLogging(logger *log.Logger) SendDecorator {
...
 
 // DelayForBackoff invokes time.After for the supplied backoff duration raised to the power of
-// passed attempt (i.e., an exponential backoff delay). Backoff may be zero. The delay may be
-// canceled by closing the passed channel. If terminated early, returns false.
+// passed attempt (i.e., an exponential backoff delay). Backoff duration is in seconds and can set
+// to zero for no delay. The delay may be canceled by closing the passed channel. If terminated early,
+// returns false.
 func DelayForBackoff(backoff time.Duration, attempt int, cancel <-chan struct{}) bool {
 	select {
-	case <-time.After(time.Duration(math.Pow(float64(backoff), float64(attempt)))):
```
brendandixon: @garimakhulbe Is there a reason you chose to remove the exponential back off?

autorest/sender_test.go
```diff
@@ -144,7 +144,7 @@ func TestSend(t *testing.T) {
 func TestAfterDelayWaits(t *testing.T) {
 	client := mocks.NewSender()
 
-	d := 5 * time.Millisecond
```
brendandixon: @garimakhulbe I purposely chose Millisecond to keep the tests from getting too long. Were those values not working of you?

autorest/utility.go
```diff
@@ -97,7 +99,63 @@ func ensureValueString(value interface{}) string {
...
+		return pathEscape(s)
+	case "query":
+		return queryEscape(s)
+	default:
+		return s
+	}
+}
+
+func pathEscape(s string) string {
+	return strings.Replace(url.QueryEscape(s), "+", "%20", -1)
```
brendandixon: @garimakhulbe Do we need to escape the % in this string?

autorest/utility_test.go
```diff
@@ -186,18 +186,52 @@ func TestEnsureStrings(t *testing.T) {
...
 		"nil":    "",
+		"bytes":  string([]byte{255, 254, 253}),
 	}
 	v := ensureValueStrings(m)
 	if !reflect.DeepEqual(v, r) {
 		t.Fatalf("autorest: ensureValueStrings returned %v\n", v)
 	}
 }
 
+func ExampleStringWithSeparator() {
```
brendandixon: @garimakhulbe This and the following should be tests. Examples become part of the documentation. I don't believe we need to or should document these methods.

autorest/sender.go
garimakhulbe:
```Go
    backoff := 2 * time.Second
	fmt.Printf("float64(backoff) : %v\n\n",float64(backoff))
    // float64(backoff) : 2e+09

	for attempt := 0; attempt < 3; attempt++ {
		s := time.Duration(math.Pow(float64(backoff), float64(attempt)))
		fmt.Printf("Attempt # %v --  %v**%v = %v ~ %v \n",
			attempt, float64(backoff), float64(attempt), math.Pow(float64(backoff), float64(attempt)), s)
	}
	// Output :
	// Attempt # 0 --  2e+09**0 = 1 ~ 1ns 
	// Attempt # 1 --  2e+09**1 = 2e+09 ~ 2s 
	// Attempt # 2 --  2e+09**2 = 4e+18 ~ 1111111h6m40s  ---> weird duration

		
	fmt.Printf("\nDuration: %v \n",time.Duration(math.Pow(backoff.Seconds(), float64(3)))) // 
	// Duration: 8ns  --> changed float64(backoff) to backoff.Seconds()) but time still in nanosec
	
	
	fmt.Printf("\nDuration: %v \n",time.Duration(math.Pow(backoff.Seconds(), float64(3)))*time.Second)
	// Duration: 8s --> in seconds ( time.Duration(....)*time.Second )

File: autorest/preparer_test.go (hunk 1 above)
Comment (garimakhulbe) [inline autorest/preparer_test.go]: @brendandixon We are using url.Opaque now and url.String() uses below form if url.Opaque is non-empty.

```scheme:opaque?query#fragment       ---- no host ```

Comment (brendandixon) [inline autorest/preparer_test.go]: @garimakhulbe Host must still be there or the URL is invalid. Opaque only means that the portion of the URI is not being further inspected (as it is scheme dependent). For the http(s) schemes, the "//host.xxx" is required.

Comment (garimakhulbe) [inline autorest/preparer_test.go]: Yes, host is there in URL.

```Go
req := &http.Request{
		URL: &url.URL{
			Host:   "microsoft.com",
			Scheme: "https",
			Opaque: "/a/b/c",
		},
	}
fmt.Println("Host: ", req.URL.Host)
fmt.Println("URL: ", req.URL)

//
// Host:  microsoft.com
// URL:  https:/a/b/c

Comment (brendandixon) [inline autorest/preparer_test.go]: @garimakhulbe Ok. I'm confused. "https:/a/b/c" is __not__ a valid URL. It lacks the required host component.

Comment (colemickens) [inline autorest/preparer_test.go]: Is some of the confusion stemming from the fact that Request has it's own Host field that overrides the URL's Host?

Comment (garimakhulbe) [inline autorest/preparer_test.go]: Removed url.Opaque. It was giving invalid url string. URL string does not fetch host from url.Host when opaque is non-empty and probably expects everything in opaque. Not sure.
```
http:/a/b/c
```

I have updated WithPathParameters() to fix this. Instead of using WithPath() to concatenate path  and then WithPathParameters() for substituting the parameters, we will only use WithPathParameters() when we have path with parameters and it will build the path and parse to url.

```Go
//For URL: http://microsoft.com/a/{b}/c with b = xyz.

Prepare(..
           WithBaseURI("http://microsoft.com")  // --> url : http://microsoft.com
           WithPathParameters("/a/{b}/c",{"b":"xyz"}))   // --> url: http:/microsoft.com/a/xyz/c

//For simple URL with no parameters in path:   http://microsoft.com/m/n

Prepare(..
           WithBaseURI("http://microsoft.com")  // --> url : http://microsoft.com
           WithPath("/m/n")     // --> //url: http:/microsoft.com/m/n
```

autorest/sender_test.go
garimakhulbe: Yes, milliseconds with above change was giving zero seconds delay so increased it.
```Go
backoff := 5* time.Millisecond
fmt.Println(backoff.Seconds())
fmt.Println(time.Duration(backoff.Seconds()))

//Output:
0.005
0

```

autorest/utility.go
garimakhulbe: We don't need to. We are escaping + to %20 and %20 is valid for path so we don't need to escape %.

autorest/utility_test.go
garimakhulbe: done.

garimakhulbe:
@brendandixon @ahmetalpbalkan Is this PR good for merge?

autorest/sender.go
```diff
@@ -184,6 +186,36 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
...
 }
 
+// DoRetryForStatusCodes returns a SendDecorator that retries for specified statusCodes for up to the specified
+// number of attempts, exponentially backing off between requests using the supplied backoff
+// time.Duration (which may be zero). Retrying may be canceled by closing the optional channel on
+// the http.Request.
+func DoRetryForStatusCodes(attempts int, backoff time.Duration, codes ...int) SendDecorator {
+	return func(s Sender) Sender {
+		return SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
+			b := []byte{}
```
ahmetalpbalkan: `var b []byte{}` would be better (1 less alloc)

autorest/sender.go
```diff
@@ -184,6 +186,36 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
...
+			if r.Body != nil {
+				b, err = ioutil.ReadAll(r.Body)
+				if err != nil {
+					return resp, err
```
ahmetalpbalkan: try wrapping the error with a message.

autorest/sender.go
```diff
@@ -184,6 +186,36 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
...
+			attempts++
+			for attempt := 0; attempt < attempts; attempt++ {
+				r.Body = ioutil.NopCloser(bytes.NewBuffer(b))
+				resp, err = s.Do(r)
+				if err != nil || !ResponseHasStatusCode(resp, codes...) {
+					return resp, err
+				}
+				DelayForBackoff(backoff, attempt, r.Cancel)
+			}
+			return resp, err
```
ahmetalpbalkan: shouldn't this error be wrapped to indicate that we had retries and this request has exceeded retries

autorest/sender.go
```diff
@@ -222,11 +254,12 @@ func WithLogging(logger *log.Logger) SendDecorator {
...
 // DelayForBackoff invokes time.After for the supplied backoff duration raised to the power of
-// passed attempt (i.e., an exponential backoff delay). Backoff may be zero. The delay may be
-// canceled by closing the passed channel. If terminated early, returns false.
+// passed attempt (i.e., an exponential backoff delay). Backoff duration is in seconds and can set
+// to zero for no delay. The delay may be canceled by closing the passed channel. If terminated early,
+// returns false.
 func DelayForBackoff(backoff time.Duration, attempt int, cancel <-chan struct{}) bool {
 	select {
-	case <-time.After(time.Duration(math.Pow(float64(backoff), float64(attempt)))):
+	case <-time.After(time.Duration(math.Pow(backoff.Seconds(), float64(attempt))) * time.Second):
```
ahmetalpbalkan: .Seconds() essentially takes your ability to use sub-second backoff durations. If I give this a 4.5s, it will use 4. If I give it a 0.8 it will always be 0.

My suggestion, use `.Milliseconds()` to get integers and use them in math.Pow() and then multiply it with `time.Milliseconds` and pass it to `time.After`.

brendandixon: @garimakhulbe @ahmetalpbalkan I'm not certain it matters much. Sub-second back off seldom makes sense.

ahmetalpbalkan: fair. I'd implement this with `backoff * (2^attempt)`. I don't think `backoff^attempt` is a very good idea, for 10s, it will go off quite exponentially.

brendandixon: @ahmetalpbalkan The idea behind an exponential is that, if timeouts occur for our process, then they're occurring for others. Exponentially waiting spreads those request farther out, faster over time to give the underlying process a chance to recover. Simple linear back off tends to keep requests packed close together, often exacerbating the problem.

ahmetalpbalkan: @brendandixon what I suggested above is still exponential.

garimakhulbe: Backoff*(2^attempt) looks good. Its exponential and won't increase steeply. We can switch to this.

backoff^attempt for 10 seconds
10, 100, 1000,..

whereas backoff * (2^attempt)
10, 20, 40, 80,..

@brendandixon Is it look fine?

brendandixon: @garimakhulbe That will work well.

ahmetalpbalkan: @garimakhulbe Great, thanks.

For more see "Example exponential back-off algorithm" in https://en.wikipedia.org/wiki/Exponential_backoff
or https://www.awsarchitectureblog.com/2015/03/backoff.html

garimakhulbe:
Backoff delay fixed. Let me know if overall LGTY. I will send small PR from now on. :)

ahmetalpbalkan:
LGTM

PR #39, opened 2016-02-29
Close temp file on token persist

ahmetalpbalkan:
Fixes the

    "The process cannot access the file because it is being used by another process." 

issue on windows.

image

@colemickens

PR #59, opened 2016-06-10
Fix 'file in use' issue when renewing token in devicetoken flow

paulmey:
I had a couple of instances where token renewal failed to replace to persisted token. There error message was "file is in use". It checked with procexp to see what was holding the file open and it turns out to be the same process. I tracked it down to here.

Also made the tests work on Windows...

### DIFF
--- a/autorest/azure/persist_test.go
+++ b/autorest/azure/persist_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
LGTM. Looks awful a lot like #39, because we choose to use json.Decoder. It is [considered harmful](https://ahmetalpbalkan.com/blog/golang-json-decoder-pitfalls/).

autorest/azure/persist_test.go
```diff
@@ -90,8 +98,12 @@ func TestSaveToken(t *testing.T) {
 		t.Fatalf("azure: unexpected error when creating temp file: %v", err)
 	}
 	defer os.Remove(f.Name())
+	f.Close()
 
 	mode := os.ModePerm & 0642
+	if runtime.GOOS == "windows" {
```
ahmetalpbalkan: maybe actually just skip the `Perm()` check below on windows?

paulmey: yeah, that makes more sense... what would we be testing anyway...
I'll change that

colemickens:
Thanks Paul!

Issue #58, opened 2016-05-20
Store RetryAfter for subsequent requests


colemickens:
While investigating a bug that wound up simply being NRP taking 3 minutes to update a NIC... I noticed a different bug in the Azure SDK for Go:

1. T=0: PUT operation to update NIC (async response)

2. T=10: GET on operation status (in progress)

3. T=**70**: GET on operation status (in progress)

4. T=**130**: GET on operation status (in progress)

5. T=**190**: GET on operation status (success)

Problem is, first response includes RetryAfter, but the responses from GET operation status do not include the header.

Proposed: stash the RetryAfter from the initial PUT and use that as a fallback before the Client's DefaultPollingInterval.

cc: @brendandixon for thoughts on if this is okay, or if the polling needs to backoff, even though the API said to retryafter a certain interval.

brendandixon:
@colemickens @garimakhulbe I agree and this is a simple change. Here's the code change to `azure.DoPollForAsynchronous`:

```go
	delay = autorest.GetRetryAfter(resp, delay)
	resp, err = autorest.SendWithSender(s, r,
				autorest.AfterDelay(delay))
```

colemickens:
Current code is:

```
    resp, err = autorest.SendWithSender(s, r,
        autorest.AfterDelay(autorest.GetRetryAfter(resp, delay)))
```

I don't see how the suggested change would fix. Further, I applied the change and I'm still seeing the same polling behavior described in the Issue description.

brendandixon:
@colemickens @garimakhulbe The change is intended update the local `delay` variable to hold that last retrieved value of the RetryAfter header. Isn't that the desire?

On first entry to `azure.DoPollforAsynchronous` the `delay` variable contains a user supplied default value. `autorest.GetRetryAfter` will return the value of the RetryAfter header or the default if the header is absent. In your case, it should update `delay` to the value of the initial RetryAfter header which will then be used on subsequent calls.

What's not working?

colemickens:
After finally 1) making the correct change and 2) making it in the right copy of go-autorest, I can confirm that @brendandixon's suggested change works.

garimakhulbe:
Ok then I will send a new PR with tests updated for this change.

garimakhulbe:
Fixed.

PR #60, opened 2016-06-15
Fix for #58: Store RetryAfter for subsequent requests

colemickens:
I think @garimakhulbe was making this change as part of a larger PR, but I'm sending this so that I can vendor it now for the PR I'm preparing for Kubernetes. This change saves me upwards of a minute or two whenever provisioning external load balancers for Kubernetes Services. I've been running this code for a while as-is; works as intended.


colemickens:
(addresses #58)

ahmetalpbalkan:
LGTM

garimakhulbe:
Thanks @colemickens

PR #61, opened 2016-06-22
Add code for parameter validation

garimakhulbe:
This code will check null, empty or other validations on input parameters. SDK will pass validation array to go-autorest Validate() method which will check listed validations on parameter value using reflect package.

For example, if ```CreateOrUpdate() ```method has resource group and subscription as required parameters, then it will call ```Validate()``` on```[]autorest.Validation``` like below...
```Go

func (c client) CreateOrUpdate(resourceGroupName string, subscriptionID string) (result, error) {
	if err := autorest.Validate([]autorest.Validation{
		{client.SubscriptionID,
			[]autorest.Constraint{{"subscriptionID", "Empty", true, nil}}},
		{resourceGroupName,
			[]autorest.Constraint{{"resourceGroupName", "Empty", true, nil},
				{"resourceGroupName", "MaxLength", 10, nil},
				{"resourceGroupName", "MinLength", 3, nil},
				{"resourceGroupName", "Pattern", "[a-zA-Z0-9]+", nil}}},
		},
	}); err!=nil {
		return err
	}

	Preparaer..
	Sender..
	Responder...
}

```

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -62,8 +62,9 @@ func DoPollForAsynchronous(delay time.Duration) autorest.SendDecorator {
 					return resp, err
 				}
 
+				delay = autorest.GetRetryAfter(resp, delay)
 				resp, err = autorest.SendWithSender(s, r,
-					autorest.AfterDelay(autorest.GetRetryAfter(resp, delay)))
+					autorest.AfterDelay(delay))
 			}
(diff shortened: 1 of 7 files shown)


garimakhulbe:
Messed up commits using squash so closing it. Sent new PR on this.

PR #62, opened 2016-06-22
Add code for parameter validation

garimakhulbe:
This code will check null, empty or other validations on input parameters. SDK will pass validation array to go-autorest Validate() method which will check listed validations on parameter value using reflect package.

For example, if ```CreateOrUpdate() ```method has resource group and subscription as required parameters, then it will call ```Validate()``` on```[]autorest.Validation``` like below...
```Go

func (c client) CreateOrUpdate(resourceGroupName string, subscriptionID string) (result, error) {
	if err := autorest.Validate([]autorest.Validation{
		{client.SubscriptionID,
			[]autorest.Constraint{{"subscriptionID", "Empty", true, nil}}},
		{resourceGroupName,
			[]autorest.Constraint{{"resourceGroupName", "Empty", true, nil},
				{"resourceGroupName", "MaxLength", 10, nil},
				{"resourceGroupName", "MinLength", 3, nil},
				{"resourceGroupName", "Pattern", "[a-zA-Z0-9]+", nil}}},
		},
	}); err!=nil {
		return err
	}

	Preparaer..
	Sender..
	Responder...
}

```

### DIFF
--- a/autorest/validation/validation.go
+++ b/autorest/validation/validation.go
@@ -1,0 +1,381 @@
+/*
+Package validation provides methods for validating parameter value using reflection.
+*/
+package validation
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+)
+
+// Constraint stores constraint name, target field name
+// Rule and chain validations.
+type Constraint struct {
+
+	// Target field name for validation.
+	Target string
+
+	// Constraint name e.g. minLength, MaxLength, Pattern, etc.
+	Name string
+
+	// Rule for constraint e.g. greater than 10, less than 5 etc.
+	Rule interface{}
+
+	// Chain Validations for struct type
+	Chain []Constraint
+}
+
+// Validation stores parameter-wise validation.
+type Validation struct {
+	TargetValue interface{}
+	Constraints []Constraint
+}
+
+// Constraint list
+const (
+	Empty            = "Empty"
+	Null             = "Null"
+	ReadOnly         = "ReadOnly"
+	Pattern          = "Pattern"
+	MaxLength        = "MaxLength"
+	MinLength        = "MinLength"
+	MaxItems         = "MaxItems"
+	MinItems         = "MinItems"
+	MultipleOf       = "MultipleOf"
+	UniqueItems      = "UniqueItems"
+	InclusiveMaximum = "InclusiveMaximum"
+	ExclusiveMaximum = "ExclusiveMaximum"
+	ExclusiveMinimum = "ExclusiveMinimum"
+	InclusiveMinimum = "InclusiveMinimum"
+)
+
+// Validate method validates constraints on parameter
+// passed in validation array.
+func Validate(m []Validation) error {
+	for _, item := range m {
+		v := reflect.ValueOf(item.TargetValue)
+		for _, constraint := range item.Constraints {
+			var err error
+			switch v.Kind() {
+			case reflect.Ptr:
+				err = validatePtr(v, constraint)
+			case reflect.String:
+				err = validateString(v, constraint)
+			case reflect.Struct:
+				err = validateStruct(v, constraint)
+			case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+				err = validateInt(v, constraint)
+			case reflect.Float32, reflect.Float64:
+				err = validateFloat(v, constraint)
+			case reflect.Array, reflect.Slice, reflect.Map:
+				err = validateArrayMap(v, constraint)
+			default:
+				err = createError(v, constraint, fmt.Sprintf("unknown type %v", v.Kind()))
+			}
+
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func validateStruct(x reflect.Value, v Constraint) error {
+	f := x.FieldByName(v.Target)
+	if isZero(f) {
+		return createError(x, v, fmt.Sprintf("field %q doesn't exist", v.Target))
+	}
+	if err := Validate([]Validation{
+		{
+			TargetValue: getInterfaceValue(f),
+			Constraints: []Constraint{v},
+		},
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validatePtr(x reflect.Value, v Constraint) error {
+	if v.Name == ReadOnly {
+		if !x.IsNil() {
+			return createError(x.Elem(), v, "readonly parameter; must send as nil or Empty in request")
+		}
+		return nil
+	}
+	if x.IsNil() {
+		return checkNil(x, v)
+	}
+	if v.Chain != nil {
+
+		return Validate([]Validation{
+			{
+				TargetValue: getInterfaceValue(x.Elem()),
+				Constraints: v.Chain,
+			},
+		})
+	}
+	return nil
+}
+
+func validateInt(x reflect.Value, v Constraint) error {
+	i := x.Int()
+	r, ok := v.Rule.(int)
+	if !ok {
+		return createError(x, v, fmt.Sprintf("rule must be integer value for %v constraint; got: %v", v.Name, v.Rule))
+	}
+	switch v.Name {
+	case MultipleOf:
+		if i%int64(r) != 0 {
+			return createError(x, v, fmt.Sprintf("value must be a multiple of %v", r))
+		}
+	case ExclusiveMinimum:
+		if i <= int64(r) {
+			return createError(x, v, fmt.Sprintf("value must be greater than %v", r))
+		}
+	case ExclusiveMaximum:
+		if i >= int64(r) {
+			return createError(x, v, fmt.Sprintf("value must be less than %v", r))
+		}
+	case InclusiveMinimum:
+		if i < int64(r) {
+			return createError(x, v, fmt.Sprintf("value must be greater than or equal to %v", r))
+		}
+	case InclusiveMaximum:
+		if i > int64(r) {
+			return createError(x, v, fmt.Sprintf("value must be less than or equal to %v", r))
+		}
+	default:
+		return createError(x, v, fmt.Sprintf("constraint %v is not applicable for type integer", v.Name))
+	}
+	return nil
+}
+
+func validateFloat(x reflect.Value, v Constraint) error {
+	f := x.Float()
+	r, ok := v.Rule.(float64)
+	if !ok {
+		return createError(x, v, fmt.Sprintf("rule must be float value for %v constraint; got: %v", v.Name, v.Rule))
+	}
+	switch v.Name {
+	case ExclusiveMinimum:
+		if f <= r {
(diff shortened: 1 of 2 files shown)


garimakhulbe:
@brendandixon @ahmetalpbalkan please comment.

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+	null             = "Null"
+	readOnly         = "ReadOnly"
+	pattern          = "Pattern"
+	maxLength        = "MaxLength"
+	minLength        = "MinLength"
+	maxItems         = "MaxItems"
+	minItems         = "MinItems"
+	multipleOf       = "MultipleOf"
+	uniqueItems      = "UniqueItems"
+	inclusiveMaximum = "InclusiveMaximum"
```
ahmetalpbalkan: can't we just export these constants as integers (with iota)? Do they have to have string values?

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+// ValidationError stroes detailed validation error
+type ValidationError struct {
+	Constraint  string
+	Target      string
+	TargetValue interface{}
+	Details     string
+}
+
+func (e ValidationError) Error() string {
+	return fmt.Sprintf("autorest/azure: validation error: Parameter=%v Value=%v Constraint=%v Details=%v",
```
ahmetalpbalkan: use %#v

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+// Rule and chain validations.
+type Constraint struct {
+
+	// Target Field name for validation.
+	Target string
+
+	// Constraint name .e.g. Minlength, Maxlength, Pattern, etc.
+	Name string
+
+	// Rule for constraint .e.g. greater than 10, less than 5 etc.
```
ahmetalpbalkan: redundant `.` before `e.g.`

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+	"reflect"
+	"regexp"
+)
+
+// Constraint stroes constraint name, target field name
+// Rule and chain validations.
+type Constraint struct {
+
+	// Target Field name for validation.
+	Target string
```
ahmetalpbalkan: `field name` (also the comment on Constraint has upper/lower-case issues as well).

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func validateStruct(x interface{}, v Constraint) error {
+	z := reflect.ValueOf(x)
+	f := z.FieldByName(v.Target)
```
ahmetalpbalkan: how do you handle if `f` is zero (i.e. when the field with this name is not found)?

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+		if _, ok := v.Rule.(int); !ok {
+			return createError(x, v, fmt.Sprintf("rule must be integer for %v constraint; got: %v", v.Name, v.Rule))
+		}
+		if z.Len() < v.Rule.(int) {
+			return createError(x, v, fmt.Sprintf("mininum item limit is %v; got: %v", v.Rule, z.Len()))
+		}
+	case uniqueItems:
+		if z.Kind() == reflect.Array || z.Kind() == reflect.Slice {
+			if !checkForUniqueInArray(x) {
+				return createError(x, v, fmt.Sprintf("all items in parameter %v must be unique; got:%v", v.Target, z))
```
ahmetalpbalkan: I suggest using %q for `v.Target`

autorest/validations.go
ahmetalpbalkan: also if e.TargetValue is map or an array, this message is going to be quite unreadable.

so perhaps move Constraint before Value?

ahmetalpbalkan:
Looks cool. It appears like there is massive test coverage. It'd be good to add reflection-specific tests as well (i.e. the field is not found, or field is not of the type we expect etc) and see how we handle those (if we are supposed to).

ahmetalpbalkan:
Also I'm gonna suggest this should be another package. It doesn't have much to do with "azure".

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
+package autorest
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+)
+
+// Constraint stroes constraint name, target field name
```
brendandixon: @garimakhulbe Please correct spelling to "stores."

garimakhulbe:
@ahmetalpbalkan What do you mean by reflection-specific tests and handling those? Code has type checks like below and we have tests for these.
```Go
case reflect.Array, reflect.Slice, reflect.Map:
	err = validateArrayMap(item.TargetValue, constraint)
default:
       err = createError(item.TargetValue, constraint, fmt.Sprintf("unknown type %v", v.Kind()))
```
```Go
r, ok := v.Rule.(int)
if !ok {
	return createError(x, v, fmt.Sprintf("rule must be integer value for %v constraint; got: %v", v.Name, v.Rule))
	}
```

ahmetalpbalkan:
@garimakhulbe fair enough, I saw that you don't handle the not-found case from `FieldByName` and adding a test should be good enough.

autorest/azure/token.go
```diff
@@ -137,7 +137,7 @@ func (secret *ServicePrincipalCertificateSecret) SignJwt(spt *ServicePrincipalTo
 
 	token := jwt.New(jwt.SigningMethodRS256)
 	token.Header["x5t"] = thumbprint
-	token.Claims = map[string]interface{}{
+	token.Claims = jwt.MapClaims{
```
ahmetalpbalkan: #64 fixes this, you should not make the change.

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+	Target string
+
+	// Constraint name .e.g. Minlength, Maxlength, Pattern, etc.
+	Name string
+
+	// Rule for constraint .e.g. greater than 10, less than 5 etc.
+	Rule interface{}
+
+	// Chain Validations for struct type
+	Chain *[]Constraint
```
ahmetalpbalkan: no need for *

autorest/validations.go
```diff
@@ -0,0 +1,359 @@
...
+// Validate method validates constraints on parameter
+// passed in validation array.
+func Validate(m []Validation) error {
+	for _, item := range m {
+		v := reflect.ValueOf(item.TargetValue)
+		for _, constraint := range item.Constraints {
+			var err error
+			switch v.Kind() {
+			case reflect.Ptr:
+				err = validatePtr(item.TargetValue, constraint)
```
ahmetalpbalkan: pass v down to these

ahmetalpbalkan:
I suggest moving this to `autorest/validation/` directory as it is completely independent and will pollute the godoc for the `autorest/azure` package.

garimakhulbe:
@ahmetalpbalkan Updated for comments. Take a look.

garimakhulbe:
@ahmetalpbalkan All comments are addressed. Please take a look.

autorest/validations/validations.go
```diff
@@ -0,0 +1,415 @@
...
+		if m[val] {
+			return false
+		}
+		m[val] = true
+	}
+	return true
+}
+
+func getConstraintName(i int) string {
+	switch i {
```
ahmetalpbalkan: this function could be just making a map lookup. also where are these numbers coming from? could these be constants/enums?

autorest/validations/validations.go
```diff
@@ -0,0 +1,415 @@
...
+
+// Constraint stores constraint name, target field name
+// Rule and chain validations.
+type Constraint struct {
+
+	// Target field name for validation.
+	Target string
+
+	// Constraint name e.g. Minlength, Maxlength, Pattern, etc.
+	Name int
```
ahmetalpbalkan: I don't know the domain but why is constraint an int? looks like you could define them as strings below in the const section instead of ints.

autorest/validations/validations.go
```diff
@@ -0,0 +1,415 @@
...
+
+// Validation stores parameter-wise validation.
+type Validation struct {
+	TargetValue interface{}
+	Constraints []Constraint
+}
+
+// Constraints
+const (
+	InclusiveMaximum = iota + 1
```
ahmetalpbalkan: not sure where the numbers are coming from, but why does this start from iota+1?

ahmetalpbalkan:
I've got 2 more general comments:

- can you please make the pkg name singular and not plural
- can you make sure errors returned from exported methods of this package are prefixed with `autorest/validate`?

autorest/validations/validations.go
garimakhulbe: These constraints are defined as enum in code generator tool (C#). 
```C#
public enum Constraint
    {
        None,  // 0
        InclusiveMaximum, // 1
        ExclusiveMaximum, // 2 and so on
        InclusiveMinimum,
        ExclusiveMinimum,
        MaxLength,
        MinLength,
        Pattern,
        MaxItems,
        MinItems,
        UniqueItems,
        MultipleOf
    }
```

so I kept the same order here when defining with iota. 

```Go
const (
 +	InclusiveMaximum = iota + 1 // 1
 + ExclusiveMaximum  // 2 and so on
 +	InclusiveMinimum
 +	ExclusiveMinimum
 +	MaxLength
 +	MinLength
 +	Pattern
 +	MaxItems
 +	MinItems
 +	UniqueItems
 +	MultipleOf
 +	Null
 +	Empty
 +	ReadOnly
 +)
```

autorest/validations/validations.go
garimakhulbe: This change to int because I am using iota for constants now. It was of string type with string constants earlier.

garimakhulbe:
@ahmetalpbalkan updated for comments.

garimakhulbe:
@ahmetalpbalkan code updated. merging it.

autorest/validation/validation.go
```diff
@@ -0,0 +1,381 @@
...
+
+// Constraint stores constraint name, target field name
+// Rule and chain validations.
+type Constraint struct {
+
+	// Target field name for validation.
+	Target string
+
+	// Constraint name e.g. minLength, maxLength, pattern, etc.
+	Name string
```
ahmetalpbalkan: so since these field is `string`, will the sdk say `Constraint{"fieldName", "Empty", ...}`? Assuming the answer is yes, instead of making SDK say `"Empty"` can't we export the constraint consts below and refer to it as `constraint.Empty` in the SDK code?

garimakhulbe: Yes, we can do that. I will update this in sdk code generator.

ahmetalpbalkan: ok, in that case you'll need to export the consts below.

garimakhulbe:
@ahmetalpbalkan Thanks for the comments. Merged.

PR #64, opened 2016-06-22
Add glide to travis build

paulmey:


.travis.yml
```diff
@@ -4,6 +4,9 @@ language: go
 
 before_script:
   - go get -u github.com/golang/lint/golint
+  - go get -u github.com/Masterminds/glide
+  - export GO15VENDOREXPERIMENT=1
```
ahmetalpbalkan: hm maybe just bump to go:1.6 below?

paulmey:
Moved to Go 1.6, needed to change an Example since `go vet` got more picky with names in 1.6

ahmetalpbalkan:
lgtm. I wonder what the error from govet was.

paulmey:
```
autorest/date/timerfc1123_test.go:11: ExampleParseTimeRFC1123 refers to unknown identifier: ParseTimeRFC1123
```
What comes after `Example` needs to be an existing 'thing' to put it in the right place in godoc.

PR #67, opened 2016-06-22
glide.lock update and fix

garimakhulbe:


garimakhulbe:
@ahmetalpbalkan glide.lock updated.

glide.lock
```diff
@@ -1,14 +1,15 @@
...
   - .
 - name: golang.org/x/crypto
-  version: 3760e016850398b85094c4c99e955b8c3dea5711
+  version: f3241ce8505855877cc8a9717bd61a0f7c4ea83c
   repo: https://github.com/golang/crypto.git
   vcs: git
   subpackages:
-  - /pkcs12
+  - pkcs12
+  - pkcs12/internal/rc2
```
ahmetalpbalkan: why are these changing and why did you update the `golang.org/x/crypto` as well?

ahmetalpbalkan:
also make sure you read go-jwt's v3 migration guide https://github.com/dgrijalva/jwt-go/blob/master/MIGRATION_GUIDE.md

glide.lock
garimakhulbe: I ran glide up. It updated these.

colemickens: It is possible for things like this to change due to change in `glide`'s operation from version to version. (Please make sure you are using the latest `glide`: `go get -u github.com/Masterminds/glide`)

garimakhulbe: Updated glide, lock file is still same. I think there is no issue in lock file. We use glide up for version update and as per glide page ``` glide up                                # Update to newest versions of the package ``` for all the packages listed in yml. If we want a version update for single package in lock file, we need to do it manually which is not recommended. 
I think its fine updating using glide up. We do this in sdk all the time whenever update go-autorest version using glide. We can add a comment in changelog for heads up for users.

garimakhulbe: @ahmetalpbalkan Is it look good to you?

ahmetalpbalkan: sure.

PR #68, opened 2016-06-23
Address go-report issues.

boumenot:
Unused variable.
Typo.

### DIFF
--- a/autorest/azure/token_test.go
+++ b/autorest/azure/token_test.go
@@ -497,6 +497,9 @@ func newServicePrincipalTokenCertificate() *ServicePrincipalToken {
 		panic(err)
 	}
 	certificate, err := x509.ParseCertificate(certificateBytes)
+	if err != nil {
+		panic(err)
+	}
 
 	spt, _ := NewServicePrincipalTokenFromCertificate(TestOAuthConfig, "id", certificate, privateKey, "resource")
 	return spt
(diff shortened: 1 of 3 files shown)


autorest/azure/token_test.go
```diff
@@ -496,7 +496,7 @@ func newServicePrincipalTokenCertificate() *ServicePrincipalToken {
 	if err != nil {
 		panic(err)
 	}
-	certificate, err := x509.ParseCertificate(certificateBytes)
+	certificate, _ := x509.ParseCertificate(certificateBytes)
```
ahmetalpbalkan: please check the error instead

ahmetalpbalkan:
LGTM except nits

autorest/azure/token_test.go
boumenot: I made the code panic, just as it does above.

ahmetalpbalkan:
@boumenot ugh whoever put these panics in there, not a good idea. test helpers can take an instance of `*testing.T` and just call `t.Fatal` when things go wrong

colemickens:
That would be me. I'm assuming panic() actually stops the tests in their tracks?

I can send a PR changing the `panic`s to `t.Fatal`s.

colemickens:
https://github.com/Azure/go-autorest/pull/70

PR #69, opened 2016-06-24
update azure error for detailed error

garimakhulbe:
Fix sdk issue [#339](https://github.com/Azure/azure-sdk-for-go/issues/339). Code to decode "details:" in service error when present.
```
error:
      code:
      message:
      **details:[]**
```

### DIFF
--- a/autorest/azure/async_test.go
+++ b/autorest/azure/async_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/mocks"
 	"io/ioutil"
 	"net/http"
+	"reflect"
 	"strings"
 	"sync"
 	"testing"
@@ -936,9 +937,17 @@ func TestDoPollForAsynchronous_ReturnsErrorForFirstPutRequest(t *testing.T) {
 		t.Fatalf("azure: returned error is not azure.RequestError: %T", err)
 	}
 
-	if expected := makeRequestErrorString("InvalidParameter",
-		"tom-service-DISCOVERY-server-base-v1.core.local' is not a valid captured VHD blob name prefix.", 400); reqError.Error() != expected {
-		t.Fatalf("azure: wrong error. expected=%q; got=%q", expected, reqError)
+	expected := &RequestError{
+		ServiceError: &ServiceError{
+			Code:    "InvalidParameter",
+			Message: "tom-service-DISCOVERY-server-base-v1.core.local' is not a valid captured VHD blob name prefix.",
+		},
+		DetailedError: autorest.DetailedError{
+			StatusCode: 400,
+		},
+	}
+	if !reflect.DeepEqual(reqError, expected) {
+		t.Fatalf("azure: wrong error. expected=%q\ngot=%q", expected, reqError)
 	}
 
 	defer res.Body.Close()
@@ -1103,9 +1112,4 @@ func newProvisioningStatusResponse(status string) *http.Response {
 
 func makeLongRunningOperationErrorString(code string, message string) string {
 	return fmt.Sprintf("Long running operation terminated with status 'Failed': Code=%q Message=%q", code, message)
-}
-
-func makeRequestErrorString(code, message string, status int) string {
-	return fmt.Sprintf("azure: Service returned an error. Code=%q Message=%q Status=%d",
-		code, message, status)
 }
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -6,6 +6,7 @@ See the included examples for more detail.
 package azure
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -29,12 +30,20 @@ const (
 
 // ServiceError encapsulates the error response from an Azure service.
 type ServiceError struct {
-	Code    string `json:"code"`
-	Message string `json:"message"`
+	Code    string         `json:"code"`
+	Message string         `json:"message"`
+	Details *[]interface{} `json:"details"`
 }
 
 func (se ServiceError) Error() string {
-	return fmt.Sprintf("Azure Error: Code=%q Message=%q", se.Code, se.Message)
+	if se.Details != nil {
+		d, err := json.Marshal(*(se.Details))
+		if err != nil {
+			return fmt.Sprintf("Code=%q Message=%q Details=%v", se.Code, se.Message, *se.Details)
+		}
+		return fmt.Sprintf("Code=%q Message=%q Details=%v", se.Code, se.Message, string(d))
+	}
+	return fmt.Sprintf("Code=%q Message=%q", se.Code, se.Message)
 }
 
 // RequestError describes an error response returned by Azure service.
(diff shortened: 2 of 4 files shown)


garimakhulbe:
@ahmetalpbalkan @brendandixon please take a look.

autorest/azure/async_test.go
```diff
@@ -1106,6 +1106,6 @@ func makeLongRunningOperationErrorString(code string, message string) string {
 }
 
 func makeRequestErrorString(code, message string, status int) string {
-	return fmt.Sprintf("azure: Service returned an error. Code=%q Message=%q Status=%d",
+	return fmt.Sprintf("autorest/azure: Service returned an error. Code=%q Message=%q Details=<nil> Status=%d",
```
ahmetalpbalkan: I think this line change is not necessary (unless you're doing some pattern matching?)

autorest/azure/azure.go
```diff
@@ -29,12 +29,13 @@ const (
...
-	Code    string `json:"code"`
-	Message string `json:"message"`
+	Code    string       `json:"code"`
+	Message string       `json:"message"`
+	Details *interface{} `json:"details"`
 }
 
 func (se ServiceError) Error() string {
-	return fmt.Sprintf("Azure Error: Code=%q Message=%q", se.Code, se.Message)
+	return fmt.Sprintf("Azure Error: Code=%q Message=%q Details=%v", se.Code, se.Message, se.Message)
```
ahmetalpbalkan: you mean se.Details?

autorest/azure/azure.go
```diff
@@ -50,8 +51,8 @@ type RequestError struct {
 
 // Error returns a human-friendly error message from service error.
 func (e RequestError) Error() string {
-	return fmt.Sprintf("azure: Service returned an error. Code=%q Message=%q Status=%d",
-		e.ServiceError.Code, e.ServiceError.Message, e.StatusCode)
+	return fmt.Sprintf("autorest/azure: Service returned an error. Code=%q Message=%q Details=%v Status=%v",
+		e.ServiceError.Code, e.ServiceError.Message, e.ServiceError.Details, e.StatusCode)
```
ahmetalpbalkan: this `Error()` function seemed a bit unnecessary. It is almost as same as `ServiceError`

What is the difference of `RequestError` and `ServiceError` again?

autorest/azure/async_test.go
garimakhulbe: It is used only once so I will remove this function and use deepEqual or string match instead.

autorest/azure/azure.go
garimakhulbe: ServiceError is response error object. ``` [error: code, message, details] ```
RequestError is ```ServiceError + RequestID + DetailedError(package name, statuscode, Method, message)```. I think we can remove ServiceError.Error() because we never refer to ServiceError directly. It always comes wrapped inside RequestError.

ahmetalpbalkan: @garimakhulbe if it is wrapped in there, you should probably just implement `Error()` on this, and instead of reaching its members, perhaps you should just do %v on the err variable, which would print the Error() output.

garimakhulbe: got it. https://play.golang.org/p/-GZ2X-Gojj

garimakhulbe:
@ahmetalpbalkan updated as per comments. Take a look.

autorest/azure/azure.go
```diff
@@ -29,12 +29,13 @@ const (
...
-	Code    string `json:"code"`
-	Message string `json:"message"`
+	Code    string       `json:"code"`
+	Message string       `json:"message"`
+	Details *interface{} `json:"details"`
 }
 
 func (se ServiceError) Error() string {
-	return fmt.Sprintf("Azure Error: Code=%q Message=%q", se.Code, se.Message)
+	return fmt.Sprintf("Code=%q Message=%q Details=%v", se.Code, se.Message, se.Details)
```
ahmetalpbalkan: I'd love to see a test validating actual output of this method (especially the Details=%v part).

Also I think we should not add `Details=<nil>` if `Details == nil`. It just looks bad. an if statement here would be preferable to prevent adding that unnecessary suffix.

autorest/azure/azure.go
```diff
@@ -50,8 +51,8 @@ type RequestError struct {
 
 // Error returns a human-friendly error message from service error.
 func (e RequestError) Error() string {
-	return fmt.Sprintf("azure: Service returned an error. Code=%q Message=%q Status=%d",
-		e.ServiceError.Code, e.ServiceError.Message, e.StatusCode)
+	return fmt.Sprintf("autorest/azure: Service returned an error. %v Status=%v",
```
ahmetalpbalkan: Hmm %v is going to be super long, you realize that right? Why not `Status=%d Error=%v`?

garimakhulbe: To keep it consistent with other error strings(like detailederror) where status comes in end. I will reorder it.

autorest/azure/azure.go
garimakhulbe: Thought users won't look for details in error when detail is nil. They will know by error string. But got your point that if for RPs that never send details, this will always show up nil. Will update.

autorest/azure/azure.go
ahmetalpbalkan: @garimakhulbe as long as it looks fine it is good. that's why I also would love to see a real string generated from this method in a test so that we can see how it actually behaves in real life and design it from there. My imagination is not strong enough to picture what kind of a string this method will return.

autorest/azure/azure.go
```diff
@@ -29,12 +30,20 @@ const (
...
 
 func (se ServiceError) Error() string {
-	return fmt.Sprintf("Azure Error: Code=%q Message=%q", se.Code, se.Message)
+	if se.Details != nil {
+		d, err := json.Marshal(*(se.Details))
+		if err != nil {
+			return fmt.Sprintf("Code=%q Message=%q Details=%v", se.Code, se.Message, *se.Details)
+		}
+		return fmt.Sprintf("Code=%q Message=%q Details=%v", se.Code, se.Message, string(d))
+	}
```
garimakhulbe: @ahmetalpbalkan I added a nil check for details. 
I use json marshal for converting details to string because if I use it directly, its printing with type..
```Go
fmt.Printf("Details=%s\n", *(r.ServiceError.Details))
// Details=[map[message:error message code:Conflict] map[code:Conflict2 message:error message]]
```
Marshaler prints it..
```Go
d, err := json.Marshal(*(se.Details))
fmt.Printf("Details=%v", string(d))
// Details=[{"code":"conflict1","message":"error message1"}]
```
Else if we don't want to use Marshal, I need to unwrap it somewhat like below. 
```Go
var s []string
	for _, v := range *(r.ServiceError.Details) {
		m, _ := v.(map[string]interface{})
		for k, val := range m {
			s = append(s, k+":"+val.(string))
		}
	}
	fmt.Println("Details=", s)

// Details= [code:Conflict message:error message code:Conflict2 message:error message]
```
What you think?

ahmetalpbalkan: @garimakhulbe JSON marshaling is just the perfect fit here. I'm cool with this. Thanks!

garimakhulbe: @ahmetalpbalkan Thanks. Is it look good to you overall? Shall I merge?

ahmetalpbalkan: go for it

PR #73, opened 2016-06-27
Serialization of Environment

### LINKED ISSUE #72: Generalized Support for Azure Stack
@colemickens - I would like your opinion on this.

We've had a [request][1] from the Packer community to support Azure Stack.  My thinking is that the user of Packer creates a JSON document detailing their [Environment][2].  Packer can then deserializes the JSON, and the code has an Environment instance appropriate to them.  I think the only change I have to on make is to add the appropriate annotations on the existing Environment type.  Does this make sense?  If so, I will prepare a PR with the change.

[1]: https://github.com/mitchellh/packer/issues/3648
[2]: https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go

boumenot:
Add serialization for the Environment type based on the style of the
Azure/azure-sdk-for-go library.  One could already serialize the types,
but this makes it explicit, and in our approved style.

Closes #72

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -11,20 +11,20 @@ const (
 
 // Environment represents a set of endpoints for each of Azure's Clouds.
 type Environment struct {
-	Name                      string
-	ManagementPortalURL       string
-	PublishSettingsURL        string
-	ServiceManagementEndpoint string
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
LGTM

PR #74, opened 2016-06-28
Add German Cloud

boumenot:
Correct some of the existing URLs.

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -55,12 +55,12 @@ var (
 		ResourceManagerEndpoint:   "https://management.usgovcloudapi.net",
 		ActiveDirectoryEndpoint:   "https://login.microsoftonline.com/",
 		GalleryEndpoint:           "https://gallery.usgovcloudapi.net/",
-		KeyVaultEndpoint:          "https://vault.azure.net/",
+		KeyVaultEndpoint:          "https://vault.usgovcloudapi.net/",
 		GraphEndpoint:             "https://graph.usgovcloudapi.net/",
 		StorageEndpointSuffix:     "core.usgovcloudapi.net",
 		SQLDatabaseDNSSuffix:      "database.usgovcloudapi.net",
-		TrafficManagerDNSSuffix:   "trafficmanager.net",
-		KeyVaultDNSSuffix:         "vault.azure.net",
+		TrafficManagerDNSSuffix:   "usgovtrafficmanager.net",
+		KeyVaultDNSSuffix:         "vault.usgovcloudapi.net",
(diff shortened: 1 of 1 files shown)


colemickens:
LGTM. Thanks for the fixes.

boumenot:
Thank you!  I think I am still missing a few more things.  I wish all of the SDK took values from an auto-generated source.  There is some divergence on what the values are between.  This is mostly in the form of re-direct links vs. exact links.  I am not sure why.

colemickens:
Thinking about this... it would be neat if there was a discovery endpoint for each cloud that returned this list of endpoints. Even if we wanted to have them statically listed in code rather than making apps dynamically discover the endpoints every time... it would make the initial consumption easier.

colemickens:
Or maybe even better, a single public endpoint for all clouds that list out the public clouds and their endpoints...

ahmetalpbalkan:
@colemickens that can be done simply by uploading a static json blob somewhere. However, if it gets hacked and somebody tampers with endpoints, CLIs will get hacked too. Also things like if this endpoint is going to be accessible from restricted zones such as China...

boumenot:
@colemickens - I would like to see something like this too.  There are a large number of URIs to keep track of, and having **one** would be nice.

I think we should open an issue against autorest to have them create a JSON blob of the URIs, and see if that could be fed into the auto-generate process.  There is a lot similarity across the SDKs, but the slight differences in naming and availability can be annoying.

colemickens:
For auto-discovery and not baking in the URLs, yeah, I'm not advocating that- more magic than is really needed, firewall / external connectivity concerns, etc. Sure.

But even for the creation of these Environment objects... I had to look in a sort of silly number of places to find them. And even then it looks like I missed some. (IIRC for Fairfax the .NET SDK had some but not all, MSDN had a different subset and then I had to look elsewhere for Mooncake. And actually... at the time... Blackforest was not supposed to be talked about so I omitted it, I didn't know the Blackforest ones were documented anywhere publicly actually)

Sure, some static manifest of endpoints hosted somewhere could be hacked, but so could the actual ARM endpoints...

Honestly, even just having a `Microsoft/azure-clouds` repo with a "endpoints.txt" would make me happy. Or I guess an MSDN page but that just brings us back to our general documentation problems.

colemickens:
@boumenot yes, that's exactly what I want. Some "config" file that can be used to auto-generate the Environment struct, and the Environment instances for each cloud. This "Bing for the URLs" or "find an internal sharepoint where some subset is enumerated" doesn't feel right.

colemickens:
https://github.com/Azure/autorest/issues/1204

PR #80, opened 2016-07-15
azure: add EnvironmentFromName

colemickens:
I need this too, so I thought to make it a helper in autorest/azure.

Rather than being so lenient, it expects exact values.

Curious about:

1. If these are the exact values we want (again, I contend we need to get ARM/someone to publish an exact guideline for this).

2. If we want to fix noun/adjective agreement

3. If we want to treat empty string as public cloud? Or if that should just be handled by caller.

CC: @boumenot @ahmetalpbalkan @paulmey

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -3,11 +3,19 @@ package azure
 import (
 	"fmt"
 	"net/url"
+	"strings"
 )
 
 const (
 	activeDirectoryAPIVersion = "1.0"
 )
+
+var environments = map[string]Environment{
+	"AZURECHINACLOUD":        ChinaCloud,
+	"AZUREGERMANCLOUD":       GermanCloud,
+	"AZUREPUBLICCLOUD":       PublicCloud,
+	"AZUREUSGOVERNMENTCLOUD": USGovernmentCloud,
+}
 
 // Environment represents a set of endpoints for each of Azure's Clouds.
 type Environment struct {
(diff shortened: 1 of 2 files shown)


boumenot:
LGTM.

autorest/azure/environments.go
```diff
@@ -101,6 +102,26 @@ var (
 	}
 )
 
+// EnvironmentFromName returns an Environment based on the common name specified
+func EnvironmentFromName(name string) (Environment, error) {
+	var env Environment
+	name = strings.ToUpper(name)
+	switch name {
```
ahmetalpbalkan: I prefer this to be just a simple map. Switch seems like an overkill here and makes the code longer.

colemickens:
@ahmetalpbalkan changed impl to map as suggested, ptal.

autorest/azure/environments.go
```diff
@@ -3,12 +3,24 @@ package azure
...
+	"strings"
 )
 
 const (
 	activeDirectoryAPIVersion = "1.0"
 )
 
+var environments map[string]Environment
+
+func init() {
```
ahmetalpbalkan: techically, no need to have`init()` you can just assign value in `var` declaration above.

autorest/azure/environments.go
```diff
@@ -101,6 +113,16 @@ var (
 	}
 )
 
+// EnvironmentFromName returns an Environment based on the common name specified
+func EnvironmentFromName(name string) (Environment, error) {
+	name = strings.ToUpper(name)
+	env, ok := environments[name]
+	if !ok {
+		return env, fmt.Errorf("There is no cloud envionment matching the name '%s'!", name)
```
ahmetalpbalkan: typo:envionment

ahmetalpbalkan: no need for `!` also instead of single quotes consider using `%q`

ahmetalpbalkan: hmm I wish we were starting errors like `autorest/azure:` but I realized many files in this pkg already don't do it. so it is fine.

colemickens:
Thank you @ahmetalpbalkan. Updated again and green on Travis.

ahmetalpbalkan:
way cleaner. LGTM

PR #82, opened 2016-07-21
WithMultipartFormdata and WithFile

garimakhulbe:
Added WithMultiPartFormdata.
Added WithFile for sending file in request body.

### DIFF
--- a/autorest/preparer.go
+++ b/autorest/preparer.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"mime/multipart"
 	"net/http"
 	"net/url"
 	"strings"
@@ -191,6 +193,64 @@ func WithFormData(v url.Values) PrepareDecorator {
 				s := v.Encode()
 				r.ContentLength = int64(len(s))
 				r.Body = ioutil.NopCloser(strings.NewReader(s))
+			}
+			return r, err
+		})
+	}
+}
+
+// WithMultiPartFormData returns a PrepareDecoratore that "URL encodes" (e.g., bar=baz&foo=quux) form parameters
+// into the http.Request body.
+func WithMultiPartFormData(formDataParameters map[string]interface{}) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			r, err := p.Prepare(r)
+			if err == nil {
+				var body bytes.Buffer
+				writer := multipart.NewWriter(&body)
+				for key, value := range formDataParameters {
+					if rc, ok := value.(io.ReadCloser); ok {
+						var fd io.Writer
+						if fd, err = writer.CreateFormFile(key, key); err != nil {
+							return r, err
+						}
+						if _, err = io.Copy(fd, rc); err != nil {
+							return r, err
+						}
+					} else {
+						if err = writer.WriteField(key, ensureValueString(value)); err != nil {
+							return r, err
+						}
+					}
+				}
+				if err = writer.Close(); err != nil {
+					return r, err
+				}
+				if r.Header == nil {
+					r.Header = make(http.Header)
+				}
+				r.Header.Set(http.CanonicalHeaderKey(headerContentType), writer.FormDataContentType())
+				r.Body = ioutil.NopCloser(bytes.NewReader(body.Bytes()))
+				r.ContentLength = int64(body.Len())
+				return r, err
+			}
+			return r, err
+		})
+	}
+}
+
+// WithFile returns a PrepareDecorator that sends file in request body.
+func WithFile(f io.ReadCloser) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			r, err := p.Prepare(r)
+			if err == nil {
+				b, err := ioutil.ReadAll(f)
+				if err != nil {
+					return r, err
+				}
+				r.Body = ioutil.NopCloser(bytes.NewReader(b))
+				r.ContentLength = int64(len(b))
 			}
 			return r, err
 		})
--- a/autorest/preparer_test.go
+++ b/autorest/preparer_test.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/Azure/go-autorest/autorest/mocks"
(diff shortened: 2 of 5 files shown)


autorest/azure/persist_test.go
```diff
@@ -133,11 +133,13 @@ func TestSaveToken(t *testing.T) {
 
 func TestSaveTokenFailsNoPermission(t *testing.T) {
 	pathWhereWeShouldntHavePermission := "/usr/thiswontwork/atall"
+	expectedSubstring := "failed to create directory"
+
 	if runtime.GOOS == "windows" {
 		pathWhereWeShouldntHavePermission = "c:\\windows\\system32\\mytokendir\\mytoken"
+		expectedSubstring = "failed to create the temp file to write the token"
```
colemickens: This is fairly confusing to me. Your windows user account has permissions to `mkdir` on `C:\\windows\\system32\\mytokendir` but unable to write a new file inside? You didn't create the `mytokendir` as an admin user did you? Just confused as to what's going on here. As-is it's currently testing different errors on Linux/Windows.

colemickens:
What's the background on url.Values -> map[string]interface{} everywhere?

1. Is it not possible to do the conversion inside `WithFormData` rather than requiring every caller to change their code for this breaking change?

2. why `map[string]interface{}` and not `map[string][]string`? That's what `url.Values` actually is anyway... from `pkg/net/url`: `type Values map[string][]string`

autorest/preparer.go
```diff
@@ -197,6 +203,66 @@ func WithFormData(v url.Values) PrepareDecorator {
...
+						var fd io.Writer
+						if fd, err = writer.CreateFormFile(key, key); err != nil {
+							goto Return
+						}
+						if _, err = io.Copy(fd, v); err != nil {
+							goto Return
+						}
+					default:
+						if err = writer.WriteField(key, ensureValueString(v)); err != nil {
+							goto Return
```
ahmetalpbalkan: why `goto`? can't you just repeat the return statement?

autorest/preparer.go
```diff
@@ -197,6 +203,66 @@ func WithFormData(v url.Values) PrepareDecorator {
...
+// into the http.Request body.
+func WithMultiPartFormData(formDataParameters map[string]interface{}) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			r, err := p.Prepare(r)
+			if err == nil {
+				body := &bytes.Buffer{}
+				writer := multipart.NewWriter(body)
+				for key, value := range formDataParameters {
+					switch v := value.(type) {
```
ahmetalpbalkan: no need to `switch := ...(type)` here. just use `if rc, ok := v.(io.ReadCloser); ok { ...} else { ...}`

autorest/preparer.go
brendandixon: @garimakhulbe `goto` in go is seldom needed.

autorest/azure/devicetoken.go
```diff
@@ -85,7 +84,7 @@ func InitiateDeviceAuth(client *autorest.Client, oauthConfig OAuthConfig, client
 		autorest.AsPost(),
 		autorest.AsFormURLEncoded(),
 		autorest.WithBaseURL(oauthConfig.DeviceCodeEndpoint.String()),
-		autorest.WithFormData(url.Values{
+		autorest.WithFormData(map[string]interface{}{
 			"client_id": []string{clientID},
```
brendandixon: @garimakhulbe I know @colemickens raised this as well, but I'm not sure I see the benefit of moving from `url.Values` to a `map[string]interface{}`. Was there an issue in using `url.Values`.

garimakhulbe:
@colemickens @brendandixon 
What's the background on url.Values -> map[string]interface{} everywhere?
--> _All other decorators are using ```map[string]interface{}``` so to keep this consistent in all._
```Go
WithQueryParameters(queryParameters map[string]interface{})
WithPathParameters(path string, pathParameters map[string]interface{})
```
_Also, in code generation, we don't need to add extra code for parameters to url.Values change as its being handle inside WithFormData. But I got your point that this will be a breaking change and if we could avoid this. My suggestion is below._

1. Is it not possible to do the conversion inside `WithFormData` rather than requiring every caller to change their code for this breaking change?
--> _Suggested below. I think this is what you are referring to._

2. why `map[string]interface{}` and not `map[string][]string`? That's what `url.Values` actually is anyway... from `pkg/net/url`: `type Values map[string][]string`
--> _Used ``ensureValueString()`` method in WithFormData() to convert map[string]interface{} to url.Values but this will be removed if we return to url.Values._

How about returning to WithFormData(url.Values) and adding a method for conversion..
```Go
func SendInFormData( arg1 int,  arg2 string) {
    var v url.Values
    v := autorest.ConvertToValues(arg1, arg2) // method to convert parameters to url.Values() 
    WithFormData(v)

   //WithFormData(autorest.ConvertToValues(arg1, arg2)) -- better
}
```
??

brendandixon:
@garimakhulbe I appreciate consistency, but semantic meaning is more important. Since these values are always and only `url.Values` using that type conveys the correct meaning. Further, `url.Values` will protect the code in unusual cases.

I think that I agree with @colemickens: Our callers should not have to convert their data for us to use. We should do that on their behalf.

ahmetalpbalkan:
> How about returning to WithFormData(url.Values) and adding a method for conversion..

sounds good to me to have something that does the map[string]interface{} to url.Values.

garimakhulbe:
@brendandixon @ahmetalpbalkan @colemickens All comments are addressed. Please take a look. Thanks.

autorest/preparer.go
```diff
@@ -197,6 +199,64 @@ func WithFormData(v url.Values) PrepareDecorator {
...
 }
 
+// WithMultiPartFormData returns a PrepareDecoratore that "URL encodes" (e.g., bar=baz&foo=quux) form parameters
+// into the http.Request body.
+func WithMultiPartFormData(formDataParameters map[string]interface{}) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			r, err := p.Prepare(r)
+			if err == nil {
+				body := &bytes.Buffer{}
```
ahmetalpbalkan: initialize as `var body bytes.Buffer` and parse as `&body` where you need to. it's more idiomatic.

autorest/preparer.go
```diff
@@ -197,6 +199,64 @@ func WithFormData(v url.Values) PrepareDecorator {
...
+}
+
+// WithFile returns a PrepareDecorator that sends file in request body.
+func WithFile(f io.ReadCloser) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			r, err := p.Prepare(r)
+			if err == nil {
+				var b []byte
+				if b, err = ioutil.ReadAll(f); err != nil {
```
ahmetalpbalkan: it is equally readable if you do

b, err := ...
if err != nil

autorest/preparer_test.go
```diff
@@ -442,6 +443,66 @@ func TestWithFormDataSetsContentLength(t *testing.T) {
 	}
 }
 
+func TestWithMultiPartFormDataSetsContentLength(t *testing.T) {
+	v := map[string]interface{}{
+		"file": ioutil.NopCloser(bytes.NewReader([]byte("\xbd\xb2\x3d\xbc\x20\xe2\x8c\x98"))),
```
ahmetalpbalkan: any comment about  what this binary data is?

autorest/utility_test.go
```diff
@@ -239,6 +240,36 @@ func TestEncodeWithValidNotPathQuery(t *testing.T) {
...
+	v := url.Values{
+		"a": []string{"a"},
+		"b": []string{"2"},
+	}
+	if !reflect.DeepEqual(MapToValues(m), v) {
+		t.Fatalf("autorest: MapToValues method failed to return correct values - expected(%v) got(%v)", v, MapToValues(m))
+	}
+}
+
+func TestMapToValuesWithArrayValues(t *testing.T) {
```
ahmetalpbalkan: I'm not sure if this test is guaranteed to pass tomorrow. Can you please validate if url.Values.Add preserves the order in which you add multiple values to the same key?

ahmetalpbalkan: indeed this test might be passing because you're initializing `v` with a map literal yourself, try using url.Values.Add, it might fail.

so I think you should test this with `url.Values.Get` (same for `TestMapToValues` above) with each key.

garimakhulbe: Godoc mentions that ```Add()``` appends to any existing values associated with key but I have updated the tests to check values assuming order is not preserved.

autorest/preparer_test.go
garimakhulbe: copy-pasted from some go example :). Now changed to string.

autorest/preparer.go
garimakhulbe: done.

autorest/preparer.go
garimakhulbe: done.

autorest/preparer_test.go
```diff
@@ -437,11 +438,76 @@ func TestWithFormDataSetsContentLength(t *testing.T) {
 		t.Fatalf("autorest: WithFormData failed with error (%v)", err)
 	}
 
+	expected := "age=42&name=Rob+Pike"
```
ahmetalpbalkan: also, you might not get these in this order. I've seen ordering of map keys in for loops change between golang versions.

ahmetalpbalkan:
LGTM

autorest/preparer_test.go
colemickens: I believe it is the case that iteration ordering is intentionally randomized to prevent people from taking a dependency on it.

https://blog.golang.org/go-maps-in-action Ctrl-F "Since Go 1"...

garimakhulbe: updated.

autorest/preparer_test.go
```diff
@@ -437,11 +439,82 @@ func TestWithFormDataSetsContentLength(t *testing.T) {
...
+		t.Fatalf("autorest:WithFormData failed to return correct string got (%v), expected (%v)", string(b), expected)
+	}
+
 	if r.ContentLength != int64(len(b)) {
 		t.Fatalf("autorest:WithFormData set Content-Length to %v, expected %v", r.ContentLength, len(b))
 	}
 }
 
+func sortString(x string) string {
+	s := strings.Split(x, "")
```
ahmetalpbalkan: shouldn't you split by &? for Rob Pike example above I'd just do if

    a == "name=Rob+Pike&age=42" || a == "age=42&name=Rob+Pike"

and that'd solve the problem

garimakhulbe:
Thanks for the comments.

PR #84, opened 2016-07-26
Add RetryDuration in client

garimakhulbe:


autorest/client.go
```diff
@@ -163,7 +167,7 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		return nil, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
 	resp, err := SendWithSender(c.sender(), r,
-		DoRetryForStatusCodes(c.RetryAttempts, defaultRetryInterval, statusCodesForRetry...))
```
ahmetalpbalkan: is `defaultRetryInterval` still being used?

garimakhulbe: No, we are not using it anywhere else.

garimakhulbe: I will remove this variable.

colemickens:
Is it worth exposing this on a per-request basis? So I could get the Sender and call `Do()` or `DoWithRetryInterval(someInterval)`. For requests that I know get done faster than the RP says to retry on, but without affecting the entire client?

Might not be worth it.

This change LGTM either way.

Issue #83, opened 2016-07-22
Test tries to access the network


TheTincho:
Hi!

While all other tests use mocks correctly to avoid depending on DNS or connectivity to the internet, `TestClientDoSetsUserAgent` is missing such mock, and triggers a DNS query (this is a [bug in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=830672)). This small patch fixes that:

```
--- a/autorest/client_test.go
+++ b/autorest/client_test.go
@@ -151,6 +151,8 @@
    ua := "UserAgent"
    c := Client{UserAgent: ua}
    r := mocks.NewRequest()
+   s := mocks.NewSender()
+   c.Sender = s

    c.Do(r)

```

ahmetalpbalkan:
@TheTincho thanks for discovering. Do you mind sending a pull request?

TheTincho:
Sorry, I did not see your reply!

Will do, along with another test that was reported as accessing the network.

jhendrixMSFT:
PR https://github.com/Azure/go-autorest/pull/97 to fix the remaining test that hits the network.

jhendrixMSFT:
The fix has been merged.

PR #88, opened 2016-09-25
Prevent tests from accessing the network.

TheTincho:
As reported in #83, there are some tests that are accessing the network, which is a policy violation in Debian, and in general not a good idea. This PR fixes two of these tests.

Please note that there is another test in token_test.go (TestServicePrincipalTokenWithAuthorizationReturnsErrorIfCannotRefresh) that is also accessing the network but I was not able to fix, as I don't really know the autorest API, I am just the Debian maintainer :-)


ahmetalpbalkan:
cc: @garimakhulbe

garimakhulbe:
@mcardosos Please take a look and merge. This looks fine to me as well.

jhendrixMSFT:
Fix looks good, thanks for the PR.  👍

PR #93, opened 2016-10-13
Add VM DNS Suffix

boumenot:


garimakhulbe:
LGTM.
@jhendrixMSFT @mcardosos please take a look.

colemickens:
One sec, there's another suffix that might make sense to add here as well. For example, I have this attached to a Public IP right now: `mickens57fe9eeb.eastus.cloudapp.azure.com`

I *think* the difference might be that this is attached to an LB rather than a compute VM? Either way, it differs from the `cloudapp.net` suffix being added here. Not sure if ASM is a consideration as well too?

Does anyone know before I keep digging in? We don't have to hold this PR up either necessarily... just noticed it.

boumenot:
I reviewed the list of suffixes, and I cannot find anything about cloudapp.azure.com.  Have you assigned a name to this VM, or is this the one that was auto-assigned by Azure?

colemickens:
```
   6 {
   5       "apiVersion": "[variables('apiVersionDefault')]",
   4       "location": "[variables('location')]",
   3       "name": "[variables('masterPublicIPAddressName')]",
   2       "properties": {
   1         "dnsSettings": {
1350           "domainNameLabel": "[variables('masterFqdnPrefix')]"
   1         },
   2         "publicIPAllocationMethod": "Dynamic"
   3       },
   4       "type": "Microsoft.Network/publicIPAddresses"
   5     },
```

`masterFqdnPrefix` was set to `mickens57fe9eeb` and this is the resulting DNS name that is assigned to the public ip: `mickens57fe9eeb.eastus.cloudapp.azure.com`.

Can you point me to the list you were looking at?

colemickens:
Yup, it's an ASM/ARM difference. https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-windows-compare-deployment-models/

Ctrl-f: `cloudapp.azure.com`.

boumenot:
Nice!

The info I was reading is internal.  Just search for black forest and cloudapp.net to get the break down of sovereign cloud URLs.

jhendrixMSFT:
LGTM, thanks for the submission.

colemickens:
Why was this merged? It should have been changed so that it was obvious that the suffix applied to ASM only.

colemickens:
Please send a follow up PR renaming this field and/or adding another field for the ARM/CRP suffix.

boumenot:
@colemickens what do you want to change it to? How about?

 * AsmVMDNSSuffix - cloudapp.net
 * ArmVMDNSSuffix - cloudapp.azure.com

colemickens:
Technically ARM/ASM are acronyms and we ought to capitalize them, but that would look awful.  Only other alternative would be "ServiceManagementVMDNSSuffix" and "ResourceManagerVMDNSSuffix" but those aren't any better or helpful.

What you suggested is fine with me.

boumenot:
.NET has better conventions for naming IMHO, but I digress.  

I like your names betters, and there is already precedence for them in the code.  I'll submit a PR with new names.

colemickens:
Thanks Chris!

PR #94, opened 2016-10-13
Add ASM and ARM VM DNS suffixes

boumenot:
/cc: @colemickens


colemickens:
(GitHub doesn't do any favors for reviewing these with these whitespace changes...)

Anyway, LGTM. Thanks again.

boumenot:
I could not find docs for most of the ARM suffixes, so they may be wrong.  ARM is new to Fairfax, which may be why I couldn't find the appropriate docs.

boumenot:
If you append ?w=1 to the URL, you will get some love, but it's not intuitive.

PR #96, opened 2016-11-10
AUD Claim Fix to have string value

jkapil:
Currently, the token API is failing for cert based secrets. The failure is because of invalid aud claim. Fixing audience (aud) claim to have string value, rather than URL object

### DIFF
--- a/autorest/azure/token.go
+++ b/autorest/azure/token.go
@@ -138,7 +138,7 @@ func (secret *ServicePrincipalCertificateSecret) SignJwt(spt *ServicePrincipalTo
 	token := jwt.New(jwt.SigningMethodRS256)
 	token.Header["x5t"] = thumbprint
 	token.Claims = jwt.MapClaims{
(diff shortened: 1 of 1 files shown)


colemickens:
I believe this was a regression stemming from the upgrade of the jwt-go library.

Since [MapClaims is a `map[string]interface{}` rather than a `map[string]string`](https://github.com/dgrijalva/jwt-go/blob/24c63f56522a87ec5339cc3567883f1039378fdb/map_claims.go#L11), the URL  struct is not properly being coerced to a simple string.

colemickens:
Thanks for this pull request and fix @jkapil 

LGTM. @mcardosos, I still have merge permissions, do you mind me merging PRs like this? If you'd rather handle it, that's fine... but can we get this merged ASAP?

Thanks all.

mcardosos:
I don't mind. As soon as Travis passes feel free to merge.

PR #97, opened 2016-11-10
Fix for a few test bugs.

jhendrixMSFT:
Fixed hard-coded assumption on windir being "c:\windows".
The test "TestSaveTokenFailsCantCreate" now passes on Windows OS.
Fixed token test to use a mock so it doesn't hit the network.


jhendrixMSFT:
@mcardosos Please take a look.

mcardosos:
LGTM

PR #99, opened 2016-11-15
Adding support for non hard-coded auth-endpoint.

jkapil:


colemickens:
Is this for Azure Stack? How do other SDKs handle discovery of endpoints of various things. The Environment struct has a lot of things that will be different from Azure Stack installs. I would have expected there to be a discovery mechanism in Azure Stack, such that the correct Environment struct could be populated at runtime. But I'm guessing that doesn't exist... 

Anyway, change looks fine to me, though I lament needing it.

jkapil:
This change is for MSI extension. The application will choose the auth endpoint and pass it as input to MSI extension. MSI extension will use that endpoint as input. This change also helps azure stack as well.

Yes, azure services provided discovery mechanism. The authentication challenge gives the user the authendpoint and resource to be used for getting token. This authendpoint will be provided by application to MSI extension.

jkapil:
Why the CI failed. There is no change made to these files.
autorest/azure/token.go:94:20: error strings should not be capitalized or end with punctuation or a newline
autorest/mocks/mocks.go:33:24: error strings should not be capitalized or end with punctuation or a newline

colemickens:
cc @mcardosos for the CI failure.

jkapil:
@mcardosos , can you help with CI failure. The CI failure is on the files not changed by this commit.

mcardosos:
Hello, this failure is because of some recent changes in golint (fixed this in another PR). Could you restart Travis?

jkapil:
all checks passed.

PR #101, opened 2016-11-16
New responder to unmarshal success or error response at one place.

jkapil:
Adding responder which can deal with both success and error response. This will also help to consolidate usage of response.Body at one place. Depending on API result, response body is unmarshalled to either success type or error type.

### DIFF
--- a/autorest/azure/token.go
+++ b/autorest/azure/token.go
@@ -302,7 +302,7 @@ func (spt *ServicePrincipalToken) refreshInternal(resource string) error {
 
 	var newToken Token
 	err = autorest.Respond(resp,
-		autorest.WithErrorUnlessOK(),
+		autorest.WithErrorUnlessStatusCode(http.StatusOK),
 		autorest.ByUnmarshallingJSON(&newToken),
 		autorest.ByClosing())
 	if err != nil {
--- a/autorest/error.go
+++ b/autorest/error.go
@@ -28,6 +28,9 @@ type DetailedError struct {
 
 	// Message is the error message.
 	Message string
+
+	// Service Error is the response body of failed API in bytes
+	ServiceError []byte
 }
 
 // NewError creates a new Error conforming object from the passed packageType, method, and
--- a/autorest/responder.go
+++ b/autorest/responder.go
@@ -165,17 +165,24 @@ func ByUnmarshallingXML(v interface{}) RespondDecorator {
 }
 
 // WithErrorUnlessStatusCode returns a RespondDecorator that emits an error unless the response
-// StatusCode is among the set passed. Since these are artificial errors, the response body
-// may still require closing.
+// StatusCode is among the set passed. On error, response body is fully read into a buffer and
+// presented in the returned error, as well as in the response body.
 func WithErrorUnlessStatusCode(codes ...int) RespondDecorator {
 	return func(r Responder) Responder {
 		return ResponderFunc(func(resp *http.Response) error {
 			err := r.Respond(resp)
 			if err == nil && !ResponseHasStatusCode(resp, codes...) {
-				err = NewErrorWithResponse("autorest", "WithErrorUnlessStatusCode", resp, "%v %v failed with %s",
+				derr := NewErrorWithResponse("autorest", "WithErrorUnlessStatusCode", resp, "%v %v failed with %s",
 					resp.Request.Method,
 					resp.Request.URL,
 					resp.Status)
+				if resp.Body != nil {
+					defer resp.Body.Close()
+					b, _ := ioutil.ReadAll(resp.Body)
+					derr.ServiceError = b
+					resp.Body = ioutil.NopCloser(bytes.NewReader(b))
+				}
+				err = derr
 			}
 			return err
 		})
--- a/autorest/responder_test.go
+++ b/autorest/responder_test.go
@@ -2,6 +2,7 @@ package autorest
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -420,6 +421,61 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
 	err := Respond(nil)
 	if err != nil {
 		t.Fatalf("autorest: Respond returned an unexpected error when given a null Response (%v)", err)
+	}
+}
+
+func TestWithErrorUnlessStatusCodeOKResponse(t *testing.T) {
+	v := &mocks.T{}
+	r := mocks.NewResponseWithContent(jsonT)
+	err := Respond(r,
+		WithErrorUnlessStatusCode(http.StatusOK),
+		ByUnmarshallingJSON(v),
+		ByClosing())
+
+	if err != nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) failed on okay response. (%v)", err)
+	}
+
+	if v.Name != "Rob Pike" || v.Age != 42 {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) corrupted the response body of okay response.")
+	}
+}
+
+func TesWithErrorUnlessStatusCodeErrorResponse(t *testing.T) {
+	v := &mocks.T{}
+	e := &mocks.T{}
+	r := mocks.NewResponseWithContent(jsonT)
+	r.Status = "400 BadRequest"
+	r.StatusCode = http.StatusBadRequest
+
+	err := Respond(r,
+		WithErrorUnlessStatusCode(http.StatusOK),
+		ByUnmarshallingJSON(v),
+		ByClosing())
+
+	if err == nil {
+		t.Fatal("autorest: WithErrorUnlessStatusCode(http.StatusOK) did not return error, on a response to a bad request.")
+	}
+
+	var errorRespBody []byte
+	if derr, ok := err.(DetailedError); !ok {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) got wrong error type : %T, expected: DetailedError, on a response to a bad request.", err)
+	} else {
+		errorRespBody = derr.ServiceError
+	}
+
+	if errorRespBody == nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) ServiceError not returned in DetailedError on a response to a bad request.")
+	}
+
+	err = json.Unmarshal(errorRespBody, e)
+	if err != nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) cannot parse error returned in ServiceError into json. %v", err)
+	}
+
+	expected := &mocks.T{Name: "Rob Pike", Age: 42}
+	if e != expected {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK wrong value from parsed ServiceError: got=%#v expected=%#v", e, expected)
 	}
 }
 


colemickens:
cc: @ahmetalpbalkan

jkapil:
$ test -z "$(gofmt -s -l -w ./autorest/. | tee /dev/stderr)"
autorest/azure/azure.go
autorest/responder.go
The command "test -z "$(gofmt -s -l -w ./autorest/. | tee /dev/stderr)"" exited with 1.
$ test -z "$(golint ./autorest/... |  tee /dev/stderr)"

Where can i get the output of above commands on travis output. I am not able to run them on my windows machine.

colemickens:
Just run the `golint` command. `test -z` just checks to see if the output is empty.

`go get -u github.com/golang/lint/golint` will download/build `golint` and put it in `$GOPATH/bin` for you.

jkapil:
problem seems with gofmt -s -l -w ./autorest/. Works fine on my machine(windows). Let me see if this has anything to do with line ending charaters

colemickens:
Looks like you pushed another commit that fixed it? Travis looks green anyway.

I'd like @ahmetalpbalkan to take a look since he did similar work for the AzureError. If he doesn't look by tomorrow afternoon, we can go ahead and merge.

Thanks for the PRs!

autorest/azure/azure.go
```diff
@@ -28,6 +28,19 @@ const (
...
+// AADServiceError encapsulates the error response from Azure Token Service.
+type AADServiceError struct {
+	ErrorTitle       string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+	ErrorCodes       []int  `json:"error_codes"`
+	TraceID          string `json:"trace_id"`
+	CorrelationID    string `json:"correlation_id"`
+}
+
+func (aade *AADServiceError) Error() string {
```
ahmetalpbalkan: there's not much point of defining this on the pointer type.

autorest/azure/token.go
```diff
@@ -301,14 +301,17 @@ func (spt *ServicePrincipalToken) refreshInternal(resource string) error {
 	}
 
 	var newToken Token
+	var aadError AADServiceError
 	err = autorest.Respond(resp,
 		autorest.WithErrorUnlessOK(),
-		autorest.ByUnmarshallingJSON(&newToken),
+		autorest.ByUnmarshallingSuccessandErrorJSON(&newToken, &aadError),
```
ahmetalpbalkan: are you sure this will work? because autorest.WithErrorUnlessOK already returns an error, so your `autorest.ByUnmarshallingSuccessandErrorJSON` most likely won't get invoked if response.Code!=200.

colemickens:
cc: @brendandixon

autorest/responder.go
```diff
@@ -142,6 +142,30 @@ func ByUnmarshallingJSON(v interface{}) RespondDecorator {
 	}
 }
 
+// ByUnmarshallingSuccessandErrorJSON returns a RespondDecorator that decodes a JSON document returned in the
+// response Body into the value pointed by success schema v or error schema e
+func ByUnmarshallingSuccessandErrorJSON(v interface{}, e interface{}) RespondDecorator {
```
brendandixon: The "and" needs to be capitalized.

autorest/azure/token.go
brendandixon: @ahmetalpbalkan is correct. I suggest you create a Responder, such as WithJAADServiceErrorUnlessOK, that unpacks the Body into an AADServiceError. Then use the normal ByUnmarshallingJSON to unpack the retrieved token.

autorest/responder.go
brendandixon: Disregard. The method needs to be renamed to `WithAADServiceErrorUnlessOK` and modified to unpack the body into an `AADServiceError` if the return code is not 200.

autorest/azure/azure.go
jkapil: oh didn't knew there is a different meaning for pointer type and non pointer type, I assumed this as syntax for adding member function to a struct. Let me revisit this.

autorest/azure/token.go
jkapil: This works. I assumed that it follows standard message handling, where responder respond calls previous responder respond. So autorest.ByUnmarshallingSuccessandErrorJSON would have called  autorest.WithErrorUnlessOK.

brendandixon: @jkapil I **strongly** believe this needs to be changed to `WithAADServiceErrorUnlessOK` as I suggested. That is much more in keeping with the go-autorest model.

jkapil: oh, i haven't looked at that comment. I am not contesting that - still reviewing. I was answering to @ahmetalpbalkan concern (are you sure this will work?) I will send another iteration for review.

autorest/responder.go
jkapil: sure

autorest/azure/azure.go
jkapil: it seems without pointers corresponds to const functions. I will make that change

autorest/azure/token.go
```diff
@@ -303,15 +303,12 @@ func (spt *ServicePrincipalToken) refreshInternal(resource string) error {
 	var newToken Token
 	var aadError AADServiceError
 	err = autorest.Respond(resp,
-		autorest.WithErrorUnlessOK(),
-		autorest.ByUnmarshallingSuccessandErrorJSON(&newToken, &aadError),
+		autorest.WithServiceErrorUnlessOK(&newToken, &aadError),
```
brendandixon: @jkapil You should not have to pass in the `aadError`. `WithServiceErrorUnlessOK` can just return it as long as it "inherits" from `error`. I believe we do this elsewhere.

autorest/responder.go
```diff
@@ -142,23 +142,34 @@ func ByUnmarshallingJSON(v interface{}) RespondDecorator {
...
+				b, errInner := ioutil.ReadAll(resp.Body)
 				if errInner != nil {
-					err = fmt.Errorf("Error occurred unmarshalling JSON - Error = '%v' JSON = '%s'", errInner, string(b))
+					err = fmt.Errorf("Error occurred reading http.Response#Body - Error = '%v'", errInner)
+				} else if len(strings.Trim(string(b), " ")) > 0 {
+					if ok {
+						errInner = json.Unmarshal(b, v)
+					} else {
+						errInner = json.Unmarshal(b, e)
+					}
```
brendandixon: @jkapil I feel this code would be simpler if it used the if-clause to select the target `interface` and then made one call to `json.Unmarshal`. Since both are interfaces, there is no need to make separate calls and the logic would then be clear (that is, it would show it is picking the target for the unmarshal).

autorest/responder.go
```diff
@@ -142,23 +142,34 @@ func ByUnmarshallingJSON(v interface{}) RespondDecorator {
 	}
 }
 
-// ByUnmarshallingSuccessandErrorJSON returns a RespondDecorator that decodes a JSON document returned in the
+// WithServiceErrorUnlessOK returns a RespondDecorator that decodes a JSON document returned in the
 // response Body into the value pointed by success schema v or error schema e
-func ByUnmarshallingSuccessandErrorJSON(v interface{}, e interface{}) RespondDecorator {
+func WithServiceErrorUnlessOK(v interface{}, e interface{}) RespondDecorator {
 	return func(r Responder) Responder {
```
brendandixon: @jkapil I like brevity, but I'm not not excited by using `v` and `e` as variable names. It feels like too much information is lost.

autorest/responder.go
jkapil: sure

autorest/responder.go
jkapil: i tried to be consistent. I didn't liked v either. I will make it more readable.

jkapil:
closing the pull request, will send new after squashing all commits into one.

ahmetalpbalkan:
@jkapil please try to reuse existing pull request and squash commits locally on your machine, then `git push -f` here. There's no point of a separate pull request.

jkapil:
Okay, reopening. Will perform squash and force push in the morning

jkapil:
@ahmetalpbalkan  please review

autorest/azure/azure.go
```diff
@@ -28,6 +28,19 @@ const (
...
+type AADServiceError struct {
+	ErrorTitle       string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+	ErrorCodes       []int  `json:"error_codes"`
+	TraceID          string `json:"trace_id"`
+	CorrelationID    string `json:"correlation_id"`
+}
+
+func (aade AADServiceError) Error() string {
+	return fmt.Sprintf("Error=%q Message=%q", aade.ErrorTitle, aade.ErrorDescription)
```
ahmetalpbalkan: maybe prefix with `Azure Active Directory Error=...`

autorest/error.go
```diff
@@ -28,6 +28,9 @@ type DetailedError struct {
 
 	// Message is the error message.
 	Message string
+
+	// Service Error is the response body of failed API.
+	ServiceError interface{}
```
ahmetalpbalkan: why is response body `interface{}`? It doesn't sound like this is actually response body.

autorest/responder_test.go
```diff
@@ -423,6 +423,45 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+		ByClosing())
+	if err == nil {
+		t.Fatalf("autorest: WithServiceErrorUnlessOK failed to return error on bad request")
+	}
+
+	if e.Name != "Rob Pike" || e.Age != 42 {
+		t.Fatalf("autorest: WithServiceErrorUnlessOK failed to properly unmarshal error response")
+	}
+
+	derr, _ := err.(DetailedError)
```
ahmetalpbalkan: make sure you capture `, ok`  and check it's true.

autorest/azure/azure.go
```diff
@@ -178,3 +191,10 @@ func WithErrorUnlessStatusCode(codes ...int) autorest.RespondDecorator {
 		})
 	}
 }
+
+// WithAADServiceErrorUnlessOK returns a RespondDecorator that decodes a JSON document returned in the
+// response Body into the value pointed by success schema succesSchema or set AADServiceError to the DetailedError
+func WithAADServiceErrorUnlessOK(succesSchema interface{}) autorest.RespondDecorator {
```
ahmetalpbalkan: needs tests.

autorest/responder_test.go
```diff
@@ -423,6 +423,45 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+	r.StatusCode = http.StatusBadRequest
+
+	err := Respond(r,
+		WithServiceErrorUnlessOK(v, e),
+		ByClosing())
+	if err == nil {
+		t.Fatalf("autorest: WithServiceErrorUnlessOK failed to return error on bad request")
+	}
+
+	if e.Name != "Rob Pike" || e.Age != 42 {
```
ahmetalpbalkan: this test is not easy to follow. please do not share `mocks.T` type for both "value" and "error" types. You're not testing if things still get deserialized into `v`.

autorest/responder_test.go
```diff
@@ -423,6 +423,45 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+	if v.Name != "Rob Pike" || v.Age != 42 {
+		t.Fatalf("autorest: WithServiceErrorUnlessOK failed to properly unmarshal")
+	}
+}
+
+func TestWithServiceErrorUnlessOK_ErrorResponse(t *testing.T) {
+	v := &mocks.T{}
+	e := &mocks.T{}
+	r := mocks.NewResponseWithContent(jsonT)
+	r.Status = "400 BadRequest"
```
ahmetalpbalkan: `Bad Request`

autorest/responder_test.go
jkapil: yes, i am using testing e.

autorest/error.go
jkapil: this is the interface to which response body was unmarshalled. I need only the bytes and can let the caller consume it and do the unmarshalling. If you are okay with that I can make that change.

ahmetalpbalkan: @jkapil there's a discrepancy with the type and the comment. If you are unmarshaling the error into this object, it's perfectly fine; just don't call it "the response body".

jkapil:
Reducing the footprint of the code. When the response do not meet the input status code, the response body is copied as bytes into err. These bytes can be unmarshalled based on the API schema. @ahmetalpbalkan if we can't resolve this by tomorrow, I will schedule a meeting to resolve this.

ahmetalpbalkan:
> These bytes can be unmarshalled based on the API schema.

@jkapil by who? I see that you just made a major change that we did not agree upon. This doesn't look like the PR we iterated on before and I'm not sure why you suddenly decide to change it all.

Previously, you used to unmarshal error into a type and you were providing a proper `error` string from the error.  Now, you're expecting people to cast `error` into `DetailedError` and retrieve `[]byte` to unmarshal to a type that they don't know when there is an error? I am very confused.

The previous version was looking almost very good to me and the last comments I left were mostly about cosmetics. By just changing the entire PR (which was pretty much ready-to-merge) without really providing a reason, I don't think you are helping anyone. :)

garimakhulbe:
@mcardosos @jhendrixMSFT @vishrutshah Can you guys look into this? 

@jkapil Above is the SDK team who can help you in this.

mcardosos:
Adding @marstr

mcardosos:
Hello @jkapil !
I'm so sorry, but I cannot see your previous commits because of the force push (for me it is easier to understand changes / iterations in different commits, don't worry about squashing commits, that can be done by github when merging PRs).
I guess you were copying the contents from response body to `DetailedError.Message`? I think that would be the easiest way to include that in the error :)

jkapil:
Hi @mcardosos I am copying the response body as bytes when the response do not meet input status code (say 200, success). The schema of response body is different based on status code, usually two but not necessarily  - for example handshake API for authentication discovery of key vault has different response body schema.
I need a place to put the byes, which I am doing in DetailedError.  As a user of the get token API, I can unmarshal this based on response code's schema. In my case though there was no need to unmarshal and I needed bytes only.

jkapil:
Hi @ahmetalpbalkan I understand previous iteration was closer to your expectation, but it diverged from the way I wanted to achieve the change. As such I have reduced the footprint of the code to reduce the surface area of review. As per your question on "who" - who, is MSI extension.

jkapil:
following is the illustrative code from MSI extension
spt, err := getSptFromCertificate(*oauthConfig, cachedCreds.ClientId, params.Resource, pfxbytes, func(t azure.Token) error { return nil })
if derr, ok := err.(autorest.DetailedError); ok {
		statusCode = derr.StatusCode.(int)
		bytes = derr.ServiceError
}

colemickens:
Correct me if I'm wrong, but it looks like the latest version of this just adds a new responder that returns an error with the body bytes as well?

Why not just make that the behavior of `WithErrorUnlessStatusCode`? Seems like a reasonable thing to provide by default anyway. The caller can choose what do with the error body. My understanding is that is desired because parsing the error into a struct is difficult due to inconsistent format of error messages from AAD...

jkapil:
Yes we could change WithErrorUnlessStatusCode, but one of the strong code feedback earlier was to use different function.

ahmetalpbalkan:
I'm actually totally fine saving the response body as `[]byte` in `WithErrorUnlessStatusCode` if you're not going to introduce any AAD concepts into the library here. The previous feedback was saying: if you're introducing AADServiceError, better have WithAADServiceErrorUnlessOK for it.

brendandixon:
@jkapil @ahmetalpbalkan That was me. The prior version introduce an ugly if-predicate. Since this is Azure specific, let's keep it that way w/o burying an if-clause. I believe it better to have small parts you can assemble.

autorest/responder.go
```diff
@@ -166,16 +166,23 @@ func ByUnmarshallingXML(v interface{}) RespondDecorator {
 
 // WithErrorUnlessStatusCode returns a RespondDecorator that emits an error unless the response
 // StatusCode is among the set passed. Since these are artificial errors, the response body
-// may still require closing.
+// may still require closing. On error, the response body is copied and replaced.
```
ahmetalpbalkan: You should probably say, 

> On error, response body is fully read into a buffer and presented in the returned error, as well as in the response body.

Also 

> Since these are artificial errors, the response body may still require closing.

doesn't hold true as we now read the response body "always" when it is available.

autorest/responder_test.go
```diff
@@ -423,6 +424,60 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+		ByUnmarshallingJSON(v),
+		ByClosing())
+
+	if err == nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) failed to return error on non-okay response.")
+	}
+
+	var errorRespBody []byte
+	if derr, ok := err.(DetailedError); !ok {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) failed to return detailed error on non-okay response. (%v)", err)
```
ahmetalpbalkan: if `!ok` your `err` is going to be `nil`. So try this.

     t.Fatalf("got wrong type: %T, expected: DetailedError", err)

autorest/responder_test.go
```diff
@@ -423,6 +424,60 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+	r.Status = "400 BadRequest"
+	r.StatusCode = http.StatusBadRequest
+
+	err := Respond(r,
+		WithErrorUnlessStatusCode(http.StatusOK),
+		ByUnmarshallingJSON(v),
+		ByClosing())
+
+	if err == nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) failed to return error on non-okay response.")
```
ahmetalpbalkan: `... did not return an error, got nil` is way better.

autorest/responder_test.go
```diff
@@ -423,6 +424,60 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+	} else {
+		errorRespBody = derr.ServiceError
+	}
+
+	if errorRespBody == nil {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) failed to set ServiceError of detailed error on non-okay response.")
+	}
+
+	err = json.Unmarshal(errorRespBody, e)
+	if err == nil {
```
ahmetalpbalkan: shouldn't this be `err != nil`? also `corrupted the response body on non-okay response. ` is very cryptic. try this:

    t.Fatalf("cannot parse error returned in ServiceError into json: %v", err)

autorest/responder_test.go
```diff
@@ -423,6 +424,60 @@ func TestRespondAcceptsNullResponse(t *testing.T) {
...
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) corrupted the response body on non-okay response. (%v)", err)
+	}
+
+	if e.Name != "Rob Pike" || e.Age != 42 {
+		t.Fatalf("autorest: WithErrorUnlessStatusCode(http.StatusOK) corrupted the response body on non-okay response.")
```
ahmetalpbalkan: you are failing  with the same error here and above. just change it to:

```
if expected := T{Name:"Rob Pike", Age: 42}; e != expected {
    t.Fatalf("wrong value from parsed ServiceError: got=%#v expected=%#v", e, expected)
}
```

This will make it clear what's going on.

jkapil:
@ahmetalpbalkan , next week is last regular working week before xmas. Let's resolve any issues offline, so that this can be submitted before 12/19

Issue #102, opened 2016-12-02
azure: Environment should include the ACR endpoint suffix


colemickens:
https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L21

Please add the ACR suffix to this struct and then for each environment that is defined.

mcardosos:
Hey @colemickens
Do you happen to know which suffix each cloud uses? I couldn't find lots of info about it. Or all clouds use `azurecr.io`?

colemickens:
Cc @sajayantony

PR #104, opened 2016-12-15
Adding ACR suffix to pulic cloud env

mcardosos:
(USgov, China and Germany missing)

Related to [issue 102](https://github.com/Azure/go-autorest/issues/102).

Still incomplete PR.

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -78,6 +78,7 @@ var (
 		ServiceBusEndpointSuffix:     "servicebus.usgovcloudapi.net",
 		ServiceManagementVMDNSSuffix: "usgovcloudapp.net",
 		ResourceManagerVMDNSSuffix:   "cloudapp.windowsazure.us",
+		ContainerRegistryDNSSuffix:   "azurecr.io",
 	}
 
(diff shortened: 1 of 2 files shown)


mcardosos:
@marstr @jhendrixMSFT @colemickens Should we merge?

colemickens:
It's a bit awkward without it filled in for the other clouds, because the field will be there in the struct, but it won't be blank for the other clouds. I can't remember what's the story on the sovereign clouds though... cc: @SajayAntony

salameer:
Hi @colemickens  and @SajayAntony

colemickens:
(I'll defer to someone else on whether or not this should be merged when the field will be empty for everything but public cloud.)

salameer:
@SajayAntony any thoughts here?

djyou:
Right now, ACR is not yet available in USGov, China, or German. We haven't decided the domain suffix for them either. If not having the filed would be concern, I suppose you can put an empty string or just "azurecr.io" there.

salameer:
Thanks @djyou, we're on it

mcardosos:
Fixed
Feel free to merge @jhendrixMSFT @marstr

PR #105, opened 2017-01-11
Fixing terraform's issue #11131

nicocesar:
is "?version-api" a leftover from some older server version deployed ? it doesn't work for me in China - see https://github.com/hashicorp/terraform/issues/11131


jhendrixMSFT:
@colemickens do you know why this version is specified or who we can ask?  I only ask you because you added the file last year.

colemickens:
Looks like a mistake. The other cloud's AAD endpoint don't have that suffix.

colemickens:
Change LGTM.

jhendrixMSFT:
I figured that was the case but wanted to get a second opinion as there's still a lot I don't understand in this space.  Thanks!

PR #106, opened 2017-01-13
Adding telemetry

mcardosos:
An example user agent looks like this...
```
Go/go1.7 (amd64-windows) go-autorest/7.2.2
```

Another example. When using ARM packages in the Azure SDK for Go, the user agent would look like this...
```
Go/go1.7 (amd64-windows) go-autorest/7.2.2 Azure-SDK-for-Go/7.0.1-beta arm-resources/2016-09-01
```

Just like in the [Ruby SDK](https://github.com/Azure/azure-sdk-for-ruby/pull/543), users can also add more info with `AddToUserAgent`.
PTAL
@kirthik @salameer @jhendrixMSFT @marstr @vishrutshah

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -152,6 +152,8 @@ func NewClientWithUserAgent(ua string) Client {
 	return c
 }
 
+// getDefaultUserAgent builds a string containing the Go version, system archityecture and OS,
+// and the go-autorest version.
 func getDefaultUserAgent() string {
 	return fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
 		runtime.Version(),
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -23,13 +23,24 @@ const (
 	DefaultRetryAttempts = 3
 )
 
-var statusCodesForRetry = []int{
-	http.StatusRequestTimeout,      // 408
-	http.StatusInternalServerError, // 500
-	http.StatusBadGateway,          // 502
-	http.StatusServiceUnavailable,  // 503
-	http.StatusGatewayTimeout,      // 504
-}
+var (
+	// defaultUserAgent builds a string containing the Go version, system archityecture and OS,
+	// and the go-autorest version.
+	defaultUserAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
+		runtime.Version(),
+		runtime.GOARCH,
+		runtime.GOOS,
+		Version(),
+	)
+
+	statusCodesForRetry = []int{
+		http.StatusRequestTimeout,      // 408
+		http.StatusInternalServerError, // 500
+		http.StatusBadGateway,          // 502
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout,      // 504
+	}
+)
 
 const (
 	requestFormat = `HTTP Request Begin ===================================================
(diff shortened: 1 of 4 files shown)


autorest/client.go
```diff
@@ -140,12 +141,30 @@ type Client struct {
...
 		RetryAttempts:   DefaultRetryAttempts,
 		RetryDuration:   30 * time.Second,
-		UserAgent:       ua,
+		UserAgent:       getDefaultUserAgent(),
+	}
+	c.AddToUserAgent(ua)
+	return c
+}
+
+func getDefaultUserAgent() string {
```
vishrutshah: minor: add documentation.

autorest/version.go
```diff
@@ -6,8 +6,8 @@ import (
 
 const (
 	major        = "7"
-	minor        = "0"
-	patch        = "0"
+	minor        = "2"
+	patch        = "2"
```
vishrutshah: just curious, why incrementing minor + patch ?

jhendrixMSFT: Agreed I think we should just update minor since this is an additive feature.

jhendrixMSFT:
Looks good!

autorest/version.go
```diff
@@ -6,8 +6,8 @@ import (
 
 const (
 	major        = "7"
-	minor        = "0"
-	patch        = "0"
+	minor        = "2"
+	patch        = "3"
```
marstr: Actually, I messed up with go-autorest and forgot to adjust this for 7.2.3. This should be updated to 7.2.4

autorest/version_test.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 func TestVersion(t *testing.T) {
-	v := "7.0.0"
+	v := "7.2.3"
```
marstr: As above, this should be 7.2.4

autorest/client.go
```diff
@@ -140,12 +141,32 @@ type Client struct {
...
+		UserAgent:       getDefaultUserAgent(),
+	}
+	c.AddToUserAgent(ua)
+	return c
+}
+
+// getDefaultUserAgent builds a string containing the Go version, system archityecture and OS,
+// and the go-autorest version.
+func getDefaultUserAgent() string {
+	return fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
```
marstr: Instead of regenerating this string everytime, we should store it. How about using one of these two patterns?

```
var defaultUserAgent string = fmt.Sprintf(
    "Go/%s (%s-%s) go-autorest/%s",
    runtime.Version(),
    runtime.GOARCH,
    runtime.GOOS,
    Version()
)
func getDefaultUserAgent() string {
    return defaultUserAgent
}
```
Alternatively:
```
var defaultUserAgent string
func getDefaultUserAgent() string {
    if "" == defaultUserAgent {
        defaultUserAgent = fmt.Sprintf(
            "Go/%s (%s-%s) go-autorest/%s",
            runtime.Version(),
            runtime.GOARCH,
            runtime.GOOS,
            Version()
        )
    }
    return defaultUserAgent
}
```

autorest/client.go
```diff
@@ -140,12 +141,32 @@ type Client struct {
...
+		runtime.Version(),
+		runtime.GOARCH,
+		runtime.GOOS,
+		Version(),
+	)
+}
+
+// AddToUserAgent adds an extension to the current user agent
+func (c *Client) AddToUserAgent(extension string) {
```
marstr: Personally, I think it may be good to return an error here if `extension` is the empty string.

autorest/version.go
mcardosos: fixed

autorest/version_test.go
mcardosos: fixed

autorest/client.go
mcardosos: fixed

autorest/client.go
mcardosos: fixed

PR #107, opened 2017-01-17
Pass attempt 0 to DelayForBackoff in AfterDelay

codablock:
Passing 1 to DelayForBackoff results in the delay being doubled. If a delay
of 1 minute is used, it is unexpected that the polling delays for 2
minutes.

Encountered while debugging https://github.com/kubernetes/kubernetes/issues/35180#issuecomment-273085063

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -73,7 +73,7 @@ func SendWithSender(s Sender, r *http.Request, decorators ...SendDecorator) (*ht
 func AfterDelay(d time.Duration) SendDecorator {
 	return func(s Sender) Sender {
 		return SenderFunc(func(r *http.Request) (*http.Response, error) {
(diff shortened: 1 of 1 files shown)


colemickens:
cc: @garimakhulbe @mcardosos 

@codablock and others are reporting that the VM Client is not polling properly. From the linked issue, there's question as to whether or not CRP is reporting the header, whether go-autorest is respecting it, and then what looks like a bug with `1` being passed here instead of `0`.

Could you take a look? Thanks!

marstr:
Thanks for digging this up, @codablock & @colemickens. The change seemed pretty obvious to me, so I went ahead and merged it in. I'll sync-up with the rest of the Azure-SDK-for-Go team, and publish a patch if they sign-off.

Issue #110, opened 2017-01-19
v7.2.3 version.go still reports v7.2.2


marstr:
The commit tagged v7.2.3 will still report as v7.2.2 using  the `autorest.Version()` function.

marstr:
CI testing should be updated to prevent this happening again.

marstr:
Because we'd like to treat tags as immutable, there isn't really anything we can do to fix this for v7.2.3. However, we can test to make sure this doesn't happen again in the future. The PR above (#111) takes care of that. Once that PR is merged, I'll close this issue.

mcardosos:
BTW, I can tag v7.2.4 once we merge the PR

PR #111, opened 2017-01-19
Updating Version() test to assert it is updated

marstr:
For the v7.2.3 release, the autorest.Version() function didn't get
updated to reflect the new release. To ensure that doesn't happen again,
the TestVersion() function was updated to make sure that when commited,
the version is greater than the most recently relesed version.

This added a tight coupling with Git and using tags as release
mechanisms.

### DIFF
--- a/autorest/version.go
+++ b/autorest/version.go
@@ -12,7 +12,12 @@ const (
 	semVerFormat = "%s.%s.%s%s"
 )
 
+var version string
+
 // Version returns the semantic version (see http://semver.org).
 func Version() string {
-	return fmt.Sprintf(semVerFormat, major, minor, patch, tag)
+	if "" == version {
+		version = fmt.Sprintf(semVerFormat, major, minor, patch, tag)
+	}
+	return version
 }
--- a/autorest/version.go
+++ b/autorest/version.go
@@ -16,7 +16,7 @@ var version string
 
 // Version returns the semantic version (see http://semver.org).
 func Version() string {
-	if "" == version {
+	if version == "" {
 		version = fmt.Sprintf(semVerFormat, major, minor, patch, tag)
 	}
 	return version
--- a/autorest/version_test.go
+++ b/autorest/version_test.go
@@ -1,13 +1,79 @@
 package autorest
 
 import (
+	"io"
+	"os/exec"
 	"testing"
+
+	"fmt"
+
+	"sync"
+
+	"github.com/Masterminds/semver"
 )
 
 func TestVersion(t *testing.T) {
-	v := "7.2.4"
-	if Version() != v {
-		t.Fatalf("autorest: Version failed to return the expected version -- expected %s, received %s",
-			v, Version())
+	var declaredVersion *semver.Version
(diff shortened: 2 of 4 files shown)


marstr:
This pull request is in response to issue #110.

autorest/version.go
```diff
@@ -12,7 +12,12 @@ const (
 	semVerFormat = "%s.%s.%s%s"
 )
 
+var version string
+
 // Version returns the semantic version (see http://semver.org).
 func Version() string {
-	return fmt.Sprintf(semVerFormat, major, minor, patch, tag)
+	if "" == version {
```
jhendrixMSFT: I don't think this style is necessary in Go as you can't have an assignment here.

marstr: Ah right you are. My old K&R C habits are shining through.

autorest/version_test.go
```diff
@@ -1,13 +1,90 @@
...
+
+			tagLister := exec.Command("git", "tag")
+			tagLister.Stdout = writer
+
+			if err := tagLister.Run(); err != nil {
+				tagFetchErr = err
+			}
+		}()
+
+		go func() {
```
jhendrixMSFT: Is there a reason this logic is split into its own go routine?  It must wait until the other go routine completes yes?

marstr: Ah actually, because they're communicating via io.Pipe, they must execute in parallel.

jhendrixMSFT: Fair enough, but does this just add more complexity?  It seems to me that these are sequential operations that can operate in on go routine.

marstr: That's true, and it's not a bad point. The reason I did it this way was so that I didn't have to worry about choosing a buffer allocation size. If you don't think this is readable, I'll re-write it to execute sequentially.

autorest/version_test.go
```diff
@@ -1,13 +1,106 @@
...
+	"github.com/Masterminds/semver"
 )
 
 func TestVersion(t *testing.T) {
-	v := "7.2.4"
-	if Version() != v {
-		t.Fatalf("autorest: Version failed to return the expected version -- expected %s, received %s",
-			v, Version())
+	var declaredVersion *semver.Version
+	if temp, err := semver.NewVersion(Version()); nil == err {
```
mcardosos: >nil == err [](start = 47, length = 10)

It is just style but for me it looks better like this...
```go
declaredVersion, err := semver.NewVersion(Version())
if err != nil {
  t.Error(err)
  t.FailNow()
}
t.Logf("Declared Version: %s", declaredVersion.String())
```

autorest/version_test.go
```diff
@@ -1,13 +1,106 @@
 package autorest
 
 import (
+	"io"
+	"os/exec"
 	"testing"
+
+	"fmt"
```
mcardosos: Nit, new lines between imports

autorest/version_test.go
marstr: Updated.

autorest/version_test.go
marstr: Updated.

Issue #109, opened 2017-01-18
JSON unmarshaling does not cope with byte-order mark


rogpeppe:
I sometimes see an error like this:

    error response cannot be parsed: "\ufeff{\"odata.error\":{\"code\":\"Request_MultipleObjectsWithSameKeyValue\",\"message\":{\"lang\":\"en\",\"value\":\"Another object with the same value for property servicePrincipalNames already exists.\"},\"requestId\":\"46a97035-e286-4d41-8eee-ae33de2f0288\",\"date\":\"2017-01-18T12:31:21\"}}" error: invalid character 'ï' looking for beginning of value

It looks like the error response from Azure sometimes contains a byte-order mark that Go's JSON decoder does not like. It should probably be trimmed before parsing.

salameer:
@mcardosos:

- Can you investigate with the service team on the reasoning for the adding this byte-order. you can work with @kirthik to nail down the service team contacts.
- Can you also check if this this is the only byte order value that they would be sending before omitting it.

salameer:
@rogpeppe 

Thanks for contacting us can you please provide a sample API that's showing this behavior.

"Sample of your code would be even better :)"

garimakhulbe:
Error is coming as `odata.error` which might be failing in json unmarshal.
```Go
ServiceError *ServiceError `json:"error"`
```

axw:
FWIW, the code in question is in:
 - https://github.com/juju/juju/blob/staging/provider/azure/internal/ad/client.go
 - https://github.com/juju/juju/blob/staging/provider/azure/internal/ad/serviceprincipals.go
 - https://github.com/juju/juju/blob/staging/provider/azure/internal/azureauth/interactive.go#L208

PR #113, opened 2017-02-01
Remove UTF-8 BOM from response body

mcardosos:
Fixes [this issue](https://github.com/Azure/go-autorest/issues/109).
PTAL @marstr @jhendrixMSFT


jhendrixMSFT:
Be sure to follow up with the applicable RP so they don't send up us the BOM in the future (sorry couldn't resist).  Change looks good.

mcardosos:
@kirthik :D

PR #116, opened 2017-02-09
Added WithCustomBaseURL func

mcardosos:
This is for `x-ms-parametrized-host` support.
PTAL @jhendrixMSFT @marstr

### DIFF
--- a/autorest/preparer.go
+++ b/autorest/preparer.go
@@ -183,6 +183,16 @@ func WithBaseURL(baseURL string) PrepareDecorator {
 	}
 }
 
+// WithCustomBaseURL returns a PrepareDecorator that replaces brace-enclosed keys within the
+// request base URL (i.e., http.Request.URL) with the corresponding values from the passed map.
+func WithCustomBaseURL(baseURL string, urlParameters map[string]interface{}) PrepareDecorator {
+	parameters := ensureValueStrings(urlParameters)
+	for key, value := range parameters {
+		baseURL = strings.Replace(baseURL, "{"+key+"}", value, -1)
+	}
+	return WithBaseURL(baseURL)
+}
+
 // WithFormData returns a PrepareDecoratore that "URL encodes" (e.g., bar=baz&foo=quux) into the
 // http.Request body.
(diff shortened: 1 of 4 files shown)


mcardosos:
Oops, travis is not passing because of the version test :I

mcardosos:
Fixed @marstr

autorest/preparer.go
```diff
@@ -183,6 +183,16 @@ func WithBaseURL(baseURL string) PrepareDecorator {
 	}
 }
 
+// WithCustomBaseURL returns a PrepareDecorator that replaces brace-enclosed keys within the
+// request base URL (i.e., http.Request.URL) with the corresponding values from the passed map.
+func WithCustomBaseURL(baseURL string, urlParameters map[string]interface{}) PrepareDecorator {
```
jhendrixMSFT: Why does the map's value type need to be interface{}?

mcardosos: Truth is it does not need to be interface. I'm following what `WithPathParameters` and `WithQueryParameters` do right now. Still will take a look if those can use `map[string]string`.

mcardosos: It is possible to gen `map[string]string` and modify go-autorest to take them instead of `map[string]interface{}`. The same goes for `WithEscapedPathParameters()`, `WithPathParameters()` and `WithQueryParameters()`. What do you guys think about changing those too? I could update this PR and add a new one for AutoRest tomorrow.
cc @marstr

marstr: Hmmm, I would support the above proposal to move them all to `map[string]string`. Even better would be `map[string]fmt.Stringer` [Stringer Type documentation](https://golang.org/pkg/fmt/#Stringer)

marstr: Nevermind, I retract the map[string]fmt.Stringer suggestion, I just played with it in [go playground](https://play.golang.org) and don't like how that would impact declaring map literals.

jhendrixMSFT: After digging into this more I understand the reasoning here, no need to change it.

Issue #117, opened 2017-02-14
ByDiscardingBody() should be added as a decorator to allow HTTP connection reuse


marstr:
This is discussed more thoroughly in Azure/azure-sdk-for-go#544 and Azure/azure-sdk-for-go#404.  However, the crux of the issue is that in order to take advantage of Go's HTTP Library and connection re-use, the Response Body must be read to completion before the connection is closed.

marstr:
Closing issue.

PR #118, opened 2017-02-14
Adding ByDiscardingBody() decorator.

marstr:
This issue is detailed in #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ and Responding. A typical pattern is:
     DoRetryForAttempts(5, time.Second))
 
   err = Respond(resp,
+		ByDiscardingBody(),
     ByClosing())
 ```
 
--- a/autorest/responder_test.go
+++ b/autorest/responder_test.go
@@ -4,12 +4,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"reflect"
 	"strings"
 	"testing"
-
-	"io/ioutil"
 
 	"github.com/Azure/go-autorest/autorest/mocks"
 )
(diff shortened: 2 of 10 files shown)


mcardosos:
Question, would this func be included in all Responder funcs in the SDK, or just in the ones that do not use the response body?

marstr:
Just ones that don't use the body. Though, in theory it shouldn't hurt to add it everywhere.

autorest/version.go
```diff
@@ -7,7 +7,7 @@ import (
 const (
 	major        = "7"
 	minor        = "2"
```
mcardosos: I think it is not a patch, but a minor version.

marstr: Ah, good point. :) I'll fix that up.

README.md
```diff
@@ -21,6 +21,7 @@ and Responding. A typical pattern is:
     DoRetryForAttempts(5, time.Second))
 
   err = Respond(resp,
+		ByDiscardingBody(),
```
jhendrixMSFT: Nit, formatting.

autorest/responder_test.go
```diff
@@ -198,7 +202,8 @@ func TestByCopying_AcceptsNilBody(t *testing.T) {
 
 func TestByClosing(t *testing.T) {
 	r := mocks.NewResponse()
-	err := Respond(r, ByClosing())
+	err := Respond(r,
+		ByClosing())
```
jhendrixMSFT: Was this supposed to be ByDiscardingBody or was the line break intentional?

README.md
jhendrixMSFT: I see this is in a code block, probably not a big deal.

marstr: +1

yeah, code in markdown is funky, especially in PRs.

autorest/responder_test.go
marstr: I had added `ByDiscardingBody()` while on auto-pilot applying my fix, but thought better of it because this test should isolate `ByClosing()`. I'll fix up this formatting so that it doesn't jump out to folks looking at diffs later.

PR #122, opened 2017-02-23
Extract the authentication code in a separate package 'adal'

cosmincojocar:
In addition, I refactored the code to remove the dependency between adal and autorest packages.  

@colemickens


autorest/adal/sender.go
```diff
@@ -0,0 +1,46 @@
+package adal
```
colemickens: What is the purpose of this? Does ADAL need it's own Sender impl? Can it not use the standard one used elsewhere?

autorest/adal/token.go
```diff
@@ -12,8 +12,10 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/Azure/go-autorest/autorest"
+	"encoding/json"
 	"github.com/dgrijalva/jwt-go"
```
colemickens: nit/style: Sort github imports separately from std imports (add space b/w them)

colemickens:
I'm quite glad to see the OAuth stuff out of Environment. When I was hacking on MSI, it was clear that I didn't quite model the oauthconfig stuff quite right. This looks more like what I've seen in other OAuth libraries... and I can tell it would make my MSI change less hacky.

autorest/azure/example/main.go
```diff
@@ -244,7 +244,7 @@ func main() {
 	}
 
 	client := &autorest.Client{}
-	client.Authorizer = spt
+	client.Authorizer = autorest.NewBearerAuthorizer(spt)
```
colemickens: Is it not possible to make the `spt` type implement the authorizer interface so that it's slightly more backward compatible? Or is that less clean?

autorest/adal/sender.go
cosmincojocar: The reason for duplicating this is to avoid circular dependencies between autorest and adal packages. The autorest should import adal for authentication.

The Sender could be moved in a small package which will be use by both packages.

autorest/azure/example/main.go
cosmincojocar: The Authorizer interface is dependent on autorest.PrepareDecorator.  

```Go
type Authorizer interface {
	WithAuthorization() PrepareDecorator
}
```
This will probably require more refactoring that the Token implements the interface without creating a dependency to autorest package. 

This implementation appears to me a bit more clean, because the authorizer is only responsible to build the HTTP header.

salameer:
Hi @cosmincojocar,
Thanks for the awesome contribution, we will be reviewing it very shortly. I have a couple of minor asks :)
- Can you kindly target the Dev branch :) as there are commits waiting to be merged into master and we wanna make sure your stuff meshes well
- Would you mind providing a brief description of the problem that your facing and trying to solve with this PR, we'd love to have the context from your perspective here

THanks again !!

cosmincojocar:
Thanks @salameer for feedback.

- The PR targets now the dev branch. As you may notice, I increased the version to 8.0.0 because there are some API breaking changes. 
-  I have a concert use case where I want to access  a RESTful service (OAuth Resource Server) with a tool written in Go (OAuth Client). This services relies on AAD when verifying the incoming token. The tool only needs is to acquire a token form AAD, but does not require any dependency to Azure sdk.

Overall is better to decouple this logic from autorest implementation. This will allow a clean dependency graph: 

               adal   azure   date     to  validation
                 \          \           |        /         /              
                                  autorest

salameer:
Hello  Folks,

to be very transparent on this one, there's alot of buzz going around with Authentication libs coming to light (ASAL & ADAL for ex.) and we'd like to limit amount of breaking changes to our customers.

We'll be following up with the new initiative and figuring out how far down the line will they be lighting up

Thank you So MUCH @cosmincojocar

cosmincojocar:
Hi @salameer, 

I moved the adal related code in a stand-alone repository https://github.com/cosmincojocar/adal which contains some more fixes. In addition, the documentation is updated. The configuration steps use now Azure CLI 2.0. Also I converted the example in a bit more generic command line tool which can be used to acquire a token for a given resource. 

cc @colemickens

autorest/adal/token_test.go
```diff
@@ -281,7 +334,7 @@ func TestServicePrincipalTokenRefreshReturnsErrorIfNotOk(t *testing.T) {
 
 	err := spt.Refresh()
 	if err == nil {
-		t.Fatal("azure: Failed to return an when receiving a status code other than HTTP 200")
+		t.Fatal("adal: Failed to return an when receiving a status code other than HTTP 200")
```
marstr: [nit] this could also be rewritten to not have the hardcoded 200

autorest/adal/token_test.go
```diff
@@ -320,21 +373,27 @@ func TestServicePrincipalTokenEnsureFreshRefreshes(t *testing.T) {
 	spt := newServicePrincipalToken()
 	expireToken(&spt.Token)
 
+	body := mocks.NewBody(newTokenJSON("test", "test"))
+	resp := mocks.NewResponseWithBodyAndStatus(body, 200, "OK")
```
marstr: Another `http.StatusOK` here

autorest/authorization_test.go
```diff
@@ -0,0 +1,137 @@
...
+	jwt := `{
+		"access_token" : "accessToken",
+		"expires_in"   : "3600",
+		"expires_on"   : "test",
+		"not_before"   : "test",
+		"resource"     : "test",
+		"token_type"   : "Bearer"
+	}`
+	body := mocks.NewBody(jwt)
+	resp := mocks.NewResponseWithBodyAndStatus(body, 200, "OK")
```
marstr: Another `http.StatusOK` here

marstr:
> I moved the adal related code in a stand-alone repository https://github.com/cosmincojocar/adal which contains some more fixes. In addition, the documentation is updated. The configuration steps use now Azure CLI 2.0. Also I converted the example in a bit more generic command line tool which can be used to acquire a token for a given resource.

This PR takes the Go ADAL library a long way towards independence in and of itself. However, I agree with taking that next step to it populating its own repository. We're trying to sort out some organizational questions on that front in terms of who would own it etc, but we'll make the right thing happen. When that happens, I would prefer to preserve the history of these files by using a filter-branch instead of just copying files into a new repo.

In the mean time, we'll accept this PR into the dev branch once the comments above have been addressed, and I would urge you to contribute here instead of in your own repository. Obviously though, this is Open Source (which is awesome) and you're free to redistribute it via your fork.

colemickens:
(Are you both in contact with the Red Team working on ASAL (overlaps significantly with ADAL for Azure contexts as I understand it)? I think I put y'all in touch, but it's extra information that's probably relevant to any organizational/ownership questions...)

cosmincojocar:
@marstr I addressed all your comments. In addition I added a README file for adal package and a generic command line tool.

colemickens:
@cosmincojocar Is there an example of how to use `adal` with `azure-sdk-for-go`? Because when I try to use the ServicePrincipalToken as an Authorizer, I get this now: 

```
example-msi-app/main.go:36: cannot use spt (type *adal.ServicePrincipalToken) as type autorest.Authorizer in assignment:
        *adal.ServicePrincipalToken does not implement autorest.Authorizer (missing WithAuthorization method)
```

cosmincojocar:
@colemickens I updated the original example autorest/azure/example/main.go to use the stand alone adal package. I suspect your application depends on master branch of go-autorest. The changes were merged only in the dev branch. Please try this branch and let me know if you still have issues.

PR #123, opened 2017-02-24
Cleaning up redundant consts.

marstr:
When available, there is no reason not to re-use constant values from stdlib.


mcardosos:
Lil comment (not about the review, it LGTM): I wonder why http package does not include the MERGE const.

PR #125, opened 2017-03-16
Add the GAS security scanner to the build steps

cosmincojocar:
GAS tool inspects source code for security problems, see more details at https://github.com/GoASTScanner/gas.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,13 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
   - go get -u github.com/stretchr/testify
+  - go get -u github.com/GoASTScanner/gas
   - glide install
 
 script:
   - test -z "$(gofmt -s -l -w ./autorest/. | tee /dev/stderr)"
   - test -z "$(golint ./autorest/... |  tee /dev/stderr)"
   - go vet ./autorest/...
+  - gas ./autorest/...
   - go build -v ./autorest/...
   - go test -v ./autorest/...
--- a/autorest/version.go
+++ b/autorest/version.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"strings"
+	"sync"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -10,11 +10,13 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
   - go get -u github.com/stretchr/testify
+  - go get -u github.com/GoASTScanner/gas
```
marstr: To be consistent with our other repositories, we should use the HewlettPackard fork of this repository which doesn't introduce breaking changes as often as the GoASTScanner main repository: github.com/HewlettPackard/gas

See here for what I'm talking about:
https://github.com/Azure/azure-sdk-for-go/blob/master/rungas.sh#L8

autorest/version.go
```diff
@@ -18,11 +18,14 @@ var version string
 // Version returns the semantic version (see http://semver.org).
 func Version() string {
```
marstr: I like the updates that you've made here, much more readable.

As long as we're at it, we should get rid of the race-condition that is possible here by updating this function to use [sync/once](https://golang.org/pkg/sync/#Once) instead of the current string check.

Feel free to ignore this comment if you'd like. I can make these changes at a later date.

.travis.yml
cosmincojocar: @marstr It seems that the github.com/HewlettPackard/gas repository was deleted from github. Probably you should update the azure-sdk-for-go to use GoASTScanner.

autorest/version.go
cosmincojocar: @marstr I modified the code as you suggested. The version is set only once if is empty.

.travis.yml
marstr: Thanks for the heads up.

autorest/version.go
marstr: Thanks @cosmincojocar! However, there are still a couple of tweaks needed here:
- The declaration, `var once sync.Once` should be made outside of the `Version` function.
- `once.Do(func(){...}` is an alternative to `if version == "" {...}`. It is safe to remove that check.

cosmincojocar:
@marstr I removed the version check.

PR #126, opened 2017-03-23
adal: add support for MSI VM Extension

colemickens:
This stacks on top of #122, please just look at the last commit.

I can't decide if it should be `NewServicePrincipalTokenFromMSI` or if it should be `NewServicePrincipalTokenFromMSIExtension`...

If it's the former, then in the future, it has to try to detect whether to use extension endpoint+api or the future endpoint+api.

If it's the latter, then in the future, application code has to be updated, rather than just updating the SDK to take advantage of MSI in the new endpoint.

### DIFF
--- a/autorest/adal/example-msi-app/main.go
+++ b/autorest/adal/example-msi-app/main.go
@@ -13,9 +13,6 @@ import (
 var cloud = azure.PublicCloud
 
 func main() {
-	//var spt *azure.ServicePrincipalToken
-	//var err error
-
 	tenantID := os.Getenv("TENANT_ID")
 	subscriptionID := os.Getenv("SUBSCRIPTION_ID")
 	resourceGroupName := os.Getenv("RESOURCE_GROUP")
--- a/autorest/adal/example-msi-app/main.go
+++ b/autorest/adal/example-msi-app/main.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
+
+var cloud = azure.PublicCloud
+
+func main() {
+	//var spt *azure.ServicePrincipalToken
+	//var err error
+
+	tenantID := os.Getenv("TENANT_ID")
+	subscriptionID := os.Getenv("SUBSCRIPTION_ID")
+	resourceGroupName := os.Getenv("RESOURCE_GROUP")
+
+	oauthConfig, err := adal.NewOAuthConfig(cloud.ActiveDirectoryEndpoint, tenantID)
+	if err != nil {
+		panic(err)
+	}
+
+	spt, err := adal.NewServicePrincipalTokenFromMSI(
+		*oauthConfig,
+		cloud.ResourceManagerEndpoint)
+	if err != nil {
+		panic(err)
+	}
+
+	client := resources.NewGroupsClientWithBaseURI(cloud.ResourceManagerEndpoint, subscriptionID)
+	client.Authorizer = autorest.NewBearerAuthorizer(spt)
+
+	resources, err := client.ListResources(resourceGroupName, "", "", nil)
+	if err != nil {
+		panic(err)
+	}
+
+	result := "resources: "
+	for _, r := range *(resources.Value) {
+		result += *r.Name + ", "
+	}
+
+	log.Println(result)
+
+	// update RG, add a tag
+	rg, err := client.Get(resourceGroupName)
+	if err != nil {
+		panic(err)
+	}
+
+	value := "success"
+	var tags map[string]*string
+	if rg.Tags != nil {
+		tags = *rg.Tags
+	} else {
+		tags = make(map[string]*string)
+	}
+	tags["msie2e"] = &value
+
+	rg.Tags = &tags
+
+	// this is a silly necessity
+	rg.Properties.ProvisioningState = nil
+	rg.ID = nil
+
+	_, err = client.CreateOrUpdate(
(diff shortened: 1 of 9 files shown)


marstr:
Howdy @colemickens,

Now that #122 has been merged, mind rebasing/cherry-picking your commit on top of the dev branch?

colemickens:
Working on it right now, got one question to @cosmincojocar before I e2e test and push new commits.

autorest/adal/token.go
```diff
@@ -247,6 +259,47 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
...
+		AuthorityEndpoint: msiTokenEndpointURL,
+	}
+
+	spt := &ServicePrincipalToken{
+		oauthConfig:   oauthConfig,
+		secret:        msiSecret,
+		resource:      resource,
+		autoRefresh:   true,
+		refreshWithin: defaultRefresh,
+		sender:        &http.Client{},
```
cosmincojocar: Here you need to assign the **callbacks** to refreshCallbacks.

autorest/adal/token.go
```diff
@@ -247,6 +259,47 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
 	)
 }
 
+func NewServicePrincipalTokenFromMSI(oauthConfig OAuthConfig, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
```
cosmincojocar: Please could you add a test for this function.

autorest/adal/example-msi-app/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
+
+var cloud = azure.PublicCloud
+
+func main() {
+	//var spt *azure.ServicePrincipalToken
+	//var err error
```
cosmincojocar: I guess you can remove this commented code.

autorest/adal/token.go
```diff
@@ -247,6 +259,47 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
...
+	// Read MSI settings
+	bytes, err := ioutil.ReadFile("/var/lib/waagent/ManagedIdentity-Settings")
```
cosmincojocar: I would provide this file as a *Reader* argument to this function. The name or the path may change in the future and this will provide more flexibility.

colemickens: I have strong opinions against doing this. This location can absolutely *never* change or it will irreparably break *any application that uses this SDK w/ MSI*. The only way MSI works as an e2e feature is for the location to be completely permanent (or to have fallback locations in the future if it is moved).

Otherwise, what, now any downstream application configuration also has to include knowledge of where some latest version of the MSI extension is dropping its settings file in order to pass it into here? That just doesn't seem right.

colemickens:
Thanks very much for the review @cosmincojocar. I'm doing some sanity testing and then will incorporate your comments and push new commits.

colemickens:
The latest commits actually work now, but haven't responded to other feedback. Will do that ASAP - esp since it would've caught a rebase mistake I made and caught with manual verification.

colemickens:
New commit pushed:

- tests written
- doc'd a public function
- removed dead code in msi example
- responded to feedback (except making the path a parameter, as I don't agree we should even think in such a way)

This has been sanity checked using: https://github.com/Azure/managed-identity-extension/pull/6 (You may not have access - it's an ARM template that uses CSE to run an example app built with this code. It updates a tag on the resource group. The wrapper script then dumps out the RG and you can see the tag was added)

colemickens:
@marstr Can you advise? I'm not really sure I understand the gas errors in the Travis builds. Things seem to be busted that are not seemingly related to my changes, but I'm unsure?

colemickens:
Example:

```
[gas] 2017/03/28 10:35:44 Error during type checking: "/home/travis/gopath/src/github.com/Azure/go-autorest/autorest/adal/cmd/adal.go:15:2: could not import github.com/Azure/go-autorest/autorest/adal (can't find import: github.com/Azure/go-autorest/autorest/adal)"
```

I didn't touch `.../adal/cmd/adal.go` for example.

marstr:
It looks like the example you gave:

> [gas] 2017/03/28 10:35:44 Error during type checking: "/home/travis/gopath/src/github.com/Azure/go-autorest/autorest/adal/cmd/adal.go:15:2: could not import github.com/Azure/go-autorest/autorest/adal (can't find import: github.com/Azure/go-autorest/autorest/adal)"

Is firing when GAS is run, but not actually causing a Travis Failure. What probably happened (but I haven't confirmed yet) is that these issues were fired in the original PR but not acted upon. I'll spin up a new issue to address that.

As for the other failures that are blocking: 
1. Most likely a `Fatal` call that should be a `Fatalf`:
>  go vet ./autorest/...
autorest/adal/token_test.go:506: possible formatting directive in Fatal call
exit status 1

2. Looks like the example exe that you wrote up doesn't download the Azure-SDK-for-Go
> go test -v ./autorest/...
autorest/adal/example-msi-app/main.go:7:2: cannot find package "github.com/Azure/azure-sdk-for-go/arm/resources/resources" in any of:
	/home/travis/gopath/src/github.com/Azure/go-autorest/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources (vendor tree)
	/home/travis/.gimme/versions/go1.7.linux.amd64/src/github.com/Azure/azure-sdk-for-go/arm/resources/resources (from $GOROOT)
	/home/travis/gopath/src/github.com/Azure/azure-sdk-for-go/arm/resources/resources (from $GOPATH)

This can be fixed either by adding `go get -u github.com/Azure/azure-sdk-for-go` to `.travis.yml` or using glide in your example package.

Issue #129, opened 2017-04-03
Fedora 25 - go test: FAIL: TestVersion


marcindulak:
```
cat /etc/redhat-release  
#Fedora release 25 (Twenty Five)
dnf -y install "compiler(go-compiler)"

go version
#go version go1.7.5 linux/amd64

dnf -y install "golang(github.com/Masterminds/semver)"
dnf -y install "golang(github.com/dgrijalva/jwt-go)"
dnf -y install "golang(github.com/stretchr/testify/require)"

mkdir -p /tmp/go/src
export GOPATH=/tmp/go:/usr/share/gocode
cd /tmp/go/src
mkdir -p github.com/Azure
cd github.com/Azure
git clone http://github.com/Azure/go-autorest
cd go-autorest/autorest
git rev-parse HEAD
# ec5f4903f77ed9927ac95b19ab8e44ada64c1356
go test 
#--- FAIL: TestVersion (0.00s)
#       version_test.go:22: Declared Version: 7.3.0
#       version_test.go:31: Current Release Version: 7.3.0
#       version_test.go:34: autorest: Assertion that the Declared version is greater than Current Release Version failed 7.3.0 7.3.0
#FAIL
#exit status 1
```

marstr:
Ahh, yes this is an expected error. We do that to ensure that we don't accidentally release without having bumped the version number in version.go. However, as you've noted it means that this error occurs after we have created the release.

@marcindulak, does our strategy block your scenario? If so, I'd be happy to find a new way of going after this.

marstr:
Tell you what, after chatting with @jhendrixMSFT about this offline, we've both agreed that it's a reasonable assumption that our tests have no expected failures when running `go test` on master.

PR #130, opened 2017-04-04
Removing Version() check in tests.

marstr:
As noted in #129, it is a reasonable expectation from our customers that `go test` should
execute without failure on master from their machines. At the moment,
there is a by design failure because the version matches the most
recently published tag.

Eventually, we should enforce this again, but not until we can do so
only for our own release pipeline. To ensure folks don't get snared on
this, we'll delete the offending test here.


mcardosos:
Question: is there a way to skip the tests unless you are on CI ?

marstr:
@mcardosos, not at the moment, but I plan on adding that back in a little later to protect ourselves. I may do it as an independent script to be run during CI instead of part of the SDK tests themselves.

PR #134, opened 2017-04-17
Add an opt-in method to always preserve the response body.

jhendrixMSFT:
For non-long running operations the response body is closed by default.
This change adds an API to opt into always preserving the response body.

### DIFF
--- a/autorest/responder.go
+++ b/autorest/responder.go
@@ -11,6 +11,13 @@ import (
 	"strings"
 )
 
+var preserveRespBody = false
+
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@marstr I'm at a loss as to why Travis keeps failing, when you have some time could you please take a peek?

marstr:
Investingating now, @jhendrixMSFT!

jhendrixMSFT:
After fixing https://github.com/Azure/autorest/issues/2079 this change is no longer necessary so closing this PR.

PR #135, opened 2017-04-17
Finished Text, Binary, and JSON for UnixTime

marstr:
In support of Azure/azure-sdk-for-go#587, I've implemented UnixTime for marshaling and unmarshaling Swaggers that include Unix Timestamps. Next will be updating the generator to recognize this as a well-known type.

### DIFF
--- a/autorest/date/unixtime_test.go
+++ b/autorest/date/unixtime_test.go
@@ -1,0 +1,247 @@
+package date_test
+
+import (
+	"bytes"
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/marstr/go-autorest/autorest/date"
+)
+
+func ExampleUnixTime_MarshalJSON() {
+	epoch := date.UnixTime(date.UnixEpoch())
+	text, _ := json.Marshal(epoch)
+	fmt.Print(string(text))
+	// Output: 0
+}
+
+func ExampleUnixTime_UnmarshalSON() {
+	var myTime date.UnixTime
+	json.Unmarshal([]byte("1.3e2"), &myTime)
+	fmt.Printf("%v", time.Time(myTime))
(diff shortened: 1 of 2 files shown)


mcardosos:
Related https://github.com/Azure/autorest/issues/1130
There are UNIX time tests in autorest too. I think we haven't added those.

autorest/date/unixtime_test.go
```diff
@@ -0,0 +1,247 @@
...
+	testCases := []time.Time{
+		date.UnixEpoch().Add(-1 * time.Second),                   // One second befote the Unix Epoch
+		time.Date(2017, time.April, 14, 20, 27, 47, 0, time.UTC), // The time this test was written
+	}
+
+	for _, tc := range testCases {
+		t.Run("", func(subT *testing.T) {
+			target := date.UnixTime(tc)
+
+			expected := string([]byte(fmt.Sprintf("%d", tc.Unix())))
```
mcardosos: Nit, for clarity, I would separate a bit more this 8 lines

autorest/date/unixtime_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.String(), func(subT *testing.T) {
+			subject := date.UnixTime(tc)
+			var marshaled []byte
+			if temp, err := json.Marshal(subject); err == nil {
+				marshaled = temp
+			} else {
+				t.Error(err)
+				return
```
mcardosos: Why?

autorest/date/unixtime_test.go
marstr: Derp, it looks like this actually should have both casts removed. Oops!

autorest/date/unixtime_test.go
marstr: Ah, it looks like this actually is a different bug on my part. This should be targeting subT not t.

Issue #133, opened 2017-04-07
Add support for AAD v2 in adal package 


ccojocar:
I am creating this issue to collect your thoughts about AAD v2. The current implementation supports AAD v1, but would be nice to have some support for v2 which implements the OIDC specs.  

https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-compare

I am not sure if there are any particularities in AAD v2, or one can use an exiting Go package such as `https://github.com/coreos/go-oidc` to acquire an id_token. 

cc  @colemickens @marstr

colemickens:
I've found AADv2's OIDC support to be very straight forward -- for example, the OIDC plugin in kubectl was able to refresh id_tokens, etc with AADv2 out-of-the-box.

A (better) open source Golang OAuth2 client *should* be able to talk to AADv2 for OAuth. Since Azure stuff can't use AADv2,  Go-Autorest and Azure SDK use our custom OAuth client. Hopefully stuff that is targeting AADv2 can use the popular community packages for OAuth/OIDC.

Not to say I'm opposed to having AADv2 support in here too, it just might not be worth it.

cosmincojocar:
Thank you @colemickens, it's fair enough. I think the AADv2 implements the OAuth 2.0 on top of OIDC which I am not sure if is mainstream. 

https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols-oauth-client-creds

The OIDC client will request access to a `resource` by passing  the URL in a `scope` parameter among other things like permissions.

colemickens:
Sure, OIDC flow is like OAuth2 but more standardized (endpoint discovery, etc). And requesting an `id_token` happens by requesting a specific scope... but any other resources included in the scope become part of the access token, not the `id_token`. In that sense, it's just a normal OAuth2 flow with an OIDC-capable/compliant endpoint.

If you look through the OIDC spec, it looks like OIDC == OAuth2 spec + `id_token` + endpoint discovery. Most of the major OAuth2 providers also implement OIDC (at the very least discovery, but many of them full fledged `id_token`s, etc).

jhendrixMSFT:
We are moving to AAD v2 as part of our track 2 SDK work.  It will be handled via our [azidentity](https://github.com/Azure/azure-sdk-for-go/tree/track2/sdk/azidentity) module.

We won't be doing anything for track 1 in go-autorest so I'm closing this.

colemickens:
[...]

PR #136, opened 2017-04-26
parse id_token field from adal token

shhsu:
We (Azure Container Registry) are working on a feature that requires knowing client's tenant ID given an adal token. The way to do it was to unpack the id_token field from adal.Token and read the tenant id field from there.

Currently adal.Token object seem to ignore that field


shhsu:
Just realized there's a discussion https://github.com/Azure/go-autorest/issues/133
So `id_token` is v2 specific?

shhsu:
I was able to get the tenant ID from access token, closing

PR #138, opened 2017-04-27
Remove the adal example.

jhendrixMSFT:
It breaks travis and doesn't really belong here anyways.  Will move it to
the examples repo.


jhendrixMSFT:
@marstr @mcardosos PTAL I would like to get this merged so travis is no longer broken.

mcardosos:
qq, will we have an auth sample?

PR #143, opened 2017-06-08
Add type RetriableRequest to wrap retrying an HTTP request.

jhendrixMSFT:
The current retry logic always makes a copy of the request body by calling
ioutil.ReadAll() which is inefficient in some cases.  This change moves
the handling of the request body into type RetriableRequest and implements
a few different stratagies for preserving the request body.

### DIFF
--- a/autorest/retriablerequest.go
+++ b/autorest/retriablerequest.go
@@ -1,0 +1,68 @@
+package autorest
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+// RetriableRequest provides facilities for retrying an HTTP request.
+type RetriableRequest struct {
+	req *http.Request
+	rc  io.ReadCloser
+	br  *bytes.Reader
+}
+
+// NewRetriableRequest returns a wrapper around an HTTP request that support retry logic.
+func NewRetriableRequest(req *http.Request) *RetriableRequest {
+	return &RetriableRequest{req: req}
+}
+
+// Request returns the wrapped HTTP request.
+func (rr *RetriableRequest) Request() *http.Request {
+	return rr.req
+}
+
+// Prepare signals that the request is about to be sent.
+func (rr *RetriableRequest) Prepare() (err error) {
+	// preserve the request body; this is to support retry logic as
+	// the underlying transport will always close the reqeust body
+	if rr.req.Body != nil {
+		if rr.req.GetBody != nil {
+			// this will allow us to preserve the body without having to
+			// make a copy.  note we need to do this on each iteration
+			rr.rc, err = rr.req.GetBody()
+			if err != nil {
+				return err
+			}
+		} else if rr.br == nil {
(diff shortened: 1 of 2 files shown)


marstr:
Looks like CI is tripping, let me know if you have to make any significant changes after I reviewed :)

jhendrixMSFT:
@marstr @mcardosos can you please take a look at the latest iteration?

autorest/retriablerequest.go
```diff
@@ -0,0 +1,77 @@
...
+					return err
+				}
+			}
+			rr.br = bytes.NewReader(b)
+			rr.req.Body = ioutil.NopCloser(rr.br)
+		}
+		// indicates that the request body needs to be reset
+		rr.reset = true
+	}
+	return err
```
mcardosos: >return err [](start = 1, length = 10)

qq, why does this error on nil bodies?

mcardosos: Clumsy me, I had not seen the method signature :/

---
In reply to: [121535692](https://github.com/Azure/go-autorest/pull/143#discussion_r121535692) [](ancestors = 121535692)

jhendrixMSFT: I could remove the variable name here as technically it's not needed.

autorest/retriablerequest.go
```diff
@@ -0,0 +1,77 @@
...
+	"io/ioutil"
+	"net/http"
+)
+
+// NOTE: the GetBody() method on the http.Request object is new in 1.8.
+//       at present we support 1.7 and 1.8 so for now the branches specific
+//       to 1.8 have been commented out.
+
+// RetriableRequest provides facilities for retrying an HTTP request.
+type RetriableRequest struct {
```
marstr: I don't think that I'd make this type or its methods public.

autorest/retriablerequest.go
```diff
@@ -0,0 +1,77 @@
...
+	// preserve the request body; this is to support retry logic as
+	// the underlying transport will always close the reqeust body
+	if rr.req.Body != nil {
+		if rr.reset {
+			/*if rr.rc != nil {
+				rr.req.Body = rr.rc
+			} else */if rr.br != nil {
+				_, err = rr.br.Seek(0, io.SeekStart)
+			}
+			rr.reset = false
```
marstr: Oh man, this is a bit of a nit, but I don't like the name of this variable. To me `reset` means, "still in the state of having been reset" as opposed to "this needs to be reset before use." You've documented it well, so it's not the end of the world, but I'd prefer something like, `needsRewind` or reversing its usage to fit `atStart`.

autorest/retriablerequest.go
jhendrixMSFT: I made it public so we can consume it from the storage SDK.

marstr: Ah, got it. Well, I suppose it's what we've gotta do then.

Issue #147, opened 2017-06-20
Convenience func to add many decorators at once


mcardosos:
As suggested by @JackQuincy in https://github.com/Azure/azure-sdk-for-go/issues/658

JackQuincy:
Thanks

JackQuincy:
If you want the design I had, I actually have a branch in my fork that has this and a unit test for it. If you want I could start a PR. Let me know

mcardosos:
Awesome! PRs are always welcome :)

jhendrixMSFT:
Closing since the PR stalled.

PR #148, opened 2017-06-20
adding func to nest PrepareDecorators

JackQuincy:
adding func discussed in https://github.com/Azure/go-autorest/issues/147

### DIFF
--- a/autorest/preparer.go
+++ b/autorest/preparer.go
@@ -80,6 +80,16 @@ func WithNothing() PrepareDecorator {
 	}
 }
 
+// Nest returns a PrepareDecorator that nests the prepare decorators together.
+func Nest(pds ...PrepareDecorator) PrepareDecorator {
+	return func(p Preparer) Preparer {
+		for _, pd := range pds {
+			p = pd(p)
+		}
+		return p
+	}
+}
+
 // WithHeader returns a PrepareDecorator that sets the specified HTTP header of the http.Request to
 // the passed value. It canonicalizes the passed header name (via http.CanonicalHeaderKey) before
 // adding the header.
--- a/autorest/preparer_test.go
+++ b/autorest/preparer_test.go
@@ -713,6 +713,20 @@ func TestWithHeaderAllocatesHeaders(t *testing.T) {
 	}
 }
 
+func TestNestedWithHeaderAllocatesHeaders(t *testing.T) {
+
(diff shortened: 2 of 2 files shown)


autorest/preparer_test.go
```diff
@@ -713,6 +713,20 @@ func TestWithHeaderAllocatesHeaders(t *testing.T) {
 	}
 }
 
+func TestNestedWithHeaderAllocatesHeaders(t *testing.T) {
+
```
mcardosos: Nit, empty line

autorest/preparer.go
```diff
@@ -80,6 +80,16 @@ func WithNothing() PrepareDecorator {
 	}
 }
 
+// Nest returns a PrepareDecorator that nests the prepare decorators together.
+func Nest(pds ...PrepareDecorator) PrepareDecorator {
```
mcardosos: I don't think `Nest` is the best name here. Maybe `JoinPrepareDecorators` (as Go does not support method overloading, this wouldn't be very cool in case we wanted to also implement a similar func for ResponseDecorators).

marstr: If this makes sense for PrepareDecorators, I think it should make sense for ResponseDecorators as well. For that reason, I think we should add them both in the same PR and make sure their names don't collide. How about:

`func AggregatePreparerDecorator(pds ...PrepareDecorator) PrepareDecorator`
and
`func AggregateResponseDecorator(pds ...ResponseDecorator) ResponseDecorator`

autorest/preparer_test.go
```diff
@@ -713,6 +713,20 @@ func TestWithHeaderAllocatesHeaders(t *testing.T) {
...
+	r, err := Prepare(mocks.NewRequest(), Nest(WithHeader("x-foo", "bar"), WithHeader("x-foo2", "bar2")))
```
marstr: There is some irony in the fact that this actually would work without the `Nest` function as well :)

``` Go
r, err := Prepare(mocks.NewRequest(), WithHeader("x-foo", "bar"), WithHeader("x-foo2", "bar2")))
```

Given that the two do exactly the same thing, does it really make sense to add this function?

I'd love to hear your thoughts! I get the sense that there's just something I'm missing. :)

jhendrixMSFT:
Closing this as there's been no progress.  If this is still required please re-open.

PR #152, opened 2017-07-14
Added OnErrorFail to utils

mcardosos:
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/autorest/utils/errors.go
+++ b/autorest/utils/errors.go
@@ -1,0 +1,14 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+)
+
+// OnErrorFail prints a failure message and exits the program if err is not nil.
+func OnErrorFail(err error, message string) {
+	if err != nil {
+		fmt.Printf("%s: %s", message, err)
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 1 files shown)


autorest/utils/errors.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+// OnErrorFail prints a failure message and exits the program if err is not nil.
+func OnErrorFail(err error, message string) {
+	if err != nil {
+		fmt.Printf("%s: %s", message, err)
+		os.Exit(1)
```
jhendrixMSFT: I recall from a previous discussion that os.Exit() will not invoke deferred procedure calls which can be troublesome (e.g. deleting a resource group in a deferred proc).  Since this is used for samples I think we can simply panic() here instead although I'm open to other ideas as well (if we do go with panic then change the function name to "OnErrorPanic").

marstr:
@mcardosos, @jhendrixMSFT, you folks okay if I just close this PR? Seems like it's not getting a lot of attention and I have some misgivings about it.

mcardosos:
ok

PR #153, opened 2017-07-27
Uncomment retry logic codepaths that depend on Go 1.8.

jhendrixMSFT:
Now that Go 1.7 has been removed from CI we can enable this
optimization that relies on GetBody() which was added in 1.8.


marstr:
Looks like the removal of 1.7 from CI hasn't happened in this repository yet. May as well roll that change into this PR.

PR #154, opened 2017-08-03
Make RetriableRequest work with multiple versions of Go.

jhendrixMSFT:
Moved version-specific code to source files with appropriate build tags
and added earlier versions of Go to travis CI.
Exclude the unixtime tests from 1.6 as they rely on functionality that
isn't available in 1.6.

### DIFF
--- a/autorest/date/unixtime_test.go
+++ b/autorest/date/unixtime_test.go
@@ -1,3 +1,5 @@
+// +build go1.7
+
 package date
 
 import (
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ language: go
 
 go: 
   - 1.8
+  - 1.7
+  - 1.6
 
 install:
   - go get -u github.com/golang/lint/golint
--- a/autorest/retriablerequest.go
+++ b/autorest/retriablerequest.go
@@ -7,14 +7,6 @@ import (
 	"net/http"
 )
 
-// RetriableRequest provides facilities for retrying an HTTP request.
-type RetriableRequest struct {
-	req   *http.Request
-	rc    io.ReadCloser
-	br    *bytes.Reader
-	reset bool
-}
-
 // NewRetriableRequest returns a wrapper around an HTTP request that support retry logic.
 func NewRetriableRequest(req *http.Request) *RetriableRequest {
 	return &RetriableRequest{req: req}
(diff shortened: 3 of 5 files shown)


autorest/date/unixtime_test.go
```diff
@@ -1,3 +1,5 @@
+// +build go1.7
```
marstr: I'm a little confused on this one, did I use a feature that is only available in 1.7+ while implementing this? Should I rewrite portions so that Unixtime is available to 1.6 users as well? If not, does that mean packages that require Unixtime would also need the 1.7 build tag?

autorest/retriablerequest_1.x.go
```diff
@@ -0,0 +1,39 @@
+// +build !go1.8
```
marstr: I'm having trouble interpreting this in conjunction with the [build constraints documentation](https://golang.org/pkg/go/build/). Will this function as <1.8 or <1.8>? i.e. when 1.9 gets released, will this get built?

autorest/date/unixtime_test.go
jhendrixMSFT: Sorry I wasn't clear on this, specifically the "t.Run" used in these tests was added in Go 1.7.  I don't think it's really worth re-writing these tests to fix that as we have coverage on later versions unless there's an easy way to do it.

marstr: > Exclude the unixtime tests from 1.6 as they rely on functionality that
isn't available in 1.6.

Whoops, I should read _gooder_. However, I am still curious about what that means for the packages that rely on UnixTime.

autorest/retriablerequest_1.x.go
jhendrixMSFT: I did too as the docs here are pretty bad.  This translates to "versions less than 1.8".

autorest/date/unixtime_test.go
marstr: Got it, got it. Disregard this concern, I think it is fair to just be skipping these particular tests in 1.6 if that's the trouble.

autorest/retriablerequest_1.x.go
mcardosos: So when go1.9 is release, we have to add it to this list?

jhendrixMSFT: No this particular file is for versions before 1.8.  The other file will be used in versions 1.8+ (maybe I didn't do such a great job in naming them).

PR #155, opened 2017-08-07
Retry with 429 status code

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/588
PTAL @jhendrixMSFT  @marstr

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"math"
 	"net/http"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


autorest/sender.go
```diff
@@ -209,13 +210,31 @@ func DoRetryForStatusCodes(attempts int, backoff time.Duration, codes ...int) Se
...
 			}
 			return resp, err
 		})
 	}
 }
 
+// DelayWithRetryAfter invokes time.After for the duration specified in the "Retry-After" header in
+// responses with status code 429
+func DelayWithRetryAfter(resp *http.Response, cancel <-chan struct{}) bool {
+	retryAfter, _ := strconv.Atoi(resp.Header.Get("Retry-After"))
```
marstr: You could also time.ParseDuration(resp.Header.Get("Retry-After") + "s"), but It's not a big deal either way.

PR #156, opened 2017-08-08
Add support for bearer authentication callbacks.

jhendrixMSFT:
Creation of bearer authorization requires the tenant ID and resource URL
and in some cases it can be tricky to obtain the correct values leading
to authentication failures.  To simplify this we can rely on the bearer
authorization header to obtain the correct tenant ID and resource URL,
providing these bits of data via an authorization callback that's
invoked when a request is submitted.

### DIFF
--- a/autorest/authorization.go
+++ b/autorest/authorization.go
@@ -3,8 +3,16 @@ package autorest
 import (
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/Azure/go-autorest/autorest/adal"
+)
+
+const (
+	bearerChallengeHeader = "Www-Authenticate"
+	bearer                = "Bearer"
+	tenantID              = "tenantID"
 )
 
 // Authorizer is the interface that provides a PrepareDecorator used to supply request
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
This is in response to https://github.com/Azure/azure-sdk-for-go/issues/697, using autorest.NewBearerAuthorizationCallback() you can obtain the correct resource ID.

autorest/authorization.go
```diff
@@ -55,3 +63,105 @@ func (ba *BearerAuthorizer) WithAuthorization() PrepareDecorator {
...
+// is "Bearer " followed by the token.  The BearerAuthorizer is obtained via a user-supplied callback.
+//
+// By default, the token will be automatically refreshed through the Refresher interface.
+func (bacb *BearerAuthorizerCallback) WithAuthorization() PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			// make a copy of the request and remove the body as it's not
+			// required and avoids us having to create a copy of it.
+			rCopy := *r
+			removeRequestBody(&rCopy)
```
marstr: This is a really clever solution to our trouble with ReadCloser. Kudos.

PR #162, opened 2017-08-18
Support for parsing the Tokens out of the Azure CLI AccessTokens file

tombuildsstuff:
Both the Node.JS and Python Azure CLI's store the authentication tokens generated from `az(ure) login` in `~/.azure/accessTokens.json`.
Attempting to load this file using `adal.LoadToken()` fails due to there being multiple Tokens within a single file with a different schema - as such I've added a new method to load the tokens out of the file generated by the Azure CLI.

I've tested this on macOS with multiple subscriptions associated with the Azure account and this seems to work, but I'm unaware of if the file contents is different for users with a single subscription on the account?

### DIFF
--- a/autorest/adal/cli_token.go
+++ b/autorest/adal/cli_token.go
@@ -1,0 +1,32 @@
+package adal
+
+import "strconv"
+
+type AzureCLIToken struct {
+	AccessToken      string `json:"accessToken"`
+	Authority        string `json:"_authority"`
+	ClientID         string `json:"_clientId"`
+	ExpiresIn        int     `json:"expiresIn"`
+	ExpiresOn        string `json:"expiresOn"`
+	IdentityProvider string `json:"identityProvider"`
+	IsMRRT           bool    `json:"isMRRT"`
+	RefreshToken     string `json:"refreshToken"`
+	Resource         string `json:"resource"`
+	TokenType        string `json:"tokenType"`
+	UserId           string `json:"userId"`
+}
+
+func AzureCLIAccessTokensPath() string {
+	return "~/.azure/accessTokens.json"
+}
+
+func (t AzureCLIToken) ToToken() Token {
+	return Token{
+		AccessToken: t.AccessToken,
+		Type: t.TokenType,
+		ExpiresIn: strconv.Itoa(t.ExpiresIn),
+		ExpiresOn: t.ExpiresOn,
+		RefreshToken: t.RefreshToken,
+		Resource: t.Resource,
+	}
+}
--- a/autorest/adal/cli_token.go
+++ b/autorest/adal/cli_token.go
@@ -7,10 +7,10 @@ type AzureCLIToken struct {
 	AccessToken      string `json:"accessToken"`
 	Authority        string `json:"_authority"`
 	ClientID         string `json:"_clientId"`
-	ExpiresIn        int     `json:"expiresIn"`
+	ExpiresIn        int    `json:"expiresIn"`
 	ExpiresOn        string `json:"expiresOn"`
 	IdentityProvider string `json:"identityProvider"`
-	IsMRRT           bool    `json:"isMRRT"`
+	IsMRRT           bool   `json:"isMRRT"`
 	RefreshToken     string `json:"refreshToken"`
 	Resource         string `json:"resource"`
 	TokenType        string `json:"tokenType"`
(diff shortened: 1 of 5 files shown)


autorest/adal/cli_token.go
```diff
@@ -0,0 +1,32 @@
...
+	IdentityProvider string `json:"identityProvider"`
+	IsMRRT           bool    `json:"isMRRT"`
+	RefreshToken     string `json:"refreshToken"`
+	Resource         string `json:"resource"`
+	TokenType        string `json:"tokenType"`
+	UserId           string `json:"userId"`
+}
+
+func AzureCLIAccessTokensPath() string {
+	return "~/.azure/accessTokens.json"
```
marstr: I don't believe this is a portable way of finding the Azure CLI directory. When I run [this program](https://play.golang.org/p/H_YnVgs8uP) on my Windows machine I get unhelpful results.

Mind tinkering with this to find a more portable method?

autorest/adal/cli_token.go
```diff
@@ -0,0 +1,32 @@
+package adal
+
+import "strconv"
+
+type AzureCLIToken struct {
```
marstr: `golint` / `govet` are going to call this out as not having a comment and should cause CI to fail. Mind adding one? This applies to `func (t AzureCLIToken) ToToken() Token` as well.

autorest/adal/persist.go
```diff
@@ -25,6 +25,24 @@ func LoadToken(path string) (*Token, error) {
 	return &token, nil
 }
 
+// LoadCLITokens restores a set of AzureCLIToken objects from a file located at 'path'.
+func LoadCLITokens(path string) (*[]AzureCLIToken, error) {
```
marstr: Any reason you're returning a pointer to the array here? The actual object that would be copied in memory would be pretty small.

tombuildsstuff: Only because I wanted to return a `nil` object when an error is returned (to match the `LoadToken` function above - I can switch this out, but returning an empty object feels like an unexpected response IMO?

autorest/adal/cli_token.go
tombuildsstuff: Looks like the simplest way to do this is with [mitchellh/go-homedir](https://github.com/mitchellh/go-homedir) - I've pushed an update for this, and can confirm this works on macOS and Windows using this sample app:

```
$ cat main.go
package main

import "log"
import "github.com/mitchellh/go-homedir"

func main() {
  path, err := homedir.Expand("~/.azure/accessTokens.json")
  log.Printf("Path: %s", path)
  log.Printf("Error: %+v", err)
}
```

screen shot 2017-08-22 at 13 05 58

screen shot 2017-08-22 at 13 04 06

tombuildsstuff:
@marstr thanks for taking a look - I've updated this to point at `dev` + made the changes requested :)

autorest/adal/cli_token.go
marstr: Awesome, thanks Tom. This looks like it'll do the trick!

@jhendrixMSFT, @mcardosos: I verified that this library is using the MIT license (which is a very permissive license) and registered our usage of it internally.

autorest/adal/persist.go
jhendrixMSFT: Slices are nil-able so you should be able to return nil here without using a pointer yes?

marstr: Ah, good catch @jhendrixMSFT. This program does make it look like Slices are nilable: https://play.golang.org/p/q3RBbtvzW4

Maybe that nudges us towards having it not be a pointer :)

tombuildsstuff: Updated :)

tombuildsstuff:
@marstr no worries, thanks for reviewing so quickly :)

off-topic: it might be because I've been contributing to TF for a while - but it'd be really handy to have a Makefile for all of the common commands (see: formatting, linting, building etc, [here's ours for reference](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/GNUmakefile#L26)) - would you be interested if I sent a PR to add that? :)

marstr:
That sounds great! Would it be a literal `makefile` or just a shortcut command?

tombuildsstuff:
@marstr @mcardosos I've put together #163 after some playing around, not sure if that's helpful :)

mcardosos:
Awesome!
@tombuildsstuff Thanks for the contribution! :D

PR #166, opened 2017-08-31
Support for parsing the list of profiles from the Azure CLI entries

tombuildsstuff:
This PR adds support for parsing the list of profiles out of the Azure CLI files. Unfortunately the file written by the Python CLI writes a UTF8BOM at the start of the file, which the built-in JSON parser isn't so happy about and as such we need to filter the request out first; hence why I've brought in github.com/dimchansky/utfbom (which is licensed under Apache2)

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,38 +1,40 @@
-hash: 5e2681e794f14699a7fea0835f0947f59ff16d24dcbed0a53c462c7c69d09e3a
-updated: 2017-08-22T11:59:00.914426089+01:00
+hash: 6e0121d946623e7e609280b1b18627e1c8a767fdece54cb97c4447c1167cbc46
+updated: 2017-08-31T13:58:01.034822883+01:00
 imports:
 - name: github.com/dgrijalva/jwt-go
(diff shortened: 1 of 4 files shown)


glide.lock
```diff
@@ -1,38 +1,40 @@
...
-updated: 2017-08-22T11:59:00.914426089+01:00
+hash: 6e0121d946623e7e609280b1b18627e1c8a767fdece54cb97c4447c1167cbc46
+updated: 2017-08-31T13:58:01.034822883+01:00
 imports:
 - name: github.com/dgrijalva/jwt-go
   version: 2268707a8f0843315e2004ee4f1d021dc08baedf
   subpackages:
   - .
+- name: github.com/dimchansky/utfbom
+  version: 6c6132ff69f0f6c088739067407b5d32c52e1d0f
```
marstr: [Book-keeping] I've verified that this dependency uses an acceptable license, and that we will not need to modify the NOTICE file. Further more, I've registered our usage of this dependency using the internal tool.

PR #168, opened 2017-09-07
MSI on windows machines

mcardosos:


mcardosos:
Close and reopen to trigger travis

mcardosos:
PTAL @jhendrixMSFT @marstr 
I think @weinong might be interested on this too.
BTW IDK what is happening now with Travis :I

autorest/adal/token.go
```diff
@@ -31,8 +32,11 @@ const (
 	// OAuthGrantTypeRefreshToken is the "grant_type" identifier used in refresh token flows
 	OAuthGrantTypeRefreshToken = "refresh_token"
 
-	// managedIdentitySettingsPath is the path to the MSI Extension settings file (to discover the endpoint)
-	managedIdentitySettingsPath = "/var/lib/waagent/ManagedIdentity-Settings"
+	// managedIdentitySettingsLinuxPath is the path to the MSI Extension settings file (to discover the endpoint)
+	managedIdentitySettingsLinuxPath = "/var/lib/waagent/ManagedIdentity-Settings"
+
+	// managedIdentitySettingsWindowsPath is the path to the MSI Extension settings file (to discover the endpoint)
+	managedIdentitySettingsWindowsPath = "C:/WindowsAzure/Config/ManagedIdentity-Settings"
```
marstr: We shouldn't assume that the windows drive will always be "C" here.

autorest/adal/token.go
```diff
@@ -31,8 +32,11 @@ const (
 	// OAuthGrantTypeRefreshToken is the "grant_type" identifier used in refresh token flows
 	OAuthGrantTypeRefreshToken = "refresh_token"
 
-	// managedIdentitySettingsPath is the path to the MSI Extension settings file (to discover the endpoint)
-	managedIdentitySettingsPath = "/var/lib/waagent/ManagedIdentity-Settings"
+	// managedIdentitySettingsLinuxPath is the path to the MSI Extension settings file (to discover the endpoint)
+	managedIdentitySettingsLinuxPath = "/var/lib/waagent/ManagedIdentity-Settings"
```
marstr: If it were me, and I had two versions of this path to define, I'd employ the same tactics that the standard library took for path separator. There they use OS build tags to get the correct behavior. See here:

https://github.com/golang/go/tree/2d69e9e259ec0f5d5fbeb3498fbd9fed135fe869/src/path/filepath

autorest/adal/token.go
```diff
@@ -265,11 +268,18 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
 }
 
 // NewServicePrincipalTokenFromMSI creates a ServicePrincipalToken via the MSI VM Extension.
-func NewServicePrincipalTokenFromMSI(oauthConfig OAuthConfig, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
-	return newServicePrincipalTokenFromMSI(oauthConfig, resource, managedIdentitySettingsPath, callbacks...)
+func NewServicePrincipalTokenFromMSI(resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
```
marstr: Will it always make sense to fetch the MSI endpoint from a file, or should we be taking the endpoint as one of the parameters here? We could expose convenience functions for fetching the endpoint from a file like you do in the method below.

autorest/adal/token.go
mcardosos: This info comes from the documentation

autorest/adal/token.go
```diff
@@ -138,9 +142,8 @@ type ServicePrincipalMSISecret struct {
 }
 
 // SetAuthenticationValues is a method of the interface ServicePrincipalSecret.
-// MSI extension requires the authority field to be set to the real tenant authority endpoint
+// MSI extension no longer requires the authority field
```
jhendrixMSFT: Can this comment just be removed?

autorest/adal/token.go
jhendrixMSFT: Can it ever be something different, i.e. is this path user-configurable?  If so how does the end-user specify that?  I don't have a link to the docs.

autorest/adal/token.go
mcardosos: Fixed

autorest/adal/token.go
mcardosos: Fixed

autorest/adal/token.go
mcardosos: Fixed

autorest/adal/msi_linux.go
```diff
@@ -0,0 +1,4 @@
+// +build !windows
+
```
jhendrixMSFT: Missing package statement.

autorest/adal/msi_linux.go
```diff
@@ -0,0 +1,4 @@
...
+// msiPath is the path to the MSI Extension settings file (to discover the endpoint)
+var	msiPath = "/var/lib/waagent/ManagedIdentity-Settings"
```
marstr: Because we're protected behind a build tag, you may as well make this a `const` as the Unix version of this isn't dynamic.

autorest/adal/msi_windows.go
```diff
@@ -0,0 +1,6 @@
+// +build windows
+
+package adal
+
+// msiPath is the path to the MSI Extension settings file (to discover the endpoint)
+var msiPath = "C:/WindowsAzure/Config/ManagedIdentity-Settings"
```
marstr: We still shouldn't assume that this file lives on the `C` drive :) 

You could dynamically generate this like the following: https://play.golang.org/p/RtnOwH5tGu

jhendrixMSFT: We might want to use environment var SystemDrive instead of C:

autorest/adal/msi_linux.go
mcardosos: Fixed

autorest/adal/msi_windows.go
mcardosos: Fixed

mcardosos: Fixed

autorest/adal/token.go
```diff
@@ -264,41 +259,57 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
...
 
-// NewServicePrincipalTokenFromMSI creates a ServicePrincipalToken via the MSI VM Extension.
-func NewServicePrincipalTokenFromMSI(oauthConfig OAuthConfig, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
-	return newServicePrincipalTokenFromMSI(oauthConfig, resource, managedIdentitySettingsPath, callbacks...)
+// Resource enumerates resources from which MSI authentication is supported
+type Resource string
+
+const (
+	// VirtualMachine specifies MSI support for VMs
+	VirtualMachine Resource = "VM"
```
marstr: I think golint should be flagging this. I think the name is supposed to be `ResourceVirtualMachine`

Separately, I think I do like this strategy better than trying to reflect over actual resources to figure out if they're supported. This will scale better.

autorest/adal/token.go
```diff
@@ -264,41 +259,57 @@ func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID s
...
+)
+
+// GetMSIEndpoint gets the correct endpoint for MSI auth depending on which resource
+// code is running on.
+func GetMSIEndpoint(resource Resource) (string, error) {
```
marstr: Do we need to leave some space for expansion here? What if some of our future helper methods need information about the resource itself to get the MSI endpoint? Maybe I'm crazy, convince me that this isn't the case :)

autorest/adal/msi_windows.go
```diff
@@ -0,0 +1,11 @@
...
+
+package adal
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// msiPath is the path to the MSI Extension settings file (to discover the endpoint)
+var msiPath = filepath.Join(os.Getenv("SystemDrive"), "WindowsAzure/Config/ManagedIdentity-Settings")
```
marstr: When I run this on my machine I get the following, which is missing a slash!

`C:WindowsAzure\Config\ManagedIdentity-Settings`

autorest/adal/token.go
mcardosos: Example?

autorest/adal/msi_windows.go
mcardosos: fixed

autorest/adal/token.go
mcardosos: Chatting offline we have concluded it is a better idea to have a func for each resource.

colemickens:
:)

mcardosos:
@colemickens :D :D :D

PR #169, opened 2017-09-08
Fixing AzureCLI / CloudShell authentication

tombuildsstuff:
- Hard-coding the value for `expiresIn` due to the Azure CLI and CloudShell authentication models having different types
 - Updating `expiresOn` to be the difference in seconds between the baseDate & now - rather than a date
 - Parsing the AzureCLI time out in the local system timezone, given it's saved in local time but with no timezone info


autorest/adal/cli.go
```diff
@@ -47,13 +50,42 @@ func AzureCLIProfilePath() (string, error) {
...
+		ExpiresOn:    strconv.Itoa(int(seconds)),
 		RefreshToken: t.RefreshToken,
 		Resource:     t.Resource,
 	}
+	return &token, nil
+}
+
+// ParseAzureCLIExpirationDate parses either a Azure CLI or CloudShell date into a time object
+func ParseAzureCLIExpirationDate(input string) (*time.Time, error) {
+	log.Printf("[DEBUG] Token Date: %s", input)
```
marstr: Let's remove this line before we commit this.

autorest/adal/cli.go
```diff
@@ -47,13 +50,42 @@ func AzureCLIProfilePath() (string, error) {
...
+	difference := tokenExpirationDate.Sub(expirationBase)
+	seconds := difference.Seconds()
+
+	token := Token{
 		AccessToken:  t.AccessToken,
 		Type:         t.TokenType,
-		ExpiresIn:    strconv.Itoa(t.ExpiresIn),
-		ExpiresOn:    t.ExpiresOn,
+		ExpiresIn:    "3600",
+		ExpiresOn:    strconv.Itoa(int(seconds)),
```
marstr: [nit] I think this would be more readable as `difference.Seconds()` and skipping the declaration of `seconds`

autorest/adal/cli.go
```diff
@@ -47,13 +50,42 @@ func AzureCLIProfilePath() (string, error) {
...
 		RefreshToken: t.RefreshToken,
 		Resource:     t.Resource,
 	}
+	return &token, nil
+}
+
+// ParseAzureCLIExpirationDate parses either a Azure CLI or CloudShell date into a time object
+func ParseAzureCLIExpirationDate(input string) (*time.Time, error) {
```
michalpristas: Q: why do you return pointer to time.
time package uses value semantics by default, returning pointer causes one more allocation, moves object from stack to heap and creates work for garbage collector. as it is used only within a calling method it should return value.
this way it would stay on stack and get disposed when method finishes. no additional work for GC.

tombuildsstuff: I’ve opted for this approach since we’re returning nil for an invalid/unparseable date; given returning an empty object in addition to an error in the failure case feels misleading as a consumer of the API

autorest/adal/cli.go
tombuildsstuff: Updated :)

autorest/adal/cli.go
tombuildsstuff: Updated :)

autorest/adal/cli.go
marstr: It seems like you use that pattern pretty religiously, @tombuildsstuff. I'm not really convinced either way, other than sorta general intuitiveness, can you build a case for why it's better to return a pointer? Personally, when I'm writing Go, I just treat retvals that accompany an error as garbage, it doesn't matter if the value is the default or memory with random bits. As you can imagine, in my workflow, changing the return type to a pointer has almost no consequence. What scenario am I missing where it's totally important? :)

autorest/adal/cli.go
```diff
@@ -1,16 +1,18 @@
 package adal
```
marstr: I thought about this a little this morning, and think that this code would be better off in a `cli` package. Particularly, I think it makes sense as a child of the [`azure` package](https://github.com/Azure/go-autorest/tree/5432abe734f8d95c78340cd56712f912906e6514/autorest/azure). In general, I'd like to have as few Azure specific references in the `adal` package as possible, that applies doubly to CLI specific code :)

tombuildsstuff: updated :)

autorest/adal/cli.go
tombuildsstuff: not particularly - I've only opted for this approach to force the consumer to handle the error case, rather than accidental comparing the values returned by the default object (it's more habit than anything else) - but totally happy to switch this out :)

marstr:
Talked with @jhendrixMSFT offline, he gave us the go-ahead to merge.

PR #170, opened 2017-09-10
Added support for unwrapped Azure API error responses

michalpristas:
Should solve problem described here: https://github.com/Azure/azure-sdk-for-go/issues/748
The problem is that result of API call is ServiceError itself, not ServiceError wrapped in error object. Which causes autorest to return 'Unknown' error instead.

Fixes: 748

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -165,7 +165,13 @@ func WithErrorUnlessStatusCode(codes ...int) autorest.RespondDecorator {
 				if decodeErr != nil {
 					return fmt.Errorf("autorest/azure: error response cannot be parsed: %q error: %v", b.String(), decodeErr)
 				} else if e.ServiceError == nil {
-					e.ServiceError = &ServiceError{Code: "Unknown", Message: "Unknown service error"}
+					// Check if error is unwrapped ServiceError
+					var se *ServiceError
+					if err := json.Unmarshal(b.Bytes(), &se); err == nil && se.Message != "" {
+						e.ServiceError = se
+					} else {
+						e.ServiceError = &ServiceError{Code: "Unknown", Message: "Unknown service error"}
+					}
 				}
 
 				e.RequestID = ExtractRequestID(resp)
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -170,7 +170,10 @@ func WithErrorUnlessStatusCode(codes ...int) autorest.RespondDecorator {
 					if err := json.Unmarshal(b.Bytes(), &se); err == nil && se.Message != "" {
 						e.ServiceError = se
 					} else {
-						e.ServiceError = &ServiceError{Code: "Unknown", Message: "Unknown service error"}
+						e.ServiceError = &ServiceError{
+							Code:    "Unknown",
+							Message: "Unknown service error",
+						}
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


autorest/azure/azure.go
```diff
@@ -165,7 +165,13 @@ func WithErrorUnlessStatusCode(codes ...int) autorest.RespondDecorator {
...
 					return fmt.Errorf("autorest/azure: error response cannot be parsed: %q error: %v", b.String(), decodeErr)
 				} else if e.ServiceError == nil {
-					e.ServiceError = &ServiceError{Code: "Unknown", Message: "Unknown service error"}
+					// Check if error is unwrapped ServiceError
+					var se *ServiceError
+					if err := json.Unmarshal(b.Bytes(), &se); err == nil && se.Message != "" {
+						e.ServiceError = se
+					} else {
+						e.ServiceError = &ServiceError{Code: "Unknown", Message: "Unknown service error"}
+					}
```
marstr: I believe this should be updated to:

``` Go
var se ServiceError
if err := json.Unmarshal(b.Bytes(), &se); err == nil && se.Message != "" {
    e.ServiceError = &se
} else {
    e.ServiceError = &ServiceError{
        Code: "Unknown",
        Message: "Unknown service error",
    }
}
```

autorest/azure/azure_test.go
```diff
@@ -333,6 +333,64 @@ func TestWithErrorUnlessStatusCode_NoAzureError(t *testing.T) {
...
+	mocks.SetResponseHeader(r, HeaderRequestID, uuid)
+	r.Request = mocks.NewRequest()
+	r.StatusCode = http.StatusInternalServerError
+	r.Status = http.StatusText(r.StatusCode)
+
+	err := autorest.Respond(r,
+		WithErrorUnlessStatusCode(http.StatusOK),
+		autorest.ByClosing())
+	if err == nil {
+		t.Fatalf("azure: returned nil error for proper error response")
```
marstr: Using `t.Fatalf` ends all test execution. Please update these calls to be of the pattern:

``` Go
t.Logf("[your stuff]", [any args you need])
t.Fail()
```

autorest/azure/azure_test.go
```diff
@@ -333,6 +333,64 @@ func TestWithErrorUnlessStatusCode_NoAzureError(t *testing.T) {
...
+	if expected := "Azure is having trouble right now."; azErr.ServiceError.Message != expected {
+		t.Fatalf("azure: got wrong Message=%s Expected=%s", azErr.Message, expected)
+	}
+
+	if expected := uuid; azErr.RequestID != expected {
+		t.Fatalf("azure: wrong request ID in error. expected=%q; got=%q", expected, azErr.RequestID)
+	}
+
+	details, _ := json.Marshal(*azErr.ServiceError.Details)
+	if string(details) != `[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]` {
```
marstr: A Go idiom is to print out "got" and "want" in this case, one would expect the Unit Test Log to read the following on failure:

```
got:  "{contents of details}"
want: "[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]"
```

Mind refactoring this?

michalpristas: i will prepare fix today

michalpristas:
@marstr is something missing?

marstr:
Sorry, to leave you hanging @michalpristas! Thanks for making the recommended changes. Merging this in now.

PR #171, opened 2017-09-13
Fix MSI path not being included on macOS builds

jbfm:
`msi_linux.go` would only be included on Linux machines, and not non-Windows machines as indicated by the build tag.


marstr:
Thanks for catching this and submitting a PR, it is much appreciated.

PR #172, opened 2017-09-15
Adding Apache 2.0 Headers

marstr:
Also adding a check for the Microsoft Copyright to CI.


autorest/mocks/mocks_test.go
```diff
@@ -1 +0,0 @@
-package mocks
```
mcardosos: LOL!

PR #174, opened 2017-09-29
Auth file support [WIP]

mcardosos:
[Useful documentation](https://docs.microsoft.com/en-us/python/azure/python-sdk-azure-authenticate?view=azure-python#mgmt-auth-file)
This would go hand in hand with generated code that I want to make it look like this...
```go
func NewAccountsClientWithAuthFile() (ac AccountsClient, err error) {
	auth, err := utils.GetTokenFromAuthFile(DefaultBaseURI)
	if err != nil {
		return
	}
	ac = NewAccountsClient(auth.File["subscriptionId"])
	ac.BaseURI = auth.BaseURI
	ac.Authorizer = auth.Authorizer
	return
}
```

I have tried this on a sample and it works.
PTAL @jhendrixMSFT @marstr


autorest/utils/auth.go
```diff
@@ -40,3 +52,72 @@ func GetEnvVarOrExit(varName string) string {
...
+	auth.Authorizer = autorest.NewBearerAuthorizer(spToken)
+	return
+}
+
+func decode(b []byte) []byte {
+	utf16leBOM := []byte{255, 254}
+	if bytes.HasPrefix(b, utf16leBOM) {
+		b = bytes.TrimPrefix(b, utf16leBOM)
+
+		decoded := new(bytes.Buffer)
```
jhendrixMSFT: Comment why you have to do this.

autorest/utils/auth.go
```diff
@@ -40,3 +52,72 @@ func GetEnvVarOrExit(varName string) string {
...
+	if err != nil {
+		return
+	}
+
+	auth.Authorizer = autorest.NewBearerAuthorizer(spToken)
+	return
+}
+
+func decode(b []byte) []byte {
+	utf16leBOM := []byte{255, 254}
```
jhendrixMSFT: Handle other BOMs?

autorest/utils/auth.go
jhendrixMSFT: You could also set a size since we know it will be smaller than b.

mcardosos:
For @yugangw-msft curiosity

autorest/utils/auth.go
mcardosos: Added more BOMs

autorest/utils/auth.go
mcardosos: Fixed

mcardosos:
So I moved this code to a new package that makes more sense than `utils`. I need to update my generator PR to reflect that. Is everything else on this PR ok for you guys @marstr @jhendrixMSFT ? So @yugangw-msft reviews too

autorest/azure/auth/authfile_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestGetTokenFromAuthFile(t *testing.T) {
+	os.Setenv("AZURE_AUTH_LOCATION", getCredsPath())
+	auth, err := GetTokenFromAuthFile("https://management.azure.com")
+	if err != nil {
+		t.Fatalf("GetTokenFromAuthFile failed, got error %v", err)
```
marstr: `Fatalf` should only be used when other errors wouldn't also make sense to report. For these first ones, it looks like that may make sense. Later in this Test though,it seems like you could replace this with subsequent calls to `t.Logf` and `t.Fail` respectively.

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Auth includes authentication details for ARM clients
+type Auth struct {
+	Authorizer *autorest.BearerAuthorizer
+	File       map[string]string
+	BaseURI    string
+}
+
+// GetTokenFromAuthFile creates an authorizer from an Azure CLI auth file
+func GetTokenFromAuthFile(baseURI string) (auth Auth, err error) {
```
marstr: I love named parameters. Thanks for using them!

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,103 @@
...
+var pcEndpoints = map[string]string{
+	"managementEndpointUrl":          "https://management.core.windows.net",
+	"resourceManagerEndpointUrl":     "https://management.azure.com",
+	"activeDirectoryEndpointUrl":     "https://login.microsoftonline.com",
+	"galleryEndpointUrl":             "https://gallery.azure.com",
+	"activeDirectoryGraphResourceId": "https://graph.windows.net",
+}
+
+// Auth includes authentication details for ARM clients
+type Auth struct {
```
marstr: I think this type (or actually I think you should swap it out for a new type based on what the CLI did, see below) should implement this interface: https://github.com/Azure/go-autorest/blob/faeb847c8586092daaa111b17cc4e6c792fd3a87/autorest/authorization.go#L21

autorest/azure/auth/authfile_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if auth.Authorizer == nil {
+		t.Fatalf("auth.Authorizer not set correctly, got nil")
+	}
+}
+
+func getCredsPath() string {
+	gopath := os.Getenv("GOPATH")
+	return filepath.Join(gopath, "src", "github.com", "Azure", "go-autorest", "autorest", "azure", "auth", "mycreds.json")
```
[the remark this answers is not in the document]
mcardosos: It is absolute so `go test` can be run from any directory in the repo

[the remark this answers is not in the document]
mcardosos: Fixed adding more files with different BOMs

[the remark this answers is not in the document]
mcardosos: Fixed

autorest/azure/auth/authfile_test.go
mcardosos: Fixed

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,103 @@
...
+	return ioutil.ReadAll(reader)
+}
+
+func getResourceKey(baseURI string) string {
+	for k, v := range pcEndpoints {
+		if baseURI == v {
+			return k
+		}
+	}
+	return ""
```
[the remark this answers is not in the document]
mcardosos: Fixed

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,103 @@
...
+		err := binary.Read(reader, binary.BigEndian, &u16)
+		if err != nil {
+			return nil, err
+		}
+		return []byte(string(utf16.Decode(u16))), nil
+	}
+	return ioutil.ReadAll(reader)
+}
+
+func getResourceKey(baseURI string) string {
```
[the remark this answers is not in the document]
mcardosos: Did something different, take a look :D

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+
+// File represents the authentication file
+type File struct {
+	ClientID                string `json:"clientId,ompitempty"`
+	ClientSecret            string `json:"clientSecret,ompitempty"`
+	SubscriptionID          string `json:"subscriptionId,ompitempty"`
+	TenantID                string `json:"tenantId,ompitempty"`
+	ActiveDirectoryEndpoint string `json:"activeDirectoryEndpointUrl,ompitempty"`
+	ResourceManagerEndpoint string `json:"resourceManagerEndpointUrl,ompitempty"`
+	GraphResourceID         string `json:"activeDirectoryGraphResourceId,ompitempty"`
```
marstr: `ompitempty` typo

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/dimchansky/utfbom"
+)
+
+// Auth includes authentication details for ARM clients
+type Auth struct {
+	Authorizer *autorest.BearerAuthorizer
```
marstr: I'd still like this to be directly exposed, but other than that, this PR looks a lot better! Thanks for entertaining me, @mcardosos!

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+	"unicode/utf16"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/dimchansky/utfbom"
+)
+
+// Auth includes authentication details for ARM clients
+type Auth struct {
```
jhendrixMSFT: This type name stutters.  I don't have enough context here to suggest a better name.

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"unicode/utf16"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/dimchansky/utfbom"
```
jhendrixMSFT: Do we feel that this is a good package to take a dependency on?

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+	ActiveDirectoryEndpoint string `json:"activeDirectoryEndpointUrl,omitempty"`
+	ResourceManagerEndpoint string `json:"resourceManagerEndpointUrl,omitempty"`
+	GraphResourceID         string `json:"activeDirectoryGraphResourceId,omitempty"`
+	SQLManagementEndpoint   string `json:"sqlManagementEndpointUrl,omitempty"`
+	GalleryEndpoint         string `json:"galleryEndpointUrl,omitempty"`
+	ManagementEndpoint      string `json:"managementEndpointUrl,omitempty"`
+}
+
+// GetTokenFromAuthFile creates an authorizer from an Azure CLI auth file
+func GetTokenFromAuthFile(baseURI string) (auth Auth, err error) {
```
jhendrixMSFT: I don't think this function name really matches with what it does, i.e. it doesn't simply return a token.  Probably could use a rename, possibly based on the renaming of the Auth type.

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,124 @@
...
+		err := binary.Read(reader, binary.BigEndian, &u16)
+		if err != nil {
+			return nil, err
+		}
+		return []byte(string(utf16.Decode(u16))), nil
+	}
+	return ioutil.ReadAll(reader)
+}
+
+func getResource(f File, baseURI string) (string, error) {
```
jhendrixMSFT: Rename to something more indicative of what this function does (I guess it translates public cloud base URIs to something?).

autorest/azure/auth/authfile.go
mcardosos: There was already a dependency on this package somewhere else on this repo

marstr: For what it's worth, I added this as a registered open source dependency a couple of months ago when Tom added it.

autorest/azure/auth/authfile.go
mcardosos: Fixed

autorest/azure/auth/authfile.go
mcardosos: Fixed

autorest/azure/auth/authfile.go
mcardosos: Fixed

autorest/azure/auth/authfile.go
```diff
@@ -0,0 +1,129 @@
...
+
+	auth.BearerAuthorizer = autorest.NewBearerAuthorizer(spToken)
+	return
+}
+
+func decode(b []byte) ([]byte, error) {
+	reader, enc := utfbom.Skip(bytes.NewReader(b))
+
+	switch enc {
+	case utfbom.UTF16LittleEndian:
```
jhendrixMSFT: A minor issue, this could probably be refactored to remove the duplication unless it makes the code harder to read.

mcardosos: Well, it would not be as cute
``` go
func decode(b []byte) (d []byte, err error) {
	reader, enc := utfbom.Skip(bytes.NewReader(b))
	if enc == utfbom.UTF16LittleEndian || enc == utfbom.UTF16BigEndian {
		u16 := make([]uint16, (len(b)/2)-1)
		if enc == utfbom.UTF16LittleEndian {
			err = binary.Read(reader, binary.LittleEndian, &u16)

		} else {
			err = binary.Read(reader, binary.BigEndian, &u16)
		}
		if err != nil {
			return nil, err
		}
		return []byte(string(utf16.Decode(u16))), nil
	}
	return ioutil.ReadAll(reader)
}
```

PR #176, opened 2017-10-05
Auto register resource providers

mcardosos:
This would also need to add more generated code, it would look  like this...
```go
func (client TopicsClient) CreateOrUpdateSender(req *http.Request) (*http.Response, error) {
	return autorest.SendWithSender(client,
		req,
		azure.RegisterResourceProvider(),
		autorest.DoRetryForStatusCodes(client.RetryAttempts, client.RetryDuration, autorest.StatusCodesForRetry...),
		azure.DoPollForAsynchronous(client.PollingDelay),
	)
}
```
PTAL @jhendrixMSFT @marstr


marstr:
Before I dive into the code too much here, I want to understand how it relates to this method:

[azure-sdk-for-go/arm/resources/resources/providers.go#L252 @ 7362b8b](https://github.com/Azure/azure-sdk-for-go/blob/7362b8bf379bde4d1e26e91ee424ea6c5d8796d3/arm/resources/resources/providers.go#L252)

Is it a case of circular references getting in our way that we wouldn't reuse this?

mcardosos:
I would prefer not to reference the SDK in the runtime. It would also be a circular reference.

jhendrixMSFT:
@marstr brings up a good point, if we're already generating code that does the work then should we adpot it instead.  Obviously we can't reference the Go SDK due to circular dependencies, but perhaps it makes sense to copy/paste this code instead (i.e. would it set us up better for future servicing of this code).

mcardosos:
I copied and pasted code, and left out what made not so much sense in this context (like unmarshaling the response body)

mcardosos:
After fixing stuff, generated code would look like this...
``` go
func (client TopicsClient) CreateOrUpdateSender(req *http.Request) (*http.Response, error) {
	return autorest.SendWithSender(client,
		req,
		azure.RegisterResourceProvider(),
		azure.DoPollForAsynchronous(client.PollingDelay),
	)
}
```

mcardosos:
This rework with the Azure specific client no longer requires changes on senders in generated code. But it will require changes in the clients file.
``` diff
// Azure EventGrid Management Client
package eventgrid

import (
)

const (
	// DefaultBaseURI is the default URI used for the service Eventgrid
	DefaultBaseURI = "https://management.azure.com"
)

// ManagementClient is the base client for Eventgrid.
type ManagementClient struct {
-	autorest.Client
+	azure.Client
	BaseURI        string
	SubscriptionID string
}

// New creates an instance of the ManagementClient client.
func New(subscriptionID string) ManagementClient {
	return NewWithBaseURI(DefaultBaseURI, subscriptionID)
}

// NewWithBaseURI creates an instance of the ManagementClient client.
func NewWithBaseURI(baseURI string, subscriptionID string) ManagementClient {
	return ManagementClient{
-		Client:         autorest.NewClientWithUserAgent(UserAgent()),
+		Client: azure.Client{
+			Client: autorest.NewClientWithUserAgent(UserAgent()),
+		},
		BaseURI:        baseURI,
		SubscriptionID: subscriptionID,
	}
}
```

PTAL @jhendrixMSFT

autorest/azure/sender.go
```diff
@@ -0,0 +1,187 @@
...
+	if re.ServiceError != nil {
+		if re.ServiceError.Details != nil && len(*re.ServiceError.Details) > 0 {
+			detail := (*re.ServiceError.Details)[0].(map[string]interface{})
+			return detail["target"].(string), nil
+		}
+	}
+	return "", errors.New("provider was not found in the response")
+}
+
+func register(azc Client, s autorest.Sender, originalReq *http.Request, re RequestError) error {
```
jhendrixMSFT: I would turn this into a method where (azc Client) is the receiver.  Also doesn't the client implement autorest.Sender?  If so then it seems you could remove this param.

autorest/azure/client.go
```diff
@@ -0,0 +1,62 @@
...
+func (azc Client) sender() autorest.Sender {
+	if azc.Sender == nil {
+		j, _ := cookiejar.New(nil)
+		return &http.Client{Jar: j}
+	}
+	return azc.Sender
+}
+
+// AzurePreparation is a prepare decorator that adds a User Agent and Authorization headers to the request
+func (azc Client) AzurePreparation() autorest.PrepareDecorator {
```
jhendrixMSFT: I don't think this needs to be exported does it?

autorest/azure/sender.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// DoRetryForStatusCodes is the Azure specific implementation for retring requests
```
jhendrixMSFT: I would update this comment to mention that it will automatically attempt to register the resource provider.

autorest/azure/sender.go
mcardosos: I need the client to get authorization again

jhendrixMSFT: Sorry I meant remove the autorest.Sender param.

mcardosos: I use that one for the test :(

autorest/azure/client.go
mcardosos: Fixed

autorest/azure/sender.go
mcardosos: Fixed

mcardosos:
With this changes, we can use the clients the same as before (the client only has a different type)
``` go
eventgridClient.RetryDuration = time.Second
eventgridClient.RetryAttempts = 3
eventgridClient.PollingDelay = time.Second * 10
eventgridClient.PollingDuration = time.Minute * 5
resp, errChan := eventgridClient.CreateOrUpdate(groupName, name, eventgrid.Topic{
	Location: to.StringPtr("westcentralus"),
}, nil)
```
This is valid code with and without the change mentioned above to generated code.

autorest/azure/sender.go
```diff
@@ -0,0 +1,189 @@
...
+						resp,
+						autorest.ByUnmarshallingJSON(&re),
+						autorest.ByClosing(),
+					)
+					if err != nil {
+						return resp, err
+					}
+
+					if re.ServiceError != nil && re.ServiceError.Code == "MissingSubscriptionRegistration" {
+						azc.Sender = s
```
jhendrixMSFT: Presumably for non-test scenarios azc.Sender and s are identical?  It would be good to comment why you need to do this assignment.

mcardosos: They are not identical, but they contain the same information

mcardosos:
Reworked code to get rid of Azure client. Very important change to note is I removed the retries from the `Do` implementation for `autorest.Client`. I have also fixed the retriable requests bug.
Generated code would look like this, there are no more breaking changes:
``` diff
func (client TopicsClient) CreateOrUpdateSender(req *http.Request) (*http.Response, error) {
	return autorest.SendWithSender(client,
		req,
+		azure.RegisterResourceProvider(client.Client),
		azure.DoPollForAsynchronous(client.PollingDelay),
	)
}
```
PTAL @jhendrixMSFT @marstr

autorest/azure/rp.go
```diff
@@ -0,0 +1,195 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// RegisterResourceProvider  tries to register the resource provider in case it is unregistered
+func RegisterResourceProvider(client autorest.Client) autorest.SendDecorator {
```
jhendrixMSFT: Since this is replacing DoRetryForStatusCodes() and augmenting it with registration I would rename this to something that reflects that, maybe DoRetryWithRegistration() or something along those lines.

autorest/azure/rp.go
```diff
@@ -0,0 +1,195 @@
...
+
+				if re.ServiceError != nil && re.ServiceError.Code == "MissingSubscriptionRegistration" {
+					// in tests, the sender should be the same for all requests
+					client.Sender = s
+					err = register(client, r, re)
+					if err != nil {
+						return resp, fmt.Errorf("failed auto registering Resource Provider: %s", err)
+					}
+				}
+				err = rr.Prepare()
```
jhendrixMSFT: I think this code and the lines below it can be removed, refactoring the code starting at line 19 into a loop.

autorest/azure/rp.go
mcardosos: Fixed

autorest/azure/rp.go
mcardosos: Fixed

autorest/azure/rp.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// DoRetryWithRegistration tries to register the resource provider in case it is unregistered.
+// It also handles request retries
+func DoRetryWithRegistration(client autorest.Client) autorest.SendDecorator {
+	return func(s autorest.Sender) autorest.Sender {
+		return autorest.SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
+			shouldRetry := true
```
jhendrixMSFT: I think the var shouldRetry can be removed.

autorest/azure/rp.go
```diff
@@ -0,0 +1,190 @@
...
+
+				resp, err = autorest.SendWithSender(s, rr.Request(),
+					autorest.DoRetryForStatusCodes(client.RetryAttempts, client.RetryDuration, autorest.StatusCodesForRetry...),
+				)
+				if err != nil {
+					return resp, err
+				}
+				shouldRetry = false
+
+				if resp.StatusCode == http.StatusConflict {
```
jhendrixMSFT: After removing the shouldRetry you can change this to exit early on the non-conflicting status.
```go
if resp.StatusCode != http.StatusConflict {
    return resp, err
}
var re RequestError
...
```

mcardosos: done :3

autorest/azure/rp.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return fmt.Errorf("missing parameter provider to register resource provider: %s", err)
+	}
+	newURL := url.URL{
+		Scheme: originalReq.URL.Scheme,
+		Host:   originalReq.URL.Host,
+	}
+
+	// taken from the resources SDK
+	// https://github.com/Azure/azure-sdk-for-go/blob/9f366792afa3e0ddaecdc860e793ba9d75e76c27/arm/resources/resources/providers.go#L252
```
marstr: Could you update this to also include an explanation of why we had to Copy/Paste?

autorest/azure/rp.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+// DoRetryWithRegistration tries to register the resource provider in case it is unregistered.
+// It also handles request retries
+func DoRetryWithRegistration(client autorest.Client) autorest.SendDecorator {
+	return func(s autorest.Sender) autorest.Sender {
+		return autorest.SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
+			attempts := 0
+			rr := autorest.NewRetriableRequest(r)
+			for attempts < client.RetryAttempts {
```
marstr: [nit] I'd rename `attempts` to something that didn't sound static like `currentAttempt`. Also, it would be a little more readable/conventional to declare `attempts` in the for body like so:

``` Go
for currentAttempt := 0; currentAttempt < client.RetryAttempts; currentAttempt++ {
```

autorest/azure/rp.go
mcardosos: Fixed

autorest/azure/rp.go
mcardosos: Fixed

autorest/azure/rp.go
marstr: Hmmm, I'm not sure I like the updated comment any better. What I'd like to see expressed is that we can't/don't want to take a dependency on the github.com/Azure/azure-sdk-for-go repository. For that reason, we must have a manually maintained copy here, and it made sense to just grab what we had generated.

PR #178, opened 2017-10-05
Updating to add AzureStack cloud.

radhikagupta5:


radhikagupta5:
Updating to add AzureStack cloud. Since Azure Stack is a variable endpoint deployment, we are adding ability to bring your own endpoint.

autorest/azure/environments.go
```diff
@@ -126,5 +135,19 @@ func EnvironmentFromName(name string) (Environment, error) {
 	if !ok {
 		return env, fmt.Errorf("autorest/azure: There is no cloud environment matching the name %q", name)
 	}
+
+	if strings.EqualFold(name, "AZURESTACKCLOUD") == true {
```
marstr: There seems like enough logic here that it calls for its own constructor. I'm thinking that maybe we just should have "AzureStackCloud" be one of the named environments.

autorest/azure/environments.go
```diff
@@ -117,6 +121,11 @@ var (
 		ResourceManagerVMDNSSuffix:   "cloudapp.microsoftazure.de",
 		ContainerRegistryDNSSuffix:   "azurecr.io",
 	}
+
+	//AzureStackCloud is the Azure enviornment running in customer datacenter
+	AzureStackCloud = Environment{
+		Name: "AzureStackCloud",
```
marstr: Without populating any of the other fields in this variable, I don't think it has any value. As mentioned in another comment, it seems like populating a Stack environment should be done as its own constructor or set of constructors.

The comment that I mentioned copied here for convenience:
> There seems like enough logic here that it calls for its own constructor. I'm thinking that maybe we just should have "AzureStackCloud" be one of the named environments.

autorest/azure/environments.go
```diff
@@ -126,5 +135,19 @@ func EnvironmentFromName(name string) (Environment, error) {
 	if !ok {
 		return env, fmt.Errorf("autorest/azure: There is no cloud environment matching the name %q", name)
 	}
+
+	if strings.EqualFold(name, "AZURESTACKCLOUD") == true {
+
+		println("autorest/azure: Reading Azure Stack Cloud config from directory", path.Join("etc", "kubernetes", "azurestackcloud.json"))
+		fbytes, err := ioutil.ReadFile(path.Join("/etc", "kubernetes", "azurestackcloud.json"))
```
marstr: We should be very deliberate about adding references to third party placement of these files. I'd really rather the path where you can find the environment file be a parameter to a constructor.

marstr:
In addition, please update this PR to target `dev` instead of `master`. Thank you!

autorest/azure/environments.go
```diff
@@ -126,5 +135,19 @@ func EnvironmentFromName(name string) (Environment, error) {
 	if !ok {
 		return env, fmt.Errorf("autorest/azure: There is no cloud environment matching the name %q", name)
 	}
+
+	if strings.EqualFold(name, "AZURESTACKCLOUD") == true {
+
+		println("autorest/azure: Reading Azure Stack Cloud config from directory", path.Join("etc", "kubernetes", "azurestackcloud.json"))
```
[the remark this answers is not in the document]
radhikagupta5: Removed.

autorest/azure/environments.go
radhikagupta5: Sure. Changed Kubernetes to azurestackcloud.

radhikagupta5: Sure. Changed kubernetes to azurestackcloud.

autorest/azure/environments.go
radhikagupta5: In AzureStack each stamp has a different environment and these values varies based on customer input. We don't know at the time of building this code what that information is. To keep this code independent of the environment we are changing to a model to bring your own endpoint. In this case we are reading the values from an existing JSON file. Yes, the json file should be present (in this case placed by ACS-Engine), but worse case we fail with an appropriate error.

radhikagupta5: For example, this is the value for an env.

"name": "AzurestackCloud",
   "managementPortalURL": "",
   "publishSettingsURL": "",
   "serviceManagementEndpoint": "https://management.azurestackci13.onmicrosoft.com/e55c980f-9a38-48ab-a059-87027f8134b1",
   "resourceManagerEndpoint": "https://management.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
   "activeDirectoryEndpoint": "https://login.windows.net/",
   "galleryEndpoint": "https://portal.redmond.ext-u15e0303.masd.stbtest.microsoft.com:30015/",
   "keyVaultEndpoint": "https://vault.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
   "graphEndpoint": "https://graph.windows.net/",
   "storageEndpointSuffix": "redmond.ext-u15e0303.masd.stbtest.microsoft.com",
   "sQLDatabaseDNSSuffix": "",
   "trafficManagerDNSSuffix": "",
   "keyVaultDNSSuffix": "vault.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
   "serviceBusEndpointSuffix": "",
   "serviceManagementVMDNSSuffix": "cloudapp.net",
   "resourceManagerVMDNSSuffix": "cloudapp.azure.com",
   "containerRegistryDNSSuffix": "azurecr.io"

radhikagupta5:
When/How does the dev changes gets merged to master?

autorest/azure/environments.go
marstr: Righto, I understand the value that Azure Stack is bringing, and that "bring your own endpoint" is an inseparable concept.

What I was trying to say in my comment was that I don't think a global variable is the best way to expose this value. It is misleading to consumers because it doesn't start off hydrated with all of the values that you mention. They would have to call the function below which first hydrates the global variable, then returns it. I'd rather have a constructor for Environments that take a file with all of the endpoints and returns back an instantiated `Environment` that matches the Azure Stack instance you're targeting.

autorest/azure/environments.go
```diff
@@ -126,5 +135,17 @@ func EnvironmentFromName(name string) (Environment, error) {
 	if !ok {
 		return env, fmt.Errorf("autorest/azure: There is no cloud environment matching the name %q", name)
 	}
+
+	if strings.EqualFold(name, "AZURESTACKCLOUD") == true {
+		fbytes, err := ioutil.ReadFile(path.Join("/etc", strings.ToLower(name), "azurestackcloud.json"))
```
marstr: Ah, I'd rather see this getting passed in as parameter, with the `"/etc/<blah>/azurestackcloud.json"`  used as a default in some capacity.

marstr:
> I will submit a new PR for dev. When/How does the dev changes gets merged to master?

No need to submit a new PR! :) You should just be able to edit this PR to target dev.

We'll manually merge `dev` into master when we're ready to release so that folks get the same version of our code if they use `go get` or a package manager like [glide](https://github.com/Masterminds/glide).

autorest/azure/environments.go
radhikagupta5: What you are recommending for AzureStackCloud was my original thought as-well. Currently the issue with this approach is that, it would require changing upstream Kubernetes code (for my case though same logic applies to all cases) and they are not accepting any PR at this moment. This would serve as a decent intermediate solution before recommend solution can be implemented.

marstr:
It ended up being easier to submit from my own branch instead of overwriting changes in your branch. I've submitted a new PR, #187. I'm going to close this one in favor of that one.

PR #180, opened 2017-10-17
Updating tests and simplifying method.

marstr:
Adding a little polish before this goes into `master`.

### DIFF
--- a/autorest/azure/azure_test.go
+++ b/autorest/azure/azure_test.go
@@ -396,7 +396,7 @@ func TestWithErrorUnlessStatusCode_UnwrappedError(t *testing.T) {
 		t.Logf("`ServiceError` was nil when it shouldn't have been.")
 		t.Fail()
 	} else if azErr.ServiceError.Details == nil {
-		t.Logf("`ServiceError.Details` was nil when it shouldn't have been %q", expectedServiceErrorDetails)
+		t.Logf("`ServiceError.Details` was nil when it should have been %q", expectedServiceErrorDetails)
 		t.Fail()
 	} else if details, _ := json.Marshal(*azErr.ServiceError.Details); expectedServiceErrorDetails != string(details) {
 		t.Logf("Error detaisl was not unmarshaled properly.\n\tgot:  %q\n\twant: %q", string(details), expectedServiceErrorDetails)
--- a/autorest/azure/azure_test.go
+++ b/autorest/azure/azure_test.go
@@ -366,39 +366,48 @@ func TestWithErrorUnlessStatusCode_UnwrappedError(t *testing.T) {
 	err := autorest.Respond(r,
 		WithErrorUnlessStatusCode(http.StatusOK),
 		autorest.ByClosing())
+
 	if err == nil {
-		t.Logf("azure: returned nil error for proper error response")
-		t.Fail()
+		t.Fatal("azure: returned nil error for proper error response")
 	}
+
 	azErr, ok := err.(*RequestError)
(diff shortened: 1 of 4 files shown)


autorest/azure/azure_test.go
```diff
@@ -366,39 +366,48 @@ func TestWithErrorUnlessStatusCode_UnwrappedError(t *testing.T) {
...
 
-	details, _ := json.Marshal(*azErr.ServiceError.Details)
-	if expected := `[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]`; string(details) != expected {
-		t.Fatalf("azure: error details is not unmarshaled properly. Got=\"%s\"; Expected=\"%s\"", expected, string(details))
+	expectedServiceErrorDetails := `[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]`
+	if azErr.ServiceError == nil {
+		t.Logf("`ServiceError` was nil when it shouldn't have been.")
+		t.Fail()
+	} else if azErr.ServiceError.Details == nil {
+		t.Logf("`ServiceError.Details` was nil when it shouldn't have been %q", expectedServiceErrorDetails)
```
mcardosos: This error is not that understandable

autorest/azure/azure_test.go
```diff
@@ -366,39 +366,48 @@ func TestWithErrorUnlessStatusCode_UnwrappedError(t *testing.T) {
...
+		t.Logf("Incorrect request ID\n\tgot:  %q\n\twant: %q", azErr.RequestID, expected)
+		t.Fail()
 	}
 
-	details, _ := json.Marshal(*azErr.ServiceError.Details)
-	if expected := `[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]`; string(details) != expected {
-		t.Fatalf("azure: error details is not unmarshaled properly. Got=\"%s\"; Expected=\"%s\"", expected, string(details))
+	expectedServiceErrorDetails := `[{"code":"conflict1","message":"error message1"},{"code":"conflict2","message":"error message2"}]`
+	if azErr.ServiceError == nil {
+		t.Logf("`ServiceError` was nil when it shouldn't have been.")
```
mcardosos: What was expected?

autorest/azure/azure_test.go
```diff
@@ -366,39 +366,48 @@ func TestWithErrorUnlessStatusCode_UnwrappedError(t *testing.T) {
...
+		t.Fail()
+	} else if azErr.ServiceError.Details == nil {
+		t.Logf("`ServiceError.Details` was nil when it shouldn't have been %q", expectedServiceErrorDetails)
+		t.Fail()
+	} else if details, _ := json.Marshal(*azErr.ServiceError.Details); expectedServiceErrorDetails != string(details) {
```
mcardosos: If the test is about unmarshaling something, wouldn't it be better to compare the unmarshaled result to the expected result, instead of marshaling again?

autorest/azure/azure_test.go
marstr: This is what the crux of the test is around :) I'll no-fix this one

autorest/azure/azure_test.go
marstr: Fixed locally

autorest/azure/azure_test.go
marstr: Meh, it doesn't hurt to round trip on this :) I'll skip this one

Issue #181, opened 2017-10-19
Broke the cookie jar


mcardosos:
And ate the cookies.
A new client and cookie jar are created everytime the `autorest.Client` `Do()` implementation is executed. However, as the retries have been moved out of `Do()`, they no longer benefit from the cookie jar.
Some acceptance tests in https://github.com/Azure/autorest.go depend on cookies.

marstr:
Is this ready to close?

PR #182, opened 2017-10-19
Creating cookiejar while creating client

mcardosos:
Fixes autorest tests with cookies
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -175,6 +175,7 @@ func NewClientWithUserAgent(ua string) Client {
 		RetryDuration:   30 * time.Second,
 		UserAgent:       defaultUserAgent,
 	}
+	c.Sender = c.sender()
 	c.AddToUserAgent(ua)
 	return c
 }
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -204,20 +204,19 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		return nil, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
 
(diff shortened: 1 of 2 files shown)


marstr:
Just to be clear, this fixes #181?

mcardosos:
@marstr yes

mcardosos:
@vladbarosan :D

autorest/client.go
```diff
@@ -202,19 +203,21 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
-	resp, err := SendWithSender(c.sender(), r)
+
+	if c.Sender == nil {
+		c.Sender = sender()
+	}
+
+	resp, err := SendWithSender(c, r)
```
jhendrixMSFT: This should have been c.Sender not c and it the cause of the SO.

autorest/client.go
```diff
@@ -202,19 +203,21 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
...
+
+	resp, err := SendWithSender(c.Sender, r)
 	Respond(resp,
 		c.ByInspecting())
 	return resp, err
 }
 
 // sender returns the Sender to which to send requests.
-func (c Client) sender() Sender {
-	if c.Sender == nil {
```
jhendrixMSFT: I'm not convinced that removing this from here was a good choice as I now see this nil check sprinkled in various places.

PR #183, opened 2017-10-19
Updating to add AzureStack cloud.

radhikagupta5:
In AzureStack each stamp has a different environment and these values varies based on customer input. We don't know at the time of building this code what that information is. To keep this code independent of the environment we are changing to a model to bring your own endpoint. In this case we are reading the values from an existing json file. Yes, the json file should be present (in this case placed by ACS-Engine), but worse case we fail with an appropriate error.

For example, this is the value for an environment,

"name": "AzureStackCloud",
 "managementPortalURL": "",
 "publishSettingsURL": "",
 "serviceManagementEndpoint": "https://management.azurestackci13.onmicrosoft.com/e55c980f-9a38-48ab-a059-87027f8134b1",
 "resourceManagerEndpoint": "https://management.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
 "activeDirectoryEndpoint": "https://login.windows.net/",
 "galleryEndpoint": "https://portal.redmond.ext-u15e0303.masd.stbtest.microsoft.com:30015/",
 "keyVaultEndpoint": "https://vault.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
 "graphEndpoint": "https://graph.windows.net/",
 "storageEndpointSuffix": "redmond.ext-u15e0303.masd.stbtest.microsoft.com",
 "sQLDatabaseDNSSuffix": "",
 "trafficManagerDNSSuffix": "",
 "keyVaultDNSSuffix": "vault.redmond.ext-u15e0303.masd.stbtest.microsoft.com",
 "serviceBusEndpointSuffix": "",
 "serviceManagementVMDNSSuffix": "cloudapp.net",
 "resourceManagerVMDNSSuffix": "cloudapp.azure.com",
 "containerRegistryDNSSuffix": "azurecr.io"


marstr:
We can still use the original PR! Let's keep the other PR, #178. :)

PR #186, opened 2017-10-26
Add type Future used to track status of long-running operations.

jhendrixMSFT:
This change adds exported type Future which will be used by generated
response types for long-running operations.  It allows the caller to
track the status of a LRO and implement their own polling mechanism.
It builds on the existing and well-tested LRO infrastructure.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -16,6 +16,7 @@ package azure
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


autorest/azure/async.go
```diff
@@ -37,6 +38,80 @@ const (
...
+	resp *http.Response
+	ps   pollingState
+}
+
+// NewFuture returns a new Future object initialized with the specified request.
+func NewFuture(req *http.Request) Future {
+	return Future{req: req}
+}
+
+// Response returns the last HTTP response or nil if there isn't one.
```
marstr: What would your thoughts be on also saving a copy of the initial Response that confirmed the LRO was kicked successfully.

autorest/azure/async.go
```diff
@@ -37,6 +38,80 @@ const (
 	operationSucceeded  string = "Succeeded"
 )
 
+// Future provides a mechanism to access the status and results of an asynchronous request.
+// Since futures are stateful they should be passed by value to avoid race conditions.
```
marstr: This language still make me a little uneasy. I'd be more comfortable adding a Mutex of some sort. However, if you still feel strongly, I will defer.

autorest/azure/async.go
```diff
@@ -37,6 +38,80 @@ const (
...
+
+	f.req, err = newPollingRequest(f.ps)
+	return false, err
+}
+
+func (f *Future) terminalReturn() (bool, error) {
+	if !f.ps.hasSucceeded() {
+		return true, f.ps
+	}
+	return true, nil
```
marstr: I'd a comment here mentioning that this being `true` is intentional. It make sense if you read this in combination with `Done` but otherwise it could be confusing.

autorest/azure/async_test.go
```diff
@@ -1024,6 +1015,77 @@ func TestDoPollForAsynchronous_StopsPollingIfItReceivesAnInvalidOperationResourc
...
+	r1.Header.Del(http.CanonicalHeaderKey(headerAsyncOperation))
+	r2 := newProvisioningStatusResponse("busy")
+	r2.Header.Del(http.CanonicalHeaderKey(headerAsyncOperation))
+	r3 := newProvisioningStatusResponse(operationSucceeded)
+	r3.Header.Del(http.CanonicalHeaderKey(headerAsyncOperation))
+
+	client := mocks.NewSender()
+	client.AppendResponse(r1)
+	client.AppendAndRepeatResponse(r2, 2)
+	client.AppendAndRepeatResponse(r3, 1)
```
mcardosos: Nit, AppendResponse

autorest/azure/async.go
jhendrixMSFT: That response will be saved.  The generated code will make the initial call to Done() before returning the future so Response() will return the initial response.

autorest/azure/async.go
jhendrixMSFT: Good point I'll add a comment.

autorest/azure/async_test.go
jhendrixMSFT: Will fix (this was copy/pasted from another method).

autorest/azure/async.go
jhendrixMSFT: If we do that then it makes everybody pay that tax (albeit a small one).  Keeping it as-is leaves it up to the consumer whether or not they need this functionality which is why documenting that it's stateful is important.

jhendrixMSFT: I can reword this though and am open to suggestions.

PR #189, opened 2017-10-30
Add method GetPollingDelay to Future.

jhendrixMSFT:
Some long-running operations will return a Retry-After response header
indicating how long a client should wait before checking the status.
Add a method exposing this information when present.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -90,6 +90,28 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 	return false, err
 }
 
+// GetPollingDelay returns a duration the application should wait before checking
+// the status of the asynchronous request and true; this value is returned from
+// the service via the Retry-After response header.  If the header wasn't returned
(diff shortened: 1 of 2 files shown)


autorest/azure/async.go
```diff
@@ -90,6 +90,27 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 	return false, err
 }
 
+// RetryAfter returns a duration the application should wait before checking
+// the status of the asynchronous request and true.  If a duration wasn't
+// specified then the function returns the zero-value time.Duration and false.
+func (f Future) RetryAfter() (time.Duration, bool) {
```
mcardosos: >RetryAfter [](start = 16, length = 10)

Not the biggest fan of this name, (it sounds as if the retry happened inside), maybe GetWaitTime?

jhendrixMSFT: Agreed, I'll fix up the name.

jhendrixMSFT: Fixed.

Issue #193, opened 2017-11-07
Recent Refactoring to Long Running Operations


tombuildsstuff:
👋 

I've seen [this recent PR](https://github.com/Azure/autorest.go/pull/39) which changes the way that Long Running Operations are handled, such that it effectively removes the polling support from the SDK and instead pushed that onto consumers of the SDK.

Within Terraform, we rely pretty heavily on the automatic polling available within the Go SDK - as such I'm concerned that removing these methods will lead to a considerable increase in code (and thus complexity) for us - when currently this is one of the major benefits of using the Go SDK over when we used to use the Riviera SDK.

As such - are there any plans around to expose a polling method available within the SDK to preserve the current behaviour? I can certainly see the desire for the SDK to return the current state - however polling until the resource is ready is also a really handy function of the SDK at this point :)

Thanks!

jhendrixMSFT:
This was something I was thinking about last week.  Adding a method to the future that waits until completion would be pretty trivial, @tombuildsstuff would this work for you?

tombuildsstuff:
@jhendrixMSFT yep, that would be great 👍

jhendrixMSFT:
This has been added in go-autorest v9.4.0.

PR #194, opened 2017-11-07
Add WaitForCompletion to Future type.

jhendrixMSFT:
The WaitForCompletion method provides a default implementation that will
wait until a future has completed, is cancelled or times out.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -125,12 +125,11 @@ func (f Future) GetPollingDelay() (time.Duration, bool) {
 // polling duration has been exceeded.  It will retry failed polling attempts based on
 // the retry value defined in the client up to the maximum retry attempts.
 func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) error {
-	start := time.Now()
+	ctx, cancel := context.WithTimeout(ctx, client.PollingDuration)
+	defer cancel()
+
 	done, err := f.Done(client)
 	for attempts := 0; !done; done, err = f.Done(client) {
-		if time.Since(start) > client.PollingDuration {
-			return autorest.NewErrorWithResponse("azure", "WaitForCompletion", f.resp, "the polling duration has been exceeded")
-		}
 		if attempts >= client.RetryAttempts {
 			return autorest.NewErrorWithError(err, "azure", "WaitForCompletion", f.resp, "the number of retries has been exceeded")
 		}
@@ -156,7 +155,7 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
 		// wait until the delay elapses or the context is cancelled
 		delayElapsed := autorest.DelayForBackoff(delay, delayAttempt, ctx.Done())
 		if !delayElapsed {
-			return ctx.Err()
+			return autorest.NewErrorWithError(ctx.Err(), "azure", "WaitForCompletion", f.resp, "context has been cancelled")
 		}
 	}
 	return err
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -16,6 +16,7 @@ package azure
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -37,6 +38,8 @@ const (
 	operationFailed     string = "Failed"
 	operationSucceeded  string = "Succeeded"
 )
+
+var pollingCodes = [...]int{http.StatusAccepted, http.StatusCreated, http.StatusOK}
 
 // Future provides a mechanism to access the status and results of an asynchronous request.
 // Since futures are stateful they should be passed by value to avoid race conditions.
@@ -78,7 +81,7 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 
 	resp, err := sender.Do(f.req)
 	f.resp = resp
-	if err != nil {
+	if err != nil || !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
 		return false, err
 	}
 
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
Fix for issue https://github.com/Azure/go-autorest/issues/193.

jhendrixMSFT:
@tombuildsstuff With this change one can write the following code.
```go
future, _ := client.Create(context.Background(), "testrg", "testcache", params)
// the request was successfully sent, now wait for it to complete
err = future.WaitForCompletion(context.Background(), client.Client)
// the operation is complete, retrieve the result
res, err := future.Result(client)
```
What do you think?

autorest/azure/async.go
```diff
@@ -117,6 +119,45 @@ func (f Future) GetPollingDelay() (time.Duration, bool) {
...
 }
 
+// WaitForCompletion will return when one of the following conditions is met: the long
+// running operation has completed, the provided context is cancelled, or the client's
+// polling duration has been exceeded.  It will retry failed polling attempts based on
+// the retry value defined in the client up to the maximum retry attempts.
+func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) error {
+	start := time.Now()
+	done, err := f.Done(client)
+	for attempts := 0; !done && attempts < client.RetryAttempts; done, err = f.Done(client) {
```
mcardosos: > attempts < client.RetryAttempts [](start = 28, length = 32)

Is returning an error in case `attempts > RetryAttempts` useful?

jhendrixMSFT: ```err``` should already be non-nil in this case.  Or do you mean wrap it in an outer error?

autorest/azure/async_test.go
```diff
@@ -1096,6 +1097,42 @@ func TestFuture_Marshalling(t *testing.T) {
...
+	sender.AppendAndRepeatResponse(r2, 2)
+	sender.AppendResponse(r3)
+
+	future := NewFuture(mocks.NewRequest())
+
+	client := autorest.Client{
+		PollingDelay:    autorest.DefaultPollingDelay,
+		PollingDuration: autorest.DefaultPollingDuration,
+		RetryAttempts:   autorest.DefaultRetryAttempts,
+		RetryDuration:   30 * time.Second,
```
mcardosos: >RetryDuration:   30 * time.Second, [](start = 2, length = 34)

How much time does this test take?

autorest/azure/async.go
mcardosos: Yes I meant adding extra info on already done retries.

autorest/azure/async_test.go
jhendrixMSFT: It is fast, but now you got me thinking it's not comprehensive enough, i.e. it doesn't test the retry-due-to-failure path.  I will update the test.

autorest/azure/async.go
jhendrixMSFT: Fixed.

autorest/azure/async_test.go
jhendrixMSFT: Fixed.

mcardosos:
We might still be able to support context package with build tags. Context used to live in a different [import path](https://golang.org/doc/go1.7#context).

autorest/azure/async.go
```diff
@@ -38,6 +39,8 @@ const (
 	operationSucceeded  string = "Succeeded"
 )
 
+var pollingCodes = [...]int{http.StatusAccepted, http.StatusCreated, http.StatusOK}
```
tombuildsstuff: 🤔 is it also worth retrying/polling for `500 InternalServerError` and `429 Too Many Requests` - or is that handled elsewhere?

tombuildsstuff:
@jhendrixMSFT it looks good - my only comment is that it'd be nice to re-use the Client and the Context used in the Create if possible (something like this, maybe?):

```
future, _ := client.Create(context.Background(), "testrg", "testcache", params)
// the request was successfully sent, now wait for it to complete
err = future.WaitForCompletion()
// the operation is complete, retrieve the result
res, err := future.Result(client)
```

but it otherwise LGTM - thanks for this 👍

jhendrixMSFT:
@tombuildsstuff In my original design for Future it worked just like this, i.e. the Future held on to the associated client and you didn't have to pass it in to Done() or Result().  Unfortunately this got a little clunky when unmarshalling a Future, you couldn't just use the unmarshalled Future without first setting the exported client field making this scenario non-intuitive.  I will revisit this before release v12 to see if there's a better way to handle this.

autorest/azure/async.go
jhendrixMSFT: We will still retry for those error codes (i.e. Done() will return false).  In this context these codes are used to determine if we should update the internal polling status or not.

autorest/azure/async_test.go
```diff
@@ -1096,6 +1098,129 @@ func TestFuture_Marshalling(t *testing.T) {
...
+		RetryDuration:   1 * time.Second,
+		Sender:          sender,
+	}
+
+	err := future.WaitForCompletion(context.Background(), client)
+	if err != nil {
+		t.Fatalf("azure: WaitForCompletion returned non-nil error")
+	}
+
+	if sender.Attempts() < 6 {
```
marstr: Instead of `6`, could we use `client.RetryAttempts` or `autorest.DefaultRetryAttempts`?

autorest/azure/async.go
```diff
@@ -152,8 +196,7 @@ func DoPollForAsynchronous(delay time.Duration) autorest.SendDecorator {
 			if err != nil {
 				return resp, err
 			}
-			pollingCodes := []int{http.StatusAccepted, http.StatusCreated, http.StatusOK}
-			if !autorest.ResponseHasStatusCode(resp, pollingCodes...) {
+			if !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
```
marstr: I'm sure you had a reason for changing this line, and it makes me super curious. [From this little sample](https://play.golang.org/p/GbQJd07wwG), I can't see any difference. What did you run into?

autorest/azure/async.go
```diff
@@ -117,6 +120,47 @@ func (f Future) GetPollingDelay() (time.Duration, bool) {
 	return d, true
 }
 
+// WaitForCompletion will return when one of the following conditions is met: the long
+// running operation has completed, the provided context is cancelled, or the client's
+// polling duration has been exceeded.  It will retry failed polling attempts based on
+// the retry value defined in the client up to the maximum retry attempts.
+func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) error {
+	ctx, cancel := context.WithTimeout(ctx, client.PollingDuration)
```
marstr: Nice, this looks a lot better to me. Thanks!

autorest/azure/async.go
```diff
@@ -117,6 +120,47 @@ func (f Future) GetPollingDelay() (time.Duration, bool) {
...
+		} else {
+			// there was an error polling for status so perform exponential
+			// back-off based on the number of attempts using the client's retry
+			// duration.  update attempts after delayAttempt to avoid off-by-one.
+			delayAttempt = attempts
+			delay = client.RetryDuration
+			attempts++
+		}
+		// wait until the delay elapses or the context is cancelled
+		delayElapsed := autorest.DelayForBackoff(delay, delayAttempt, ctx.Done())
```
marstr: optional, if we see that `ctx.Deadline()` will come before the backoff delay would complete we could fail a little faster.

autorest/azure/async.go
jhendrixMSFT: It's because pollingCodes is an array not a slice (it tripped me up at first too).

autorest/azure/async_test.go
jhendrixMSFT: The number on the right is the number of mock responses.  Let me update the mock sender to track the number of responses added so we don't have to guess like this (and forget to update it).

autorest/azure/async.go
jhendrixMSFT: In this case wouldn't the context get cancelled and DelayForBackoff() return false as soon as the timeout expires?

autorest/azure/async.go
jhendrixMSFT: ~0~

jhendrixMSFT: Sigh that was supposed to be noodle arms.

Issue #195, opened 2017-11-13
Polling should be case-insensitive


tombuildsstuff:
:wave:

We're provisioning a Search Service through the Azure SDK for Go - and have noticed since ~mid July that the deployment times have been increasing.

Upon further investigation - it appears that AutoRest doesn't handle the `provisioningState` field being lower-case - as is the case in the Search Service API:

After sending this request:
```
{
	"location": "westeurope",
	"tags": {
		"database": "test",
		"environment": "staging"
	},
	"properties": {},
	"sku": {
		"name": "standard"
	}
}
```

and polling:

```
{
	"id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tharvey-dev2/providers/Microsoft.Search/searchServices/tharveysearch",
	"name": "tharveysearch",
	"type": "Microsoft.Search/searchServices",
	"location": "West Europe",
	"tags": {
		"database": "test",
		"environment": "staging"
	},
	"properties": {
		"replicaCount": 1,
		"partitionCount": 1,
		"status": "provisioning",
		"statusDetails": "",
		"provisioningState": "provisioning",
		"hostingMode": "Default"
	},
	"sku": {
		"name": "standard"
	}
}
```

which, eventually returns the following:

```
{
	"id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tharvey-dev2/providers/Microsoft.Search/searchServices/tharveysearch",
	"name": "tharveysearch",
	"type": "Microsoft.Search/searchServices",
	"location": "West Europe",
	"tags": {
		"database": "test",
		"environment": "staging"
	},
	"properties": {
		"replicaCount": 1,
		"partitionCount": 1,
		"status": "running",
		"statusDetails": "",
		"provisioningState": "succeeded",
		"hostingMode": "Default"
	},
	"sku": {
		"name": "standard"
	}
}
```

However the SDK continues polling indefinitely (see below, the first `successful` response is highlighted):

<img width="314" alt="screen shot 2017-11-13 at 15 44 51" src="https://user-images.githubusercontent.com/666005/32734412-dbe54aae-c889-11e7-9411-191354bbbc75.png">

Upon updating [this line](https://github.com/Azure/go-autorest/blob/master/autorest/azure/async.go#L39) to be lower-case and rebuilding - the SDK appears to work as expected.

As such - would it be possible to update AutoRest to allow this field to be case insensitive?

[...]

PR #201, opened 2017-11-17
Use case-insensitive string comparison for polling state.

jhendrixMSFT:
We have seen a case where the polling state string is all lower case
which caused an infinite loop while polling.


jhendrixMSFT:
Fixes issue #https://github.com/Azure/go-autorest/issues/195.

PR #204, opened 2017-11-22
Fix hangs during failed authentication.

jhendrixMSFT:
Validate credential parameters.
Don't retry on http.StatusUnauthorized.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -54,6 +54,12 @@ type OAuthTokenProvider interface {
 	OAuthToken() string
 }
 
+// TokenRefreshError is an interface used by errors returned during token refresh.
+type TokenRefreshError interface {
+	error
+	Response() *http.Response
+}
+
 // Refresher is an interface for token refresh functionality
 type Refresher interface {
 	Refresh() error
@@ -76,6 +82,17 @@ type Token struct {
 
 	Resource string `json:"resource"`
 	Type     string `json:"token_type"`
+}
+
+// IsZero returns true if the token object is zero-initialized.
+func (t Token) IsZero() bool {
+	return len(t.AccessToken) == 0 &&
+		len(t.RefreshToken) == 0 &&
+		len(t.ExpiresIn) == 0 &&
+		len(t.ExpiresOn) == 0 &&
+		len(t.NotBefore) == 0 &&
+		len(t.Resource) == 0 &&
+		len(t.Type) == 0
 }
 
 // Expires returns the time.Time when the Token expires.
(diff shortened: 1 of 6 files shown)


jhendrixMSFT:
Fix for issue https://github.com/Azure/azure-sdk-for-go/issues/870.

autorest/client.go
```diff
@@ -204,7 +204,13 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		c.WithInspection(),
 		c.WithAuthorization())
 	if err != nil {
-		return nil, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
+		var resp *http.Response
+		if detErr, ok := err.(DetailedError); ok {
+			// if the authorization failed (e.g. invalid credentials) there will
```
mcardosos: What happens if the error in not a `DetailedError`

autorest/adal/token.go
```diff
@@ -78,6 +84,17 @@ type Token struct {
 	Type     string `json:"token_type"`
 }
 
+// IsZero returns true if the token object is zero-initialized.
+func (t Token) IsZero() bool {
+	return len(t.AccessToken) == 0 &&
```
mcardosos: For curiosity, why not just compare to an empty struct?

autorest/adal/token.go
```diff
@@ -258,6 +318,15 @@ func NewServicePrincipalToken(oauthConfig OAuthConfig, clientID string, secret s
 
 // NewServicePrincipalTokenFromCertificate create a ServicePrincipalToken from the supplied pkcs12 bytes.
 func NewServicePrincipalTokenFromCertificate(oauthConfig OAuthConfig, clientID string, certificate *x509.Certificate, privateKey *rsa.PrivateKey, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
```
mcardosos: >certificate *x509.Certificate, privateKey *rsa.PrivateKey [](start = 87, length = 57)

`certificate` and `privateKey` can be nil without problem?
Since this func is `NewServicePrincipalTokenFromCertificate` I would guess at least cert should not be nil.

autorest/client.go
jhendrixMSFT: It will just behave as it used to.

autorest/adal/token.go
jhendrixMSFT: That would work, however I think it feels a bit unnatural, i.e. I'd have to create some temporary Token first then compare the two.

autorest/adal/token.go
jhendrixMSFT: Good catch, fixed.

autorest/adal/token.go
marstr: I'm kinda with @mcardosos on this one, implementing like this: https://play.golang.org/p/bH2OYqg6yv
would be durable to future elements being added to the Token struct, and would be a little easier to read IMHO.

jhendrixMSFT: How about we make the guts of IsZero() use this comparison then?

Issue #207, opened 2017-11-30
API Calls don't support slice parameters which have []string as the underlying type


vladbarosan:

vladbarosan:
This is blocking proper support for Cognitive Support

PR #208, opened 2017-11-30
Add support for API calls with non string slice parameters

vladbarosan:

### DIFF
--- a/autorest/utility.go
+++ b/autorest/utility.go
@@ -138,33 +138,38 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
-// ToStringSlice method converts interface{} to []string.
-func ToStringSlice(s interface{}) []string {
-
+// AsStringSlice method converts interface{} to []string. This expects a
+//that the parameter passed to be a slice or array of a type that has the underlying
+//type a string.
+func AsStringSlice(s interface{}) ([]string, error) {
 	v := reflect.ValueOf(s)
-	stringSlice := make([]string, v.Len())
+	if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
+		return nil, NewError("autorest", "AsStringSlice", "the value's type is not an array.")
+	}
+	stringSlice := make([]string, 0, v.Len())
 
 	for i := 0; i < v.Len(); i++ {
-		stringSlice[i] = v.Index(i).String()
+		stringSlice = append(stringSlice, v.Index(i).String())
 	}
-
-	return stringSlice
+	return stringSlice, nil
 }
 
 // String method converts interface v to string. If interface is a list, it
-// joins list elements using separator.
+// joins list elements using the seperator. Note that only sep[0] will be used for
+// joining if any separator is specified.
 func String(v interface{}, sep ...string) string {
-	if len(sep) > 0 {
-		stringSlice, ok := v.([]string)
-
-		if ok == false {
-			stringSlice = ToStringSlice(v)
-		}
-		stringValue := strings.Join(stringSlice, strings.Join(sep, ""))
-		return ensureValueString(stringValue)
-
+	if len(sep) == 0 {
+		return ensureValueString(v)
 	}
-	return ensureValueString(v)
+	stringSlice, ok := v.([]string)
+	if ok == false {
+		var err error
+		stringSlice, err = AsStringSlice(v)
+		if err != nil {
+			panic(fmt.Sprintf("autorest: Couldn't convert value to a string %s.", err))
+		}
+	}
+	return ensureValueString(strings.Join(stringSlice, sep[0]))
 }
 
 // Encode method encodes url path and query parameters.
--- a/autorest/utility.go
+++ b/autorest/utility.go
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
+// ToStringSlice method converts interface{} to []string.
+func ToStringSlice(s interface{}) []string {
+
+	v := reflect.ValueOf(s)
+	stringSlice := make([]string, v.Len())
+
+	for i := 0; i < v.Len(); i++ {
+		stringSlice[i] = v.Index(i).String()
+	}
+
+	return stringSlice
+}
+
 // String method converts interface v to string. If interface is a list, it
 // joins list elements using separator.
 func String(v interface{}, sep ...string) string {
 	if len(sep) > 0 {
-		return ensureValueString(strings.Join(v.([]string), sep[0]))
+		stringSlice, ok := v.([]string)
+
+		if ok == false {
+			stringSlice = ToStringSlice(v)
+		}
+		stringValue := strings.Join(stringSlice, strings.Join(sep, ""))
+		return ensureValueString(stringValue)
+
 	}
 	return ensureValueString(v)
 }
--- a/autorest/utility_test.go
+++ b/autorest/utility_test.go
@@ -229,9 +229,49 @@ func ExampleString() {
 
 func TestStringWithValidString(t *testing.T) {
 	i := 123
-	if String(i) != "123" {
-		t.Fatal("autorest: String method failed to convert integer 123 to string")
+	if got, want := String(i), "123"; got != want {
+		t.Logf("got:  %q\nwant: %q", got, want)
+		t.Fail()
 	}
+}
+
+func TestStringWithStringSlice(t *testing.T) {
+	s := []string{"string1", "string2"}
+	if got, want := String(s, ","), "string1,string2"; got != want {
+		t.Logf("got:  %q\nwant: %q", got, want)
+		t.Fail()
+	}
+}
+
+func TestStringWithEnum(t *testing.T) {
+	type TestEnumType string
+	s := TestEnumType("string1")
+	if got, want := String(s), "string1"; got != want {
+		t.Logf("got:  %q\nwant: %q", got, want)
+		t.Fail()
+	}
+}
+
+func TestStringWithEnumSlice(t *testing.T) {
+	type TestEnumType string
+	s := []TestEnumType{"string1", "string2"}
+	if got, want := String(s, ","), "string1,string2"; got != want {
+		t.Logf("got:  %q\nwant: %q", got, want)
+		t.Fail()
+	}
+}
+
+func ExampleAsStringSlice() {
+	type TestEnumType string
+
+	a := []TestEnumType{"value1", "value2"}
+	b, _ := AsStringSlice(a)
+	for _, c := range b {
+		fmt.Println(c)
+	}
+	// Output:
+	// value1
+	// value2
 }
 
 func TestEncodeWithValidPath(t *testing.T) {


vladbarosan:
This addresses #207

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
...
+	for i := 0; i < v.Len(); i++ {
+		stringSlice[i] = v.Index(i).String()
+	}
+
+	return stringSlice
+}
+
 // String method converts interface v to string. If interface is a list, it
 // joins list elements using separator.
 func String(v interface{}, sep ...string) string {
```
mcardosos: >String [](start = 5, length = 6)

I would add a test for when the array is made with enums.

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
...
 	if len(sep) > 0 {
-		return ensureValueString(strings.Join(v.([]string), sep[0]))
+		stringSlice, ok := v.([]string)
+
+		if ok == false {
+			stringSlice = ToStringSlice(v)
+		}
+		stringValue := strings.Join(stringSlice, strings.Join(sep, ""))
+		return ensureValueString(stringValue)
+
```
mcardosos: Nit, newline

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
+// ToStringSlice method converts interface{} to []string.
+func ToStringSlice(s interface{}) []string {
+
```
mcardosos: Nit, newline

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
+// ToStringSlice method converts interface{} to []string.
```
marstr: There are qualifiers on this, I'd update the comment to make clear that the `interface{}` must have a particular shape. I.e. it must be a slice of a type derived from string.

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
...
+func ToStringSlice(s interface{}) []string {
+
+	v := reflect.ValueOf(s)
```
marstr: Please add a test here to make sure that `v.Kind()` is a `reflect.Slice` or a `reflect.Array`.

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
...
+	stringSlice := make([]string, v.Len())
+
+	for i := 0; i < v.Len(); i++ {
+		stringSlice[i] = v.Index(i).String()
+	}
+
+	return stringSlice
```
marstr: [nit] the more conventional pattern would be to write the above as follows :)

``` Go
stringSlice := make([]string, 0, v.Len())

for i := 0; i < v.Len(); i++ {
    stringSlice =append(stringSlice, v.Index(i).String())
}

return stringSlice
```

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
+// ToStringSlice method converts interface{} to []string.
+func ToStringSlice(s interface{}) []string {
```
marstr: I think this function should return an error when `s` isn't something convertible to a string slice.

autorest/utility.go
```diff
@@ -138,11 +138,31 @@ func MapToValues(m map[string]interface{}) url.Values {
...
 // joins list elements using separator.
 func String(v interface{}, sep ...string) string {
 	if len(sep) > 0 {
-		return ensureValueString(strings.Join(v.([]string), sep[0]))
+		stringSlice, ok := v.([]string)
+
+		if ok == false {
+			stringSlice = ToStringSlice(v)
+		}
+		stringValue := strings.Join(stringSlice, strings.Join(sep, ""))
```
marstr: Instead of joining all of the separators, I'd rather panic if somebody provides more than one. Honestly, this method probably needs to be deprecated and replaced soon. It is very gross.

autorest/utility.go
vladbarosan: Agreed, kind of sucks that all the functions there don't follow this pattern. I created a proposal here for improving #211

autorest/utility.go
vladbarosan: #211 . I would leave it with them joined until we improve the whole thing here in some future breaking change. if someone provides multiple separators the expected behavior would be to use all of them. they might see it as instead of doing `":;"` they can do `":", ";"`

autorest/utility.go
vladbarosan: wanted to but, saw in the file its not like that so naturally assumed that's what we consider more conventional :)

autorest/utility.go
vladbarosan: would add also a test for when there is an array, apparently right only when a string is passed is  being tested

autorest/utility.go
```diff
@@ -138,13 +138,38 @@ func MapToValues(m map[string]interface{}) url.Values {
...
+	stringSlice, ok := v.([]string)
+	if ok == false {
+		var err error
+		stringSlice, err = AsStringSlice(v)
+		if err != nil {
+			panic(fmt.Sprintf("autorest: Couldn't convert value to a string %s.", err))
+		}
 	}
-	return ensureValueString(v)
+	stringValue := strings.Join(stringSlice, strings.Join(sep, ""))
```
jhendrixMSFT: The original implementation used sep[0] so it seems to me this change will introduce new behavior.  Is that intentional?

vladbarosan: It is intentional yes. I would say it introduces proper behavior (i.e. fixes a bug). e.g. if someone passed previously multiple separators only the first one would be used when the intent would be clearly to use multiple separators.

jhendrixMSFT: Are there any cases in go-autorest where more than one separator is specified?  I suspect the reason the method has this sig is to treat the sep param as optional, not because they want more than one separator.

vladbarosan: That might have been the intent. In our case, the only use by us is coming from generated code for parameters, namely here https://github.com/Azure/autorest.go/blob/395cbeb9664982d64b89581d9a4b10c7ba427a26/src/Extensions.cs#L226

But the issue we have I assume if it is used by client code ( since its publicly exposed). Im thinking as a client, if I see method that takes multiple parameters as seperator, do I want it to actually do that or get a comment saying that oh actually only the first one is used and the rest will be ignored.

vladbarosan: actually given that is a join and not a split, doesn't really make sense to have more, so will leave it like it is to avoid any unforeseen consequences.

autorest/utility_test.go
```diff
@@ -234,6 +234,29 @@ func TestStringWithValidString(t *testing.T) {
...
 }
 
+func TestStringWithStringSlice(t *testing.T) {
+	s := []string{"string1", "string2"}
+	if String(s, ",") != "string1,string2" {
+		t.Fatal("autorest: String method failed to convert string slice to string")
+	}
+}
+
+type TestEnumType string
```
marstr: [nit] I'd just declare this type inside of the tests that use it as to not pollute the namespace.

autorest/utility_test.go
```diff
@@ -234,6 +234,29 @@ func TestStringWithValidString(t *testing.T) {
...
+
+func TestStringWithEnum(t *testing.T) {
+	s := TestEnumType("string1")
+	if String(s) != "string1" {
+		t.Fatal("autorest: String method failed to convert slice of enums to string")
+	}
+}
+
```
marstr: In addition to the tests that are here, I'd add an `Example` for the `AsStringSlice` to try to make more clear what it does.

I think this is a really natural candidate for an `Example` function instead of a `test` function. Consider this:

``` Go
func ExampleAsStringSlice() {
    type TestEnumType string

    a := []TestEnumType{"value1", "value2"}
    b, err := AsStringSlice(s)
    for _, c := range b {
        fmt.Println(c)
    }
    // Output:
    // value1
    // value2
}
```

autorest/utility.go
```diff
@@ -138,13 +138,38 @@ func MapToValues(m map[string]interface{}) url.Values {
...
-		return ensureValueString(strings.Join(v.([]string), sep[0]))
+	if len(sep) == 0 {
+		return ensureValueString(v)
+	}
+	stringSlice, ok := v.([]string)
+	if ok == false {
+		var err error
+		stringSlice, err = AsStringSlice(v)
+		if err != nil {
+			panic(fmt.Sprintf("autorest: Couldn't convert value to a string %s.", err))
```
marstr: If our hands are tied because this is an existing method I understand, but it really does bother me that we're stuck using a panic in this situation. It may be worth adding a work item to the next major release milestone of `go-autorest` to go fix this up.

autorest/utility.go
```diff
@@ -138,13 +138,38 @@ func MapToValues(m map[string]interface{}) url.Values {
...
+		}
 	}
-	return ensureValueString(v)
+	return ensureValueString(strings.Join(stringSlice, sep[0]))
```
marstr: In the work item to be created (mentioned above), I'd probably add a note that we should clean up this separator non-sense as well.

autorest/utility_test.go
```diff
@@ -234,6 +234,29 @@ func TestStringWithValidString(t *testing.T) {
 	}
 }
 
+func TestStringWithStringSlice(t *testing.T) {
+	s := []string{"string1", "string2"}
+	if String(s, ",") != "string1,string2" {
+		t.Fatal("autorest: String method failed to convert string slice to string")
```
marstr: [nit] the more go-ish way of expressing this would be to do something like the following:

``` Go
func TestStringWithStringSlice(t *testing.T) {
    s := []string{"string1", "string2"}
    if got, want := String(s, ","), "string1,string2"; got != want {
        t.Logf("got:  %q\nwant: %q", got, want)
        t.Fail()
    }
}
```

autorest/utility_test.go
```diff
@@ -234,6 +234,29 @@ func TestStringWithValidString(t *testing.T) {
...
+	}
+}
+
+type TestEnumType string
+
+func TestStringWithEnum(t *testing.T) {
```
marstr: [nit] You may consider using [the compound test syntax](https://golang.org/pkg/testing/#T.Run) for these three pretty similar tests.

autorest/utility.go
thesunwave: Hey, sorry for the late question but, why is this function accepts an interface, then cast elements of array/slice through `.String()`? I'll explain why I'm asking, for example, here https://github.com/Azure/azure-sdk-for-go/blob/master/services/cognitiveservices/v1.0/face/largefacelist.go#L288-L290, this function accepts several arguments one of them is `targetFace []int32`, then `autorest.Encode` called with this `[]int`, so as to result the query param `targetFace` looks like `<int32 Value> <int32 Value> <int32 Value> <int32 Value>`, because of `v.Index(i).String()`.

I've created an issue for `azure-sdk-for-go`, but also I've fixed it in my fork but didn't make a pull request. You can see that here https://github.com/thesunwave/go-autorest/commit/f36f77b4f073a6f651955e977a4ccfb01e697e38, what do you think, is it a valid solution?

Issue #214, opened 2017-12-07
Receiving too many 429s will eventually fail retries


jhendrixMSFT:
In DoRetryForStatusCodes() we will honor the Retry-After header (if available) however it will count as a retry and once the cap is reached the retry will abort.  One idea is to not count 429s against the total count of retries, needs investigation.

jhendrixMSFT:
Fixed in v9.5.1

PR #215, opened 2017-12-08
Don't count 429 as a retry attempt.

jhendrixMSFT:
If an API returns a 429 don't include it in the retry count so that we
can keep retrying until the operation succeeds.
Don't skip retry logic when SkipResourceProviderRegistration is enabled.

### DIFF
--- a/autorest/azure/rp.go
+++ b/autorest/azure/rp.go
@@ -30,9 +30,6 @@ import (
 func DoRetryWithRegistration(client autorest.Client) autorest.SendDecorator {
 	return func(s autorest.Sender) autorest.Sender {
 		return autorest.SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
-			if client.SkipResourceProviderRegistration {
-				return autorest.SendWithSender(s, r)
-			}
 			rr := autorest.NewRetriableRequest(r)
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
I updated this PR to fix a regression I introduced (SkipResourceProviderRegistration), can you guys please take another look.  @vladbarosan @marstr

autorest/azure/rp.go
```diff
@@ -60,7 +57,7 @@ func DoRetryWithRegistration(client autorest.Client) autorest.SendDecorator {
 				}
 				err = re
 
-				if re.ServiceError != nil && re.ServiceError.Code == "MissingSubscriptionRegistration" {
+				if re.ServiceError != nil && re.ServiceError.Code == "MissingSubscriptionRegistration" && !client.SkipResourceProviderRegistration {
```
vladbarosan: > && !client.SkipResourceProviderRegistration  [](start = 90, length = 45)

is there something we cna add to the `rp_test.go` for this scenario ?

jhendrixMSFT: Good call, and adding a test made me realize the fix is wrong.  :/

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/214

Issue #202, opened 2017-11-20
Add Azure auth method with default behavior logic


vladbarosan:
Offer opinionated method for obtaining an Authorizer which tries in order: 1) Service Principal from consts and/or env vars 2) Managed Service Identity (MSI)

vladbarosan:
linking to the enhancement of this https://github.com/Azure/azure-sdk-for-go/issues/881

PR #216, opened 2017-12-08
Fix bugs blocking Cognitive Services

vladbarosan:
- Fix bug https://github.com/Azure/go-autorest/issues/220
- Fix bug https://github.com/Azure/go-autorest/issues/219 


vladbarosan:
Addresses https://github.com/Azure/go-autorest/issues/202

vladbarosan:
We want to get this in for the SDK v12 release. Otherwise I can move this to dev branch.

marstr:
Because this is additive, I don't think we need to stress about what the timing is on merging these changes. Even if we release 12 before this, it'll be trivial to ship/adopt these changes.

jhendrixMSFT:
As a general comment I think this functionality should go into a more appropriately named package, perhaps an "auth" package under the utils directory?  Having a generic utils package is kind of an anti-pattern and I'd hate for it to end up as a random collection of unrelated bits.

vladbarosan:
@jhendrixMSFT Agreed, was thinking of leaving it there is there is already a `GetAuthorizer` method and move it next breaking change. But we can move it now and just mark the one already in `utils` as deprecated.

autorest/utils/auth.go
```diff
@@ -26,6 +26,9 @@ import (
 // GetAuthorizer gets an Azure Service Principal authorizer.
 // This func assumes "AZURE_TENANT_ID", "AZURE_CLIENT_ID",
 // "AZURE_CLIENT_SECRET" are set as environment variables.
+//
+// Deprecated: Use GetAuthorizerFromClientCredentials from the
```
vladbarosan: Marking this as `Deprecated` following Go's convention (https://blog.golang.org/godoc-documenting-go-code)

vladbarosan: @jhendrixMSFT @marstr . Should we document the deprecation in the changelog ?

vladbarosan:
Also addresses #219

autorest/preparer.go
```diff
@@ -452,11 +452,16 @@ func WithQueryParameters(queryParameters map[string]interface{}) PrepareDecorato
...
 				v := r.URL.Query()
 				for key, value := range parameters {
-					v.Add(key, value)
+					d, err := url.QueryUnescape(value)
+					if err != nil {
+						return r, err
+					}
+					v.Add(key, d)
 				}
-				r.URL.RawQuery = createQuery(v)
```
jhendrixMSFT: Is the createQuery() func still needed?  If not I think we should delete it.

autorest/utils/auth.go
jhendrixMSFT: Good question, it seems like it would be good to have a "Deprecated" section similar to how we have "New Features" and "Bugs" so that it's clear to folks what's been deprecated.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,197 @@
...
+	options.CertificatePath = os.Getenv("AZURE_CERTIFICATE_PATH")
+	options.CertificatePassword = os.Getenv("AZURE_CERTIFICATE_PASSWORD")
+
+	envName := os.Getenv("AZURE_ENVIRONMENT")
+	env, err := azure.EnvironmentFromName(envName)
+
+	if err == nil {
+		options.Environment = env
+	} else {
+		options.Environment = azure.PublicCloud
```
jhendrixMSFT: Defaulting like this seems dangerous.  What if I specified AZURE_ENVIRONMENT but misspelled the name?  I think this should be something like the following.
```golang
envName := os.Getenv("AZURE_ENVIRONMENT")
if envName == "" {
    options.Environment = azure.PublicCloud
} else {
    env, err := azure.EnvironmentFromName(envName)
    if err != nil {
        panic(err)
    }
    options.Environment = env
}
```

vladbarosan: Yes, good point

autorest/preparer.go
vladbarosan: done

vladbarosan:
Also addresses bug https://github.com/Azure/go-autorest/issues/220

PR #221, opened 2017-12-14
Add Authorizer helper methods to auth package

vladbarosan:
Add helper methods for obtaining authorizers. See this PR for any comments https://github.com/Azure/go-autorest/pull/216


autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,197 @@
...
+	options.TenantID = os.Getenv("AZURE_TENANT_ID")
+	options.ClientID = os.Getenv("AZURE_CLIENT_ID")
+	options.ClientSecret = os.Getenv("AZURE_CLIENT_SECRET")
+	options.CertificatePath = os.Getenv("AZURE_CERTIFICATE_PATH")
+	options.CertificatePassword = os.Getenv("AZURE_CERTIFICATE_PASSWORD")
+
+	envName := os.Getenv("AZURE_ENVIRONMENT")
+	env, err := azure.EnvironmentFromName(envName)
+
+	if err == nil {
```
jhendrixMSFT: This doesn't look like the updated version from the other PR.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,197 @@
...
+
+// GetAuthorizerFromMSI gets an authorizer from MSI. Note that this will only work when running in an Azure environment
+func GetAuthorizerFromMSI(env azure.Environment) (*autorest.BearerAuthorizer, error) {
+	msiEndpoint, err := adal.GetMSIVMEndpoint()
+
+	if err != nil {
+		return nil, err
+	}
+
+	spToken, err := adal.NewServicePrincipalTokenFromMSI(msiEndpoint, env.ResourceManagerEndpoint)
```
mcardosos: This wont work with graph sdk

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// GetAuthorizerFromClientCredentials gets an authorizer from client credentials.
+func GetAuthorizerFromClientCredentials(options AuthorizerOptions) (*autorest.BearerAuthorizer, error) {
+	oauthConfig, err := adal.NewOAuthConfig(options.Environment.ActiveDirectoryEndpoint, options.TenantID)
+	if err != nil {
+		return nil, err
+	}
+
+	spToken, err := adal.NewServicePrincipalToken(*oauthConfig, options.ClientID, options.ClientSecret, options.Environment.ResourceManagerEndpoint)
```
jhendrixMSFT: This uses a hard-coded endpoint, this might need to be a paramater.  Thoughts?

joshgav: can we factor the endpoint into the authorizer options so another endpoint can be chosen?

vladbarosan:
@jhendrixMSFT , @mcardosos  tag again for review. Also I didn't include it yet but I want to make a rather big change since its now or never, namely rename `Authorizer` to `Credentials`. The only problem I see with this is that we need to make the go-autorest change almost at the same time with the sdk otherwise people using go get will be broken

jhendrixMSFT:
@vladbarosan I think we should hold off on renaming Authorizer -> Credentials, I think it would be pretty disruptive and I don't see the benefit (but maybe I'm missing something).  Plus this should go away when we move to the pipeline.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,408 @@
...
+	resource := os.Getenv("AZURE_AD_RESOURCE")
+
+	var env azure.Environment
+	if envName == "" {
+		env = azure.PublicCloud
+	} else {
+		var err error
+		env, err = azure.EnvironmentFromName(envName)
+		if err != nil {
+			panic(err)
```
mcardosos: No return error?

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,408 @@
...
+	case azure.PublicCloud.GalleryEndpoint:
+		return f.GalleryEndpoint, nil
+	case azure.PublicCloud.GraphEndpoint:
+		return f.GraphResourceID, nil
+	}
+	return "", fmt.Errorf("auth: base URI not found in endpoints")
+}
+
+// NewClientCredentialsConfig creates an AuthorizerConfig object configured to obtain an Authorizer through Client Credentials.
+func NewClientCredentialsConfig(clientID string, clientSecret string, tenantID string) ClientCredentialsConfig {
```
mcardosos: >NewClientCredentialsConfig [](start = 5, length = 26)

Should this funcs also receive an Azure environment as parameter?

autorest/azure/auth/auth.go
vladbarosan: yes can return an error here also

autorest/azure/auth/auth.go
mcardosos: It is okay if you prefer to keep minimum parameters. I don't dislike the idea of those extra functions.
I just don't want auth to be painful for sovereign cloud users. It can also be clear that this func defaults to public cloud.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+	"golang.org/x/crypto/pkcs12"
+)
+
+// NewAuthorizerFromEnvironment creates an Authorizer configured from environment variables in the order:
+// 1. Client credentials
+// 2. Client certificate
+// 3. Username password
+// 4. MSI
+func NewAuthorizerFromEnvironment() (autorest.Authorizer, error) {
+	tenantID := os.Getenv("AZURE_TENANT_ID")
```
marstr: This may be as good of a time to start using a `viper` instance as any. It would allow people to specify these variables in a host of different ways. Thoughts?

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
+package auth
+
+// Copyright 2017 Microsoft Corporation
```
marstr: [nit] we've updated this to 2018 other places. Though honestly, I don't know what the correct behavior is in terms of adding copyright information.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+	spToken, err := adal.NewServicePrincipalToken(*config, file.ClientID, file.ClientSecret, resource)
+	if err != nil {
+		return nil, err
+	}
+
+	return autorest.NewBearerAuthorizer(spToken), nil
+}
+
+// File represents the authentication file
+type File struct {
```
marstr: [nit] Is there a better name for this than `File`?

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+	ClientID                string `json:"clientId,omitempty"`
+	ClientSecret            string `json:"clientSecret,omitempty"`
+	SubscriptionID          string `json:"subscriptionId,omitempty"`
+	TenantID                string `json:"tenantId,omitempty"`
```
marstr: All of these properties are marked `omitempty`. For required properties, is that the correct behavior?

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+	ManagementEndpoint      string `json:"managementEndpointUrl,omitempty"`
+}
+
+func decode(b []byte) ([]byte, error) {
+	reader, enc := utfbom.Skip(bytes.NewReader(b))
+
+	switch enc {
+	case utfbom.UTF16LittleEndian:
+		u16 := make([]uint16, (len(b)/2)-1)
+		err := binary.Read(reader, binary.LittleEndian, &u16)
```
marstr: I thought that `github.com/dimchansky/utfbom` would take care of mux'ing this for us. Do we need this `switch`?

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+// NewClientCredentialsConfig creates an AuthorizerConfig object configured to obtain an Authorizer through Client Credentials.
+// Defaults to Public Cloud and Resource Manager Endpoint.
+func NewClientCredentialsConfig(clientID string, clientSecret string, tenantID string) ClientCredentialsConfig {
+	return ClientCredentialsConfig{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
+		TenantID:     tenantID,
+		Resource:     azure.PublicCloud.ResourceManagerEndpoint,
+		AADEndpoint:  azure.PublicCloud.ActiveDirectoryEndpoint,
+	}
```
marstr: I'd add either another function that takes an environment, or add it as another parameter here.

autorest/azure/auth/auth.go
```diff
@@ -0,0 +1,412 @@
...
+		ClientID:    clientID,
+		TenantID:    tenantID,
+		Resource:    azure.PublicCloud.ResourceManagerEndpoint,
+		AADEndpoint: azure.PublicCloud.ActiveDirectoryEndpoint,
+	}
+}
+
+//AuthorizerConfig provides an authorizer from the configuration provided.
+type AuthorizerConfig interface {
+	Authorizer() (autorest.Authorizer, error)
```
marstr: Is there any opportunity to flatten and make it so that we have have Authorizers directly? Do we need both levels here?

autorest/azure/auth/auth.go
mcardosos: Remember it will be referred to as `auth.File`. But if there is a better name, well, welcome

autorest/azure/auth/auth.go
vladbarosan: well from what I know copyright is available for 99 years or something and this presumably would update the start time form when to count. Not sure in US, but for Berne Convention adherents you dont even require a copyright notice

autorest/azure/auth/auth.go
vladbarosan: Yea I brought it up in last week sync up but people opposed as they didnt want to take a dependency on it.

autorest/azure/auth/auth.go
vladbarosan: I would leave it as file. No need to call it `AuthenticationFile`

autorest/azure/auth/auth.go
vladbarosan: @mcardosos  ?

autorest/azure/auth/auth.go
vladbarosan: @mcardosos  ?

autorest/azure/auth/auth.go
vladbarosan: As we discussed also in last week we might want to pass the config as a parameter to the constructors based on the feedback and hide the authorizer.

autorest/azure/auth/auth.go
vladbarosan: @marstr  I wouldn't add it as a parameter since most of the cases they won't be changed. Was thinking of adding a withEnvironment and with Resource as functions but want to keep the interface generic for situation where it doesn't make sense to have a resource id or an environment. Maybe we can pass just the `Environment` as a parameter ?

autorest/azure/auth/auth.go
mcardosos: If you prefer to use the package, it is welcome :) Tests just need to pass :D

autorest/azure/auth/auth.go
mcardosos: Not all properties are required in all cases (I think tenantID is the only one that is required all the time, graph does not need subID).

autorest/azure/auth/auth.go
marstr: Ugh... alright I'll have to keep campaigning then.

autorest/azure/auth/auth.go
marstr: I guess the real concern is that this JSON blob doesn't necessarily need to come from a file, a buffer would be fine. Is `Options` or `Identity` something like that.

autorest/azure/auth/auth.go
marstr: Oooh, I like the "WithEnvironment" approach. I'd be game for something like that.

autorest/azure/auth/auth.go
jhendrixMSFT: Such dependencies should be made upstack, i.e. we should be providing the primitives for folks to build on top of and how the values are obtained should be at the discretion of the app writer.

autorest/azure/auth/auth.go
marstr: We should talk about this one a little more offline. I want to make sure that I understand the user experience here.

autorest/azure/auth/auth.go
vladbarosan: They can use the other authorization helpers, this was intended to be used only for CLI generated auth file. Actually im going to make it non exported.

autorest/azure/auth/auth.go
vladbarosan: Like we discussed yesterday, since in most cases these 2 params wont change and in general would prefer to keep things out and put them later then putting them in and getting stuck with them  

To mention that they can still modify the returned config to update to another environment or resource id.

autorest/azure/auth/auth.go
vladbarosan: Had a look and returning an authorizer directly form the constructor brings back the problem of the 2 extra params ( the resource and env ) since they wont be able to modify the authorizer. 
I would prefer either to:
 - Keep it as it is, so the constructor returns the configm they can modify the resource if they need to.
- Less preferred, add another `Custom` constructor with the extra params.

autorest/azure/auth/auth.go
marstr: @vladbarosan and I did a little research offline: https://godoc.org/github.com/dimchansky/utfbom#Reader

jhendrixMSFT: Sadly there are no tagged releases for this package so I don't know how the maintainer manages breaking changes.  Is this functionality that should be in go-autorest?  I'm all for not reinventing the wheel but after the shenanigans with satori's uuid package I am nervous about ending up in a similar situation.

jhendrixMSFT: Hmm looks like we already rely on this package elsewhere.

autorest/azure/auth/auth.go
marstr: @vladbarosan and I had a good long chat about this offline and agreed that this is an appropriate way to ease some Authorizer creation scenarios.

PR #227, opened 2018-01-06
Add target and innererror fields to ServiceError

demyanenko:
The data contract for ARM errors includes `innererror` field which is a property bag with unspecified contents as per [OData 4.0 spec](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091). This field wasn't used by ARM up until now, but recently there was a change related to Azure Policy that makes heavy use of this property bag. It is already deployed in all regions and consumed by Azure Portal, so now would be a good time to support it in SDKs as well.

The new errors are generated by the ARM Frontdoor itself, so CloudErrorData seems like the most appropriate contract to include it. This change also adds the optional `target` field which was missing for some reason.

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -44,20 +44,37 @@ const (
 
 // ServiceError encapsulates the error response from an Azure service.
 type ServiceError struct {
-	Code    string         `json:"code"`
-	Message string         `json:"message"`
-	Details *[]interface{} `json:"details"`
+	Code       string                 `json:"code"`
+	Message    string                 `json:"message"`
+	Target     *string                `json:"target"`
+	Details    *[]interface{}         `json:"details"`
+	InnerError map[string]interface{} `json:"innererror"`
 }
 
 func (se ServiceError) Error() string {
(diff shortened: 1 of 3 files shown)


autorest/azure/azure.go
```diff
@@ -44,20 +44,37 @@ const (
...
 // ServiceError encapsulates the error response from an Azure service.
 type ServiceError struct {
-	Code    string         `json:"code"`
-	Message string         `json:"message"`
-	Details *[]interface{} `json:"details"`
+	Code       string         `json:"code"`
+	Message    string         `json:"message"`
+	Target     *string        `json:"target"`
+	Details    *[]interface{} `json:"details"`
+	InnerError *interface{}   `json:"innererror"`
```
jhendrixMSFT: For JSON objects the type should be map[string]interface{} (and it doesn't need to be a pointer to map).

jhendrixMSFT:
Actually I'm going to move this to the dev branch, it can ship as part of v10.

Issue #225, opened 2018-01-05
Update to use dep


vladbarosan:
We are using dep in the sdk, we should update here also.

jhendrixMSFT:
Fixed in dev for v10.

PR #231, opened 2018-01-11
Switch from glide to dep.

jhendrixMSFT:
Since dep requires go 1.8+ I have removed 1.7 from travis.


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/225.

Gopkg.toml
```diff
@@ -0,0 +1,41 @@
...
+[[constraint]]
+  branch = "master"
+  name = "github.com/dimchansky/utfbom"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/mitchellh/go-homedir"
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
```
mcardosos: Interesting. I didn't know we were using testify :O

mcardosos:
@jhendrixMSFT  Are you including this in a changelog?

jhendrixMSFT:
@mcardosos I will when we ship v10 (this is in the dev branch as it's a breaking change).

Gopkg.toml
jhendrixMSFT: Is that bad?  It's been there for quite some time (i.e. was in the glide file).

mcardosos: Oh no! It is just surprise. Testify is somewhat popular when doing tests.

Issue #232, opened 2018-01-11
LROs that fail don't capture complete error information.


jhendrixMSFT:
If you look at updatePollingState() in https://github.com/Azure/go-autorest/blob/master/autorest/azure/async.go near the end all it does is copy over the status code and message.

jhendrixMSFT:
Related to https://github.com/Azure/go-autorest/issues/142.

jhendrixMSFT:
I believe the error returned is an OData v4 error, thus we should use a ```ServiceError``` type here.

jhendrixMSFT:
Fix has been merged in the dev branch.

PR #234, opened 2018-01-12
Extend pollingState to use ServiceError

jhendrixMSFT:
LROs might return error information in OData v4 format, so change the
pollingState struct to capture the complete ServiceError object.
Updated Future.Done() to return a ServiceError object if the polling
failed and the response contains an error body.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -462,18 +462,14 @@ func updatePollingState(resp *http.Response, ps *pollingState) error {
 	// -- Response
 	// -- Otherwise, Unknown
 	if ps.hasFailed() {
-		if ps.PollingMethod == PollingAsyncOperation {
-			or := pt.(*operationResource)
+		if or, ok := pt.(*operationResource); ok {
 			ps.ServiceError = &or.OperationError
+		} else if p, ok := pt.(*provisioningStatus); ok && p.hasProvisioningError() {
+			ps.ServiceError = &p.ProvisioningError
 		} else {
-			p := pt.(*provisioningStatus)
-			if p.hasProvisioningError() {
-				ps.ServiceError = &p.ProvisioningError
-			} else {
-				ps.ServiceError = &ServiceError{
-					Code:    "Unknown",
-					Message: "None",
-				}
+			ps.ServiceError = &ServiceError{
+				Code:    "Unknown",
+				Message: "None",
 			}
 		}
 	}
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -81,8 +81,35 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 
 	resp, err := sender.Do(f.req)
 	f.resp = resp
-	if err != nil || !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
+	if err != nil {
 		return false, err
+	}
+
+	if !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
+		// check response body for error content
+		if resp.Body != nil {
+			type respErr struct {
+				ServiceError ServiceError `json:"error"`
+			}
+			re := respErr{}
+
+			defer resp.Body.Close()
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/232.

autorest/azure/async.go
```diff
@@ -81,10 +81,37 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 
 	resp, err := sender.Do(f.req)
 	f.resp = resp
-	if err != nil || !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
+	if err != nil {
 		return false, err
```
vladbarosan: so here previously we would just return nil for `err` in the case the error was in the body. when would it be one or the other ? ( i.e. err != nil or the error in the body).

autorest/azure/async.go
```diff
@@ -81,10 +81,37 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
...
 	}
 
+	if !autorest.ResponseHasStatusCode(resp, pollingCodes[:]...) {
+		// check response body for error content
+		if resp.Body != nil {
+			type respErr struct {
```
vladbarosan: is this just so we can add the json tag to the field ?

autorest/azure/async.go
```diff
@@ -434,16 +464,16 @@ func updatePollingState(resp *http.Response, ps *pollingState) error {
...
-			ps.Code = or.OperationError.Code
-			ps.Message = or.OperationError.Message
+			ps.ServiceError = &or.OperationError
 		} else {
 			p := pt.(*provisioningStatus)
 			if p.hasProvisioningError() {
-				ps.Code = p.ProvisioningError.Code
-				ps.Message = p.ProvisioningError.Message
+				ps.ServiceError = &p.ProvisioningError
 			} else {
```
vladbarosan: is this for  "just in case" it happens ? otherwise when would it be an `unknown` case ?

autorest/azure/async.go
```diff
@@ -434,16 +464,16 @@ func updatePollingState(resp *http.Response, ps *pollingState) error {
...
 			or := pt.(*operationResource)
-			ps.Code = or.OperationError.Code
-			ps.Message = or.OperationError.Message
+			ps.ServiceError = &or.OperationError
 		} else {
 			p := pt.(*provisioningStatus)
 			if p.hasProvisioningError() {
-				ps.Code = p.ProvisioningError.Code
-				ps.Message = p.ProvisioningError.Message
+				ps.ServiceError = &p.ProvisioningError
```
vladbarosan: are these type assertions guaranteed to succeed ? 

if so, can we chain the type assertion with the `hasProvisioniningError()` so we reduce the extra if else ?

autorest/azure/async.go
jhendrixMSFT: sender.Do() would return a non-nil error e.g. cases of transport-level failure (say the network connection is lost) and I'm sure there are other cases (I would think of this case where the HTTP client fails to send anything over the wire).

autorest/azure/async.go
jhendrixMSFT: That's correct.  The body should look like this.
```json
{
  "error": {
    "code": "foo",
    "message": "bar"
  }
}
```

autorest/azure/async.go
jhendrixMSFT: Yes I believe this is a fall-back for a "just in case" (it's been like this since April 2016).  That said, hitting this clause is indicative of a bug somewhere in the system, i.e. either the service didn't return any error info, it was in a format that we couldn't unmarshal, or some other bug in our runtime.

autorest/azure/async.go
jhendrixMSFT: They are guaranteed to succeed as ```pt``` is set to one of the two types at the beginning of the function depending on the type of polling used.  I will clean this up.

Issue #224, opened 2018-01-02
Unify and document error return types


jhendrixMSFT:
The types of errors returned from the runtime (and by extension the generated code) is not uniform or documented making it hard for consumers to reason about errors.

jhendrixMSFT:
In go-autorest there is defined type autorest.DetailedError.  We could return this error type in all cases.

jhendrixMSFT:
Fixed in v10.0.0.

PR #237, opened 2018-01-19
Make validation failures return a unique error type.

jhendrixMSFT:
Added type validation.Error to be returned when validation of an APIs
arguments fails one or more constraints.

### DIFF
--- a/autorest/validation/validation.go
+++ b/autorest/validation/validation.go
@@ -387,9 +387,3 @@ func createError(x reflect.Value, v Constraint, err string) error {
 	return fmt.Errorf("autorest/validation: validation failed: parameter=%s constraint=%s value=%#v details: %s",
 		v.Target, v.Name, getInterfaceValue(x), err)
 }
-
-// NewErrorWithValidationError appends package type and method name in
-// validation error.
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Part of the fix for https://github.com/Azure/go-autorest/issues/224.

autorest/validation/validation.go
```diff
@@ -388,8 +388,7 @@ func createError(x reflect.Value, v Constraint, err string) error {
 		v.Target, v.Name, getInterfaceValue(x), err)
 }
 
-// NewErrorWithValidationError appends package type and method name in
-// validation error.
+// NewErrorWithValidationError returns a validation.Error type with the specified arguments.
 func NewErrorWithValidationError(err error, packageType, method string) error {
```
vladbarosan: >NewErrorWithValidationErro [](start = 5, length = 26)

This name suggests more that the returned error has an inner error a member that is validationError. I know its a breaking change, but maybe it makes sense to call it just `newValidationError` ?

jhendrixMSFT: Hmm I'm tempted to just remove this function entirely and have the generator emit calls to validation.NewError() instead.  What do you think?

PR #241, opened 2018-01-26
Fixes in the method that get cloud information from local file in case of Hybrid cloud model.

radhikagupta5:


marstr:
What is the rationale behind this change? As far as I can tell it is a breaking-change that could be just as easily navigated around by assigning the `Name` field after you've received the `Environment`.

marstr:
Oh oh, I see what you need. Let me add a commit to your PR that will accomplish what you need without introducing a breaking change.

radhikagupta5:
What is the breaking change here? This change should work for all current code paths.

radhikagupta5:
Fixed go fmt issue and test issue.

marstr:
Adding a parameter to an exported function is a breaking change :) https://github.com/Azure/go-autorest/pull/241/files#diff-3bdeb422d37dc52e27d000a3978232fdR168

radhikagupta5:
No one except us is using this function as it's in the if condition. It should be totally fine.

marstr:
Closing in favor of #242

PR #242, opened 2018-01-27
Populate Environment Name if Empty

marstr:
When reading an environment from name, that name should be assumed by default when reading from the stack file.

This PR replaces #241 
---

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -153,6 +153,12 @@ func EnvironmentFromName(name string) (Environment, error) {
 		if err != nil {
 			return unmarshaled, err
 		}
+
+		// In general, environments can be rehydrated from any file, which is how the `EnironmentFromFile`
+		// method should be used. For Stack, we introduced an environment variable "AZURE_ENVIRONMENT_FILEPATH"
+		// which will be read by default when the environment name "AzureStackCloud" is used.
+		// In the event that the Azure Stack environment was used, and no name was specifed, the clause
+		// below assigns a default name with the casing as it was handed to this function.
 		if unmarshaled.Name == "" {
 			unmarshaled.Name = name
 		}
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -149,7 +149,14 @@ func EnvironmentFromName(name string) (Environment, error) {
 	// directly call `EnvironmentFromFile`. Until then, we rely on dispatching Azure Stack environment creation
 	// from this method based on the name that is provided to us.
 	if strings.EqualFold(name, "AZURESTACKCLOUD") {
-		return EnvironmentFromFile(os.Getenv(EnvironmentFilepathName))
+		unmarshaled, err := EnvironmentFromFile(os.Getenv(EnvironmentFilepathName))
+		if err != nil {
+			return unmarshaled, err
+		}
+		if unmarshaled.Name == "" {
+			unmarshaled.Name = name
+		}
+		return unmarshaled, nil
 	}
 
 	name = strings.ToUpper(name)
--- a/autorest/azure/environments_test.go
+++ b/autorest/azure/environments_test.go
@@ -21,12 +21,33 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"sync"
 	"testing"
 )
 
 // This correlates to the expected contents of ./testdata/test_environment_1.json
 var testEnvironment1 = Environment{
 	Name:                         "--unit-test--",
+	ManagementPortalURL:          "--management-portal-url",
+	PublishSettingsURL:           "--publish-settings-url--",
+	ServiceManagementEndpoint:    "--service-management-endpoint--",
+	ResourceManagerEndpoint:      "--resource-management-endpoint--",
+	ActiveDirectoryEndpoint:      "--active-directory-endpoint--",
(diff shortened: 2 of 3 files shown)


marstr:
@radhikagupta5, please take a look and make sure this solves your issue. :)

radhikagupta5:
This does not solve all my issues. I will really appreciate if you can approve my original pull request that has all the fixes. Original change was added to a Kubernetes image and thoroughly verified. I understand that original PR change has a function definition changed but as I explained that only AzureStack folks are using that function and it should not cause any breaking changes.

marstr:
We cannot approve your original pull-request, @radhikagupta5. You cannot guarantee that nobody else was using the function that had its header changed. It would force us to release as a new major version of both this library and the Azure-SDK-for-Go. These are complete non-starters for this type of issue. 

@salameer and @devigned

autorest/azure/environments.go
```diff
@@ -149,7 +149,14 @@ func EnvironmentFromName(name string) (Environment, error) {
 	// directly call `EnvironmentFromFile`. Until then, we rely on dispatching Azure Stack environment creation
 	// from this method based on the name that is provided to us.
 	if strings.EqualFold(name, "AZURESTACKCLOUD") {
-		return EnvironmentFromFile(os.Getenv(EnvironmentFilepathName))
+		unmarshaled, err := EnvironmentFromFile(os.Getenv(EnvironmentFilepathName))
+		if err != nil {
+			return unmarshaled, err
+		}
+		if unmarshaled.Name == "" {
```
jhendrixMSFT: Can you add a comment when unmarshaled.Name would be empty string.

marstr: Done

autorest/azure/environments.go
```diff
@@ -149,7 +149,20 @@ func EnvironmentFromName(name string) (Environment, error) {
...
+		// which will be read by default when the environment name "AzureStackCloud" is used.
+		// In the event that the Azure Stack environment was used, and no name was specifed, the clause
+		// below assigns a default name with the casing as it was handed to this function.
+		if unmarshaled.Name == "" {
+			unmarshaled.Name = name
+		}
+		return unmarshaled, nil
 	}
 
 	name = strings.ToUpper(name)
```
radhikagupta5: This needs to go above the AZURESTACKCLOUD check.

radhikagupta5: I can't comment on the EnvironmentFromFile function as I don't see it here. The function doesn't return an env. even when the name is specified.

marstr: > @radhikagupta5 says:
This needs to go above the AZURESTACKCLOUD check.

This line doesn't need to go above the AZURESTACKCLOUD check, becuase [strings.EqualFold is a case-insensitive comparison](https://godoc.org/strings#EqualFold). The only reason we do a to-upper is for the sake of retrieving well-known (i.e. PublicCloud) environments from the map of basically constant ones.

>@radhikagupta5 says:
I can't comment on the EnvironmentFromFile function as I don't see it here. The function doesn't return an env. even when the name is specified.

The function signature includes a returned environement. As you can see here: https://github.com/Azure/go-autorest/blob/c2a68353555b68de3ee8455a4fd3e890a0ac6d99/autorest/azure/environments.go#L167

Further more, we have tests in place that suggest it works, at least in basic cases, that can be seen here: https://github.com/Azure/go-autorest/blob/c2a68353555b68de3ee8455a4fd3e890a0ac6d99/autorest/azure/environments_test.go#L48

If you can post an issue with a repro where it isn't working, it may help get to the bottom of the issue you're seeing. :)

radhikagupta5: @marstr This line doesn't need to go above the AZURESTACKCLOUD check, becuase strings.EqualFold is a case-insensitive comparison. The only reason we do a to-upper is for the sake of retrieving well-known (i.e. PublicCloud) environments from the map of basically constant ones.

I see this not working (being case sensitive) in Linux environments. If it does not hurt can you move this above?

radhikagupta5: @marstr If you can post an issue with a repro where it isn't working, it may help get to the bottom of the issue you're seeing. :)

I have to insert this in a Kubernetes image and produce a repro. It would be difficult to share that. So far what I understand is that the issue with original PR is due to changing function signature. Let me work on a fix without change in function.

marstr: > @radhikagupta5 says:
I see this not working (being case sensitive) in Linux environments. If it does not hurt can you move this above?

I'm not sure what you mean by this. Are you saying that strings.EqualFold doesn't do case-insensitive checks in Linux environments?

marstr: > @marstr says:
Are you saying that strings.EqualFold doesn't do case-insensitive checks in Linux environments?

Sorry, that must not be what you're saying. Are you saying the when there is no name provided in the file, that you'd like to override the casing that was handed to the `EnvironmentFromName` function and make it all-caps? That is fine if that is your desired behavior, but I don't see what Linux has to do with it. Sorry, I'm not trying to be difficult!

jhendrixMSFT:
@radhikagupta5 is there any update?  We are planning on releasing a new go-autorest today so if you want it included in that release we need to close on this ASAP (we can always add it in a minor release later if we're still iterating).

PR #244, opened 2018-01-31
fix the service bus suffix and add the service bus aad resource uri

devigned:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v9.10.0
+- Fix the Service Bus suffix in Azure public env
+- Add Service Bus Endpoint (AAD ResourceURI) for use in [Azure Service Bus RBAC Preview](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-role-based-access-control)
+
 ## v9.9.0
 
 ### New Features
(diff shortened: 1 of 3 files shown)


boumenot:
LGTM.

Thank you for fixing this!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
```
marstr: Because this is targeting the `master` branch, it'll be released immediately. For that reason, I'd label this section `## v9.10.0`. If this endpoint is useful yet, I'd submit this PR to the `dev` branch instead and not address the changelog yet.

autorest/azure/environments.go
```diff
@@ -108,6 +111,7 @@ var (
 		GalleryEndpoint:              "https://gallery.chinacloudapi.cn/",
 		KeyVaultEndpoint:             "https://vault.azure.cn/",
 		GraphEndpoint:                "https://graph.chinacloudapi.cn/",
+		ServiceBusEndpoint:           "https://servicebus.chinacloudapi.net/",
```
jhendrixMSFT: Is this the correct TLD?

devigned: based on this: https://msdn.microsoft.com/en-us/library/azure/dn578439.aspx, it appears that it should be `.cn`, which makes more sense. I'll update.

devigned:
I've corrected the `.cn` endpoints. Great catch, @jhendrixMSFT!

Issue #240, opened 2018-01-26
Concurrency safety of bearer token authorizer / service principal token?


krancour:
My team has noticed a possible race between code that uses (e.g. authorizer) or refreshes (replaces) the actual token that is wrapped by a `ServicePrincipalToken`, since these code segments seem not to be synchronizing with one another.

This line is most illustrative of our concern:

https://github.com/Azure/go-autorest/blob/c2a68353555b68de3ee8455a4fd3e890a0ac6d99/autorest/adal/token.go#L632

Note: I suppose there's a question of "does it matter?" If at any moment, there's the possibility that an authorizer is using one token or another, it _might_ not matter as long as they're both valid at that point in time. But whether that is true or not is determined by whether or not the oauth provider expires the old token upon refresh. Even supposing it does not, ideally, we shouldn't rely on that implementation detail of the oauth provider to spare us from the hazards of this race.

cc @jhendrixMSFT

jhendrixMSFT:
Fixed in v9.9.0.

krancour:
@jhendrixMSFT thanks for fixing this so quickly.

I am curious whether there is still a secondary race. Suppose that in one goroutine, an outbound HTTP request has been fully prepared by the authorizer using one token we can call `x`, but that request hasn't been submitted yet. Meanwhile, another goroutine refreshes the token, resulting in `x'`. Back in the first goroutine, the outbound HTTP request is now using an old token.

Question: Is the old token `x` still valid even after `x'` has been issued? Or is it immediately invalidated? If immediately invalidated, is there then another race here where a fully prepared request can fail authentication because another concurrent process refreshed the token?

krancour:
Also-- separate question unrelated to the RPC-- is there still race between the code that refreshes tokens and the code (e.g. authorizer) that uses them?

jhendrixMSFT:
@krancour Regarding your first comment you are correct, the token is read outside the lock so the fix is incomplete.  I will look into the second issue as I don't know the answer.

jhendrixMSFT:
I would add though that for the second case if the request fails due to the token having been expired during a refresh the retry logic will kick in and attempt the request again (not ideal but there should be no perceived error).

jhendrixMSFT:
@devigned I don't suppose you know what happens in regards to this (see above for the full context).

> Question: Is the old token x still valid even after x' has been issued?

krancour:
> the retry logic will kick in and attempt the request again

[...]

PR #245, opened 2018-01-31
Fix race condition in token auto-refresh.

jhendrixMSFT:
The previous fix was incomplete as the token could be read outside of
the lock.  Switched to a reader-writer lock for finer-grain access to
the token; this required decomposing Token from ServicePrincipalToken.
This is a breaking change.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Azure/go-autorest/autorest/date"
@@ -242,13 +243,13 @@ func (secret *ServicePrincipalCertificateSecret) SetAuthenticationValues(spt *Se
 
 // ServicePrincipalToken encapsulates a Token created for a Service Principal.
 type ServicePrincipalToken struct {
-	Token
-
+	token         Token
 	secret        ServicePrincipalSecret
 	oauthConfig   OAuthConfig
 	clientID      string
 	resource      string
 	autoRefresh   bool
+	refreshLock   *sync.RWMutex
 	refreshWithin time.Duration
 	sender        Sender
 
@@ -282,6 +283,7 @@ func NewServicePrincipalTokenWithSecret(oauthConfig OAuthConfig, id string, reso
 		clientID:         id,
 		resource:         resource,
 		autoRefresh:      true,
+		refreshLock:      &sync.RWMutex{},
 		refreshWithin:    defaultRefresh,
 		sender:           &http.Client{},
 		refreshCallbacks: callbacks,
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Updated fix for https://github.com/Azure/go-autorest/issues/240.  Note that this doesn't solve the race where the token has expired after we checked for expiration but before the request is sent.  Unfortunately this particular race is difficult to fix with our current model (it would require a pretty significant overhaul to our retry logic).  Given that we haven't had any reports of this being a problem I'm postponing fixing this for now (this will be implicitly fixed when we move to the pipeline model).

jhendrixMSFT:
@vladbarosan After thinking about this some more I agree with you that all the methods that access the token should be synchronized.

autorest/adal/token.go
```diff
@@ -534,8 +537,13 @@ func newTokenRefreshError(message string, resp *http.Response) TokenRefreshError
 // EnsureFresh will refresh the token if it will expire within the refresh window (as set by
 // RefreshWithin) and autoRefresh flag is on.
 func (spt *ServicePrincipalToken) EnsureFresh() error {
-	if spt.autoRefresh && spt.WillExpireIn(spt.refreshWithin) {
-		return spt.Refresh()
+	if spt.autoRefresh && spt.token.WillExpireIn(spt.refreshWithin) {
+		// take the write lock then check to see if the token was already refreshed
+		spt.refreshLock.Lock()
+		defer spt.refreshLock.Unlock()
+		if spt.token.WillExpireIn(spt.refreshWithin) {
```
vladbarosan: >	if spt.token.WillExpireIn(spt.refreshWithin) { [](start = 1, length = 47)

we do this extra check because the token probably it was already refreshed  if we got stock on the lock ?

jhendrixMSFT: Since the write lock serializes the refresh, we check again once we hold it to see if some other goroutine already refreshed the token (no point in refreshing again).

autorest/adal/token.go
```diff
@@ -649,3 +661,17 @@ func (spt *ServicePrincipalToken) SetRefreshWithin(d time.Duration) {
 // SetSender sets the http.Client used when obtaining the Service Principal token. An
 // undecorated http.Client is used by default.
 func (spt *ServicePrincipalToken) SetSender(s Sender) { spt.sender = s }
+
+// OAuthToken implements the OAuthTokenProvider interface.  It returns the current access token.
+func (spt *ServicePrincipalToken) OAuthToken() string {
+	spt.refreshLock.RLock()
+	defer spt.refreshLock.RUnlock()
+	return spt.token.OAuthToken()
+}
```
vladbarosan: this seems fine but at one point we need to take a look if we really need all these wrapping token types.

autorest/adal/token.go
vladbarosan: sounds good

PR #248, opened 2018-02-08
Add validation.NewErrorWithValidationError() back.

jhendrixMSFT:
Add this function back to avoid a breaking change.  The function has
been marked as deprecated and can be removed in the future.

### DIFF
--- a/autorest/validation/validation.go
+++ b/autorest/validation/validation.go
@@ -387,3 +387,11 @@ func createError(x reflect.Value, v Constraint, err string) error {
 	return fmt.Errorf("autorest/validation: validation failed: parameter=%s constraint=%s value=%#v details: %s",
 		v.Target, v.Name, getInterfaceValue(x), err)
 }
+
+// NewErrorWithValidationError appends package type and method name in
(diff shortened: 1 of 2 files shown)


autorest/validation/validation.go
```diff
@@ -387,3 +387,11 @@ func createError(x reflect.Value, v Constraint, err string) error {
 	return fmt.Errorf("autorest/validation: validation failed: parameter=%s constraint=%s value=%#v details: %s",
 		v.Target, v.Name, getInterfaceValue(x), err)
 }
+
+// NewErrorWithValidationError appends package type and method name in
+// validation error.
+//
+// Deprecated: Please use validation.NewError() instead.
+func NewErrorWithValidationError(err error, packageType, method string) error {
```
vladbarosan: Are we adding it back just so we don't remove it without warning ?

jhendrixMSFT: It's to avoid breaking the "go get" users.  We need to release v10 of go-autorest today and removing this would break everybody that uses "go get" for the SDK.

PR #258, opened 2018-03-20
added new method to load environment from URL and added UTs

seyadava:


autorest/azure/metadata_environment.go
```diff
@@ -0,0 +1,250 @@
...
+// OverrideProperty represents property name and value that clients can override
+type OverrideProperty struct {
+	Key   EnvironmentProperty
+	Value string
+}
+
+// EnvironmentFromURL loads an Environment from a URL
+// This function is particularly useful in the Hybrid Cloud model, where one may define their own
+// endpoints.
+func EnvironmentFromURL(endpoint string, properties ...OverrideProperty) (environment Environment, err error) {
```
bganapa: rename endpoint to ArmEndpoint or resourceManagerEndpoint

autorest/azure/environments.go
```diff
@@ -78,6 +79,7 @@ var (
 		ServiceManagementVMDNSSuffix: "cloudapp.net",
 		ResourceManagerVMDNSSuffix:   "cloudapp.azure.com",
 		ContainerRegistryDNSSuffix:   "azurecr.io",
+		TokenAudience:                "https://management.azure.com/",
```
jhendrixMSFT: This is a duplicate of ```ResourceManagerEndpoint```.  Why is this necessary?

autorest/azure/testdata/test_metadata_environment_1.json
```diff
@@ -0,0 +1,9 @@
+{
+    "galleryEndpoint":"https://portal.local.azurestack.external:30015/",
+    "graphEndpoint":"https://graph.windows.net/",
+    "portalEndpoint":"https://portal.local.azurestack.external/",
+    "authentication":{
+        "loginEndpoint":"https://login.windows.net/",
+        "audiences":["https://management.azurestackci04.onmicrosoft.com/3ee899c8-c137-4ce4-b230-619961a09a73"]
+    }
+}
```
jhendrixMSFT: Missing newline.

autorest/azure/metadata_environment.go
```diff
@@ -0,0 +1,250 @@
...
+	jsonResponse, err := ioutil.ReadAll(response.Body)
+	err = json.Unmarshal(jsonResponse, &environment)
+	return
+}
+
+func getStampDNSSuffix(portalEndpoint string) string {
+	return strings.TrimSuffix(strings.TrimPrefix(strings.Replace(portalEndpoint, strings.Split(portalEndpoint, ".")[0], "", 1), "."), "/")
+}
+
+func userAgent() string {
```
jhendrixMSFT: Can we use ```defaultUserAgent``` instead?  See [here](https://github.com/Azure/go-autorest/blob/19642ad23ee58fb8c895dfdc97c979c15c133118/autorest/client.go#L46).

autorest/azure/metadata_environment.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return environment, err
+	}
+	defer response.Body.Close()
+	jsonResponse, err := ioutil.ReadAll(response.Body)
+	err = json.Unmarshal(jsonResponse, &environment)
+	return
+}
+
+func getStampDNSSuffix(portalEndpoint string) string {
```
jhendrixMSFT: It looks like this is called in only one place, can it be folded into the caller?

autorest/azure/environments.go
bganapa: For AzureStack, the TokenAudience is different from the ResourceManagerEndpoint, Hence this is needed to support AzureStack better

jhendrixMSFT: I see, it just defaults to the same value as ResourceManagerEndpoint?

bganapa: Yes, in case of AzureClouds audiences value is same as RM endpoint e.g. 
pls look at audiences value
https://management.azure.com/metadata/endpoints?api-version=1.0
https://management.usgovcloudapi.net/metadata/endpoints?api-version=1.0

in case of azure stack, the audiences value differs with each deployment. (i am not sure whether you have access to the following endpoints)
https://management.redmond.azurestack.corp.microsoft.com/metadata/endpoints?api-version=1.0
https://management.shanghai.azurestack.corp.microsoft.com/metadata/endpoints?api-version=1.0

autorest/azure/metadata_environment.go
seyadava: defaultUserAgent is not an exposed variable, so we cannot use it to instantiate client. 
However, instead of using NewClientWithUserAgent method, I could directly instantiate a client, like: 
client := &autorest.Client{
		PollingDelay:    autorest.DefaultPollingDelay,
		PollingDuration: autorest.DefaultPollingDuration,
		RetryAttempts:   autorest.DefaultRetryAttempts,
		RetryDuration:   autorest.DefaultRetryDuration,
	}
This way we don't need to pass userAgent. Thoughts?

autorest/azure/metadata_environment.go
seyadava: Renamed it to "resourceManagerEndpoint"

autorest/azure/testdata/test_metadata_environment_1.json
seyadava: fixed.

autorest/azure/metadata_environment.go
seyadava: Folded it into the caller

autorest/azure/metadata_environment.go
jhendrixMSFT: I think you can continue to use ```NewClientWithUserAgent```, just pass the empty string for the param.

seyadava: Done.

autorest/azure/metadata_environment.go
```diff
@@ -0,0 +1,242 @@
...
+func retrieveMetadataEnvironment(endpoint string) (environment environmentMetadataInfo, err error) {
+	client := autorest.NewClientWithUserAgent("")
+	managementEndpoint := fmt.Sprintf("%s%s", strings.TrimSuffix(endpoint, "/"), "/metadata/endpoints?api-version=1.0")
+	req, _ := http.NewRequest("GET", managementEndpoint, nil)
+	response, err := client.Do(req)
+	if err != nil {
+		return environment, err
+	}
+	defer response.Body.Close()
+	jsonResponse, err := ioutil.ReadAll(response.Body)
```
jhendrixMSFT: Can you please update this to return if err is not nil?

autorest/azure/metadata_environment.go
```diff
@@ -0,0 +1,245 @@
...
+// OverrideProperty represents property name and value that clients can override
+type OverrideProperty struct {
+	Key   EnvironmentProperty
+	Value string
+}
+
+// EnvironmentFromURL loads an Environment from a URL
+// This function is particularly useful in the Hybrid Cloud model, where one may define their own
+// endpoints.
+func EnvironmentFromURL(resourceManagerEndpoint string, properties ...OverrideProperty) (environment Environment, err error) {
```
jhendrixMSFT: Since this is variadic the caller can elect to omit this parameter entirely.  What happens in that case, should it return an error?

seyadava: If OverrideProperty is not passed, it means caller does not want to override any environment value. In this case method uses default values for all of the environment properties and no error should be returned.

PR #259, opened 2018-03-24
Add Azure resource ID helper

vladbarosan:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v10.4.0
+
+### New Features
+- Added helper for parsing Azure Resource id's.
+- Added deprecation message to utils.GetEnvVarOrExit()
+
 ## v10.3.0
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## v10.4.0
 
 ### New Features
-- Added helper for parsing Azure Resource id's.
+- Added helper for parsing Azure Resource ID's.
 - Added deprecation message to utils.GetEnvVarOrExit()
 
 ## v10.3.0
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
 )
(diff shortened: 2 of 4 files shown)


autorest/azure/azure.go
```diff
@@ -22,6 +22,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"regexp"
+	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
```
vladbarosan: I preffered this package since its azure specific but if you prefer the autorest package ( inside utility.go ) let me know

vladbarosan:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/322

autorest/azure/azure.go
```diff
@@ -147,6 +149,40 @@ func IsAzureError(e error) bool {
 	return ok
 }
 
+// Resource contains details about an Azure resource
+type Resource struct {
+	subscription string
+	resourceGroup string
+	provider string
+	resourceType string
+	resourceName string
```
mcardosos: Should this ones be exported?

vladbarosan: what do you mean by these ones ? The type itself needs to be exported, that is what will contain the parsed result.

vladbarosan: or you want to return an unexported type from the parse function ? I dont think that is a good pattern. Golint even complains about that https://github.com/golang/lint/issues/210. 

That being said if we have a good reason to not export it we can do that.

jhendrixMSFT: She's asking if the fields should be exported which I think they should yes?  If not then how do you access their values?

vladbarosan: Oh sorry my bad yes

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v10.4.0
+
+### New Features
+- Added helper for parsing Azure Resource id's.
```
jhendrixMSFT: Nit, "id" should be capitalized.

autorest/azure/azure.go
```diff
@@ -147,6 +149,40 @@ func IsAzureError(e error) bool {
 	return ok
 }
 
+// Resource contains details about an Azure resource
```
jhendrixMSFT: Can you add a link to the docs on Azure resource IDs (the one you had sent me).

vladbarosan: 👍

autorest/azure/azure.go
mcardosos: What @jhendrixMSFT is what I meant, yes

PR #261, opened 2018-03-30
Make retry logic use the request's context.

jhendrixMSFT:
When polling, use the request's context object instead of the deprecated
Cancel object to see if polling should continue or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+## v10.5.0
+
+### New Features
+
+- Added NewPollingRequestWithContext() for use with polling asynchronous operations.
+
+### Bug Fixes
+
+- Make retry logic use the request's context instead of the deprecated Cancel object.
+
 ## v10.4.0
 
 ### New Features
--- a/autorest/autorest.go
+++ b/autorest/autorest.go
@@ -136,14 +136,14 @@ func NewPollingRequest(resp *http.Response, cancel <-chan struct{}) (*http.Reque
 func NewPollingRequestWithContext(ctx context.Context, resp *http.Response) (*http.Request, error) {
 	location := GetLocation(resp)
 	if location == "" {
-		return nil, NewErrorWithResponse("autorest", "NewPollingRequest", resp, "Location header missing from response that requires polling")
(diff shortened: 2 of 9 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/1450

autorest/autorest.go
```diff
@@ -130,3 +131,20 @@ func NewPollingRequest(resp *http.Response, cancel <-chan struct{}) (*http.Reque
 
 	return req, nil
 }
+
+// NewPollingRequestWithContext allocates and returns a new http.Request with the specified context to poll for the passed response.
+func NewPollingRequestWithContext(ctx context.Context, resp *http.Response) (*http.Request, error) {
+	location := GetLocation(resp)
+	if location == "" {
+		return nil, NewErrorWithResponse("autorest", "NewPollingRequest", resp, "Location header missing from response that requires polling")
```
mcardosos: >NewPollingRequest [](start = 48, length = 17)

WithContext?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v10.4.1
```
mcardosos: >4.1 [](start = 7, length = 3)

Shouldn't this be 10.5?

autorest/autorest.go
jhendrixMSFT: It's basically a copy of NewPollingRequest() that takes a context param (see the preceding func def).

CHANGELOG.md
jhendrixMSFT: Bleh you're right, I'll update.

jhendrixMSFT: Fixed.  This also prompted me to notice that I forgot to update version.go 👍

jhendrixMSFT:
@devigned No worries (besides who likes breaking changes?  :p).  I'd be interested to know if this resolves the issue you were seeing.

autorest/autorest.go
mcardosos: I mean `NewErrorWithResponse` second parameter should have a different value

jhendrixMSFT: Derp, will fix.

devigned:
@jhendrixMSFT I've verified that my test reacts as I would expect when the context is cancelled. Thank you for the fix!

Issue #262, opened 2018-04-02
There is Azure specific code in the agnostic autorest package


mcardosos:
I refer to the `authorization.go`  file. I think anything that uses the `adal` package should be moved there.

marstr:
Done!

PR #263, opened 2018-04-03
Add support for MSI through IDMS

mcardosos:
Related issue https://github.com/Azure/go-autorest/issues/262
Because adal package is referenced by the autorest package, I cannot reference autorest in adal package because of circular dependencies.
I would like that to be clean for the next major release.
Therefore I am copying and pasting a bunch of code related to retries :(


autorest/adal/token.go
```diff
@@ -54,6 +54,9 @@ const (
 
 	// metadataHeader is the header required by MSI extension
 	metadataHeader = "Metadata"
+
+	// imdsEndpoint is the well known endpoint for getting authentications tokens inside Azure virtual machines
```
vladbarosan: I would change the comment to say that IDMS has a well known endpoint for obtaining an MSI token

autorest/azure/auth/auth.go
```diff
@@ -38,7 +38,7 @@ import (
 // 1. Client credentials
 // 2. Client certificate
 // 3. Username password
-// 4. MSI
+// 4. IMDS
```
vladbarosan: >IMDS [](start = 6, length = 4)

is still through MSI, IDMS is just a service that exists on every VM in Azure

autorest/azure/auth/auth.go
```diff
@@ -89,10 +89,10 @@ func NewAuthorizerFromEnvironment() (autorest.Authorizer, error) {
 		return config.Authorizer()
 	}
 
-	//4. By default return MSI
-	config := NewMSIConfig()
+	// 4. IDMS
+	config := NewIDMSConfig()
```
vladbarosan: >NewIDMSConfig [](start = 11, length = 13)

same thing, no need for a breaking change

autorest/azure/auth/auth.go
```diff
@@ -89,10 +89,10 @@ func NewAuthorizerFromEnvironment() (autorest.Authorizer, error) {
...
 	config.Resource = resource
-
+	config.ClientID = clientID
```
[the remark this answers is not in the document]
vladbarosan: it seems from here that its not [https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/active-directory/pp/msi-how-to-use-vm-msi-token.md](https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/active-directory/pp/msi-how-to-use-vm-msi-token.md)

---
In reply to: [178685218](https://github.com/Azure/go-autorest/pull/263#discussion_r178685218) [](ancestors = 178685218)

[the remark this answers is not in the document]
mcardosos: It is an optional parameter

autorest/azure/auth/auth.go
```diff
@@ -393,16 +400,30 @@ type MSIConfig struct {
...
 	spToken, err := adal.NewServicePrincipalTokenFromMSI(msiEndpoint, mc.Resource)
-
 	if err != nil {
 		return nil, fmt.Errorf("failed to get oauth token from MSI: %v", err)
 	}
 
 	return autorest.NewBearerAuthorizer(spToken), nil
 }
+
+// IDMSConfig provides the options to get a bearer authorizer through IDMS.
```
vladbarosan: >IDMSConfig [](start = 3, length = 10)

IMDS should be an implementation detail for users. They just want to use MSI, they shouldnt care if its through vm extension or IMDS.

autorest/azure/auth/auth.go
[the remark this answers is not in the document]
vladbarosan: Oh ok, this is for user assigend MSI I presume

---
In reply to: [178685577](https://github.com/Azure/go-autorest/pull/263#discussion_r178685577) [](ancestors = 178685577)

vladbarosan:
> }

we will need then to add here the client Id also

---
Refers to: autorest/azure/auth/auth.go:398 in 1dec803. [](commit_id = 1dec803377041dc6e54482e0b6d7b7291ab40784, deletion_comment = False)

autorest/adal/token.go
```diff
@@ -588,35 +627,48 @@ func (spt *ServicePrincipalToken) getGrantType() string {
...
+				return err
+			}
+		}
+
+		s := v.Encode()
+		body := ioutil.NopCloser(strings.NewReader(s))
+		req.ContentLength = int64(len(s))
+		req.Header.Set(contentType, mimeTypeFormPost)
+		req.Body = body
+	}
```
vladbarosan: if its an MSI request we dont need this ?

mcardosos: Nope

autorest/adal/retry.go
```diff
@@ -0,0 +1,180 @@
+// +build go1.8
+
+// Copyright 2017 Microsoft Corporation
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
```
vladbarosan: also update changelog

autorest/adal/retry.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	rr.br = bytes.NewReader(b)
+	rr.req.Body = ioutil.NopCloser(rr.br)
+	return err
+}
+
+func retry(req *http.Request, backoff time.Duration, attempts int) (resp *http.Response, err error) {
+	s := http.Client{}
+	rr := newRetriableRequest(req)
```
vladbarosan: >(req) [](start = 26, length = 5)

just checking, is this mantaining the initials request context ?

mcardosos: No

autorest/adal/retry.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	return err
+}
+
+func removeRequestBody(req *http.Request) {
+	req.Body = nil
+	req.GetBody = nil
+	req.ContentLength = 0
+}
```
vladbarosan: is this being used

autorest/adal/token.go
```diff
@@ -588,35 +627,48 @@ func (spt *ServicePrincipalToken) getGrantType() string {
...
+
 	if _, ok := spt.secret.(*ServicePrincipalMSISecret); ok {
 		req.Header.Set(metadataHeader, "true")
 	}
-	resp, err := spt.sender.Do(req)
+
+	resp, err := retry(req, 1*time.Second, 5)
```
vladbarosan: > 5 [](start = 39, length = 2)

maybe put this in a constant

autorest/adal/token.go
mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: Fixed

autorest/adal/retry.go
mcardosos: Nope

autorest/adal/token.go
mcardosos: Fixed

autorest/adal/retry.go
mcardosos: Fixed

autorest/adal/retry.go
jhendrixMSFT: Why not?

jhendrixMSFT:
@mcardosos I think we need to figure out a way to not duplicate the retry code.  It's hard enough to get right without having to maintain two copies.  I recall we have this circular dependency problem, is there no way to remove adal's dependency on autorest?  Seems like it should be able to stand on its own.

mcardosos:
@jhendrixMSFT I really don't want to duplicate code.
autorest is the one that depends on adal, not the other way around. We can change this, but it would be a breaking change.

mcardosos:
I propose we fix this dependency stuff in the dev branch for when we are ready for v11

jhendrixMSFT:
@mcardosos is the only thing you need the RetriableRequest?  What if we copied it into its own package outside of autorest, deprecate the version in autorest, and switch all the code to use the new one?  We can then remove the deprecated copy in v11.

mcardosos:
@jhendrixMSFT I would actually like to also use some of the sender methods.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+
+## v10.6.0
+
+- Changed MSI token implementation. Now, the token endpoint is a well known endpoint. It is not longer got from MSI file in VMs.
```
vladbarosan: >. Now, the token endpoint is a well known endpoint. It is not longer got from MSI file in VMs. [](start = 34, length = 94)

I would mention just that now is using an IMDS endpoint to get the token

autorest/adal/token.go
```diff
@@ -54,6 +55,9 @@ const (
 
 	// metadataHeader is the header required by MSI extension
 	metadataHeader = "Metadata"
+
+	// imdsEndpoint is the well known endpoint for getting MSI authentications tokens
+	imdsEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
```
vladbarosan: >imdsEndpoint [](start = 1, length = 12)

also not sure if this is confusing or not, maybe still name it MSI endpoint as this is what this endpoint is ( just that is part of IMDS now )

vladbarosan:
> 

are these files moved or copied ? 

---
Refers to: autorest/retriablerequest/autorest.go:119 in 6fc02b8. [](commit_id = 6fc02b8592fd9508aec03c12f2659b53ab604a80, deletion_comment = False)

mcardosos:
@vladbarosan Copied, to avoid breaking changes

vladbarosan:
@mcardosos then are we deprecating the other ones ?

autorest/azure/auth/auth.go
```diff
@@ -90,10 +90,10 @@ func NewAuthorizerFromEnvironment() (autorest.Authorizer, error) {
 		return config.Authorizer()
 	}
 
-	//4. By default return MSI
+	// 4. MSI
 	config := NewMSIConfig()
 	config.Resource = resource
-
+	config.ClientID = clientID
```
vladbarosan: >ClientID = clientID [](start = 8, length = 19)

you tested? its ok if you leave the ClientId env variable unset ?

PR #269, opened 2018-04-18
Retries

mcardosos:

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -696,6 +696,7 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
 				delayed = delay(resp, req.Cancel)
 			}
 			if !delayed {
+				time.Sleep(time.Second)
 				attempt++
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


autorest/adal/token.go
```diff
@@ -653,6 +658,78 @@ func (spt *ServicePrincipalToken) refreshInternal(resource string) error {
...
+			return
+		}
+
+		if resp.StatusCode == http.StatusOK {
+			return
+		}
+		if containsInt(retries, resp.StatusCode) {
+			delayed := false
+			if resp.StatusCode == http.StatusTooManyRequests {
+				delayed = delay(resp, req.Cancel)
```
vladbarosan: >delay [](start = 14, length = 5)

we should check the delay for the other status codes also. But lets keep the logic for StatusTooManyRequests to retyr only if the retryAfter header is present.

autorest/adal/token.go
```diff
@@ -653,6 +658,78 @@ func (spt *ServicePrincipalToken) refreshInternal(resource string) error {
 	return spt.InvokeRefreshCallbacks(token)
 }
 
+func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
```
vladbarosan: >retry(s [](start = 5, length = 7)

Also lets simulate the retry to make sure it works fine.

PR #270, opened 2018-04-18
Pass resource instead off AAD endpoint

mcardosos:


autorest/azure/auth/auth.go
```diff
@@ -328,7 +328,7 @@ type DeviceFlowConfig struct {
 func (dfc DeviceFlowConfig) Authorizer() (autorest.Authorizer, error) {
```
vladbarosan: changelog update ?

mcardosos: Added

Issue #256, opened 2018-03-06
Add key vault dataplane authorizer


jhendrixMSFT:
The key vault [dataplane sample](https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/master/keyvault/operations.go) uses a custom authorizer defined in the sample helper lib [here](https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/4559164aef694aad3b11d9cd76bde6ea008e72c8/iam/oauth.go#L194).  Seems like this should be productized in go-autorest.

vladbarosan:
Seems here just some duplicate logic from go-autorest. Should be fixed when we get do https://github.com/Azure-Samples/azure-sdk-for-go-samples/issues/92

vladbarosan:
@mcardosos feel free to close this as a duplicate if you want or wait until Azure-Samples/azure-sdk-for-go-samples#92 is fixed

PR #273, opened 2018-04-20
Add keyvault authorizer

mcardosos:
Fixes https://github.com/Azure/go-autorest/issues/256

### DIFF
--- a/autorest/azure/auth/auth.go
+++ b/autorest/azure/auth/auth.go
@@ -41,42 +41,16 @@ import (
 // 3. Username password
 // 4. MSI
 func NewAuthorizerFromEnvironment() (autorest.Authorizer, error) {
-	settings := getEnvVars()
-
-	env, err := getEnv(settings.envName)
+	settings, err := GetAuthSettings()
 	if err != nil {
 		return nil, err
 	}
 
-	if settings.resource == "" {
-		settings.resource = env.ResourceManagerEndpoint
-	}
-
-	return getAuth(settings, env)
+	return settings.GetAuth()
 }
 
-// NewKeyvaultAuthorizerFromEnvironment creates a keyvault dataplane Authorizer configured from environment variables in the order:
-// 1. Client credentials
-// 2. Client certificate
-// 3. Username password
-// 4. MSI
-func NewKeyvaultAuthorizerFromEnvironment() (autorest.Authorizer, error) {
-	settings := getEnvVars()
-	settings.resource = os.Getenv("AZURE_KEYVAULT_RESOURCE")
-
-	env, err := getEnv(settings.envName)
-	if err != nil {
-		return nil, err
-	}
-
-	if settings.resource == "" {
-		settings.resource = strings.TrimSuffix(env.KeyVaultEndpoint, "/")
-	}
-
-	return getAuth(settings, env)
-}
-
-type envSettings struct {
+// Settings reporesnt settings for authentication
+type Settings struct {
 	tenantID            string
 	clientID            string
 	clientSecret        string
@@ -85,11 +59,14 @@ type envSettings struct {
 	username            string
 	password            string
 	envName             string
-	resource            string
+	Resource            string
+	Environment         azure.Environment
 }
 
-func getEnvVars() envSettings {
-	return envSettings{
+// GetAuthSettings gets authentication settings. This is
+// meant to be used later by Settings.GetAuth()
+func GetAuthSettings() (s Settings, err error) {
+	s = Settings{
 		tenantID:            os.Getenv("AZURE_TENANT_ID"),
 		clientID:            os.Getenv("AZURE_CLIENT_ID"),
 		clientSecret:        os.Getenv("AZURE_CLIENT_SECRET"),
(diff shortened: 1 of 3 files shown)


autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,105 @@ import (
...
+
+	return getAuth(settings, env)
+}
+
+// NewKeyvaultAuthorizerFromEnvironment creates a keyvault dataplane Authorizer configured from environment variables in the order:
+// 1. Client credentials
+// 2. Client certificate
+// 3. Username password
+// 4. MSI
+func NewKeyvaultAuthorizerFromEnvironment() (autorest.Authorizer, error) {
```
vladbarosan: I see the point in case you need to auth to multiple resources . but does it make more sense to have a `NewAuthorizerFromEnvironmentWithResource` instead of creating a seperate function for each ? That way whatever resource they have they have they can pass it without making a request.

autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,105 @@ import (
...
+	}
+
+	if settings.resource == "" {
+		settings.resource = strings.TrimSuffix(env.KeyVaultEndpoint, "/")
+	}
+
+	return getAuth(settings, env)
+}
+
+type envSettings struct {
```
marstr: I'd like to change this name. I think it'll be easily confused with this: https://github.com/Azure/go-autorest/blob/1ff28809256a84bb6966640ff3d0371af82ccba4/autorest/azure/environments.go#L37

Not to mention, I think it also implies that the only way to populate these are with Environment variables. (Which totally is one good way, but maybe we want expose others someday. Command line args parser, for example.)

Maybe `identityKeys`, `authKeys` or `authSettings` would be better?

autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,105 @@ import (
...
-		}
-	}
-
-	if resource == "" {
-		resource = env.ResourceManagerEndpoint
 	}
+	return
+}
 
+func getAuth(settings envSettings, env azure.Environment) (autorest.Authorizer, error) {
```
jhendrixMSFT: I think this should be a method on envSettings.

autorest/azure/auth/auth.go
jhendrixMSFT: What if we had both, with the specialized ones building on the general ones?  E.g. look at line 73, we see customers use env.KeyVaultEndpoint as the resource for key vault which fails due to trailing '/' character.  I'm sure there are other solutions to that particular issue, but I think there's value in having "brain-dead simple" authorizers.

jhendrixMSFT: To add to my previous comment, if we think it makes more sense to have these specialized ones as convenience methods in the SDK (i.e. not go-autorest) it works for me.

mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: Fixed

autorest/azure/auth/auth.go
mcardosos: I opened a PR on the SDK with Joel's suggestion

mcardosos:
Poke @jhendrixMSFT  @vladbarosan @marstr 
I would like to merge this soon

autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,100 @@ import (
...
+	settings, err := GetAuthSettings()
+	if err != nil {
+		return nil, err
 	}
+	settings.Resource = resource
+	return settings.GetAuth()
+}
+
+// Settings reporesnt settings for authentication
+type Settings struct {
```
vladbarosan: im not sure it provides any value to expose this extra type since you added the `NewAuthorizerFromEnvironmentWithResource` .

autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,100 @@ import (
...
+		s.Environment = azure.PublicCloud
+	} else {
+		s.Environment, err = azure.EnvironmentFromName(s.envName)
 	}
+	return
+}
 
+// GetAuth gets sn autorest.Authorizer from the provided
+// settings. This is meant to be used after GetAuthSettings()
+func (settings Settings) GetAuth() (autorest.Authorizer, error) {
```
vladbarosan: getAuthorizer maybe ?

autorest/azure/auth/auth.go
```diff
@@ -41,59 +41,100 @@ import (
...
+	envName             string
+	Resource            string
+	Environment         azure.Environment
+}
 
-	if resource == "" {
-		resource = env.ResourceManagerEndpoint
+// GetAuthSettings gets authentication settings. This is
+// meant to be used later by Settings.GetAuth()
+func GetAuthSettings() (s Settings, err error) {
```
vladbarosan: this also should be private. and to be less confusing maybe getAuthorizerSettings

Issue #276, opened 2018-05-04
The retry for MSI token doesn't follow the retry guidance


bingosummer:
It seems that the retry in https://github.com/Azure/go-autorest/pull/269 is different from the retry guidance in the MSI doc.
https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/how-to-use-vm-token#retry-guidance

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v10.9.1

PR #281, opened 2018-05-30
IMDS retry logic should use exponential backoff.

jhendrixMSFT:
Fixed retry logic to use exponential backoff per retry guidelines.
Added StatusGone to the list of IMDS-specific retry status codes.
Changed IsTemporaryNetworkError to be conservative, so errors that don't
implement the net.Error interface are now considered temporary.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -716,8 +716,7 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
 	// see https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/how-to-use-vm-token#retry-guidance
 
 	const maxAttempts int = 5
-	const deltaDelay int = 2          // seconds
-	const maxDelay time.Duration = 60 // seconds
+	const maxDelay time.Duration = 60 * time.Second
 
 	attempt := 0
 	delay := time.Duration(0)
@@ -732,13 +731,14 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
 		// perform exponential backoff with a cap.
 		// must increment attempt before calculating delay.
 		attempt++
-		delay += time.Duration(math.Pow(float64(deltaDelay), float64(attempt)))
+		// the base value of 2 is the "delta backoff" as specified in the guidance doc
+		delay += (time.Duration(math.Pow(2, float64(attempt))) * time.Second)
 		if delay > maxDelay {
 			delay = maxDelay
 		}
 
 		select {
-		case <-time.After(delay * time.Second):
+		case <-time.After(delay):
 			// intentionally left blank
 		case <-req.Context().Done():
 			err = req.Context().Err()
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"net"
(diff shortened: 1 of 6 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/1848 and https://github.com/Azure/go-autorest/issues/276

autorest/adal/token.go
```diff
@@ -709,36 +713,51 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
 		http.StatusNotExtended,
 		http.StatusNetworkAuthenticationRequired)
 
+	// see https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/how-to-use-vm-token#retry-guidance
+
+	const maxAttempts int = 5
+	const deltaDelay int = 2          // seconds
```
marstr: I'm not sure this needs to be a constant, it just seems like it's forcing us to choose an awkward name, and it's not even really in seconds. I'd just get rid of it and use `math.Pow(2.0, float64(attempt))` below.

autorest/adal/token.go
```diff
@@ -709,36 +713,51 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
...
+	const maxDelay time.Duration = 60 // seconds
```
marstr: Nit: I'd declare this as `const maxDelay = 60 * time.Second`

autorest/adal/token.go
```diff
@@ -709,36 +713,51 @@ func retry(sender Sender, req *http.Request) (resp *http.Response, err error) {
...
-			case <-time.After(time.Second):
-				attempt++
-			case <-req.Context().Done():
-				err = req.Context().Err()
-				return
-			}
+		// perform exponential backoff with a cap.
+		// must increment attempt before calculating delay.
+		attempt++
+		delay += time.Duration(math.Pow(float64(deltaDelay), float64(attempt)))
```
marstr: Should this be `delay = time...`  instead of  `delay += time...`? Seems weird to me, but I'm also not familiar with the particular timing expectations.

marstr: If you fixed all of the of the other comments, this line would become:

``` Go
delay = time.Seconds * time.Duration(math.Pow(2, float64(attempt)))
```

autorest/adal/token.go
jhendrixMSFT: I would like to avoid magic numbers.  The name comes from the spec here https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/how-to-use-vm-token#retry-guidance (agree it isn't the best but I wanted to keep it close).

jhendrixMSFT: Hmm thinking about this a bit more I will remove this const and add a comment where it's being used.

autorest/adal/token.go
jhendrixMSFT: That's how I would have written it too, I'm just trying to match what they documented.

autorest/adal/token.go
jhendrixMSFT: Fixed.

autorest/adal/token.go
jhendrixMSFT: Fixed.

autorest/adal/token.go
jhendrixMSFT: Fixed.

PR #293, opened 2018-06-21
Fix panic in LRO implementation

jhendrixMSFT:
If an LRO returns http.StatusOK on the initial response with no async
headers return the response body from Future.GetResult().
If there is no "final GET URL" return an error.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -245,7 +245,13 @@ func (f Future) PollingURL() string {
 // It makes the final GET call to retrieve the resultant payload.
 func (f Future) GetResult(sender autorest.Sender) (*http.Response, error) {
 	if f.pt.finalGetURL() == "" {
-		return nil, nil
+		// we can end up in this situation if the async operation returns a 200
+		// with no polling URLs.  in that case return the response which should
+		// contain the JSON payload.
+		if lr := f.pt.latestResponse(); lr != nil {
+			return lr, nil
+		}
+		return nil, autorest.NewError("Future", "GetResult", "missing URL for retrieving result")
 	}
 	req, err := http.NewRequest(http.MethodGet, f.pt.finalGetURL(), nil)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2090

autorest/azure/async.go
```diff
@@ -245,7 +245,13 @@ func (f Future) PollingURL() string {
...
 func (f Future) GetResult(sender autorest.Sender) (*http.Response, error) {
 	if f.pt.finalGetURL() == "" {
-		return nil, nil
+		// we can end up in this situation if the async operation returns a 200
+		// with no polling URLs.  in that case return the response which should
+		// contain the JSON payload.
+		if lr := f.pt.latestResponse(); lr != nil {
+			return lr, nil
+		}
+		return nil, autorest.NewError("Future", "GetResult", "missing URL for retrieving result")
```
vladbarosan: qq here: in the cases where the original response is actually the response (i.e. a 200) is it mandatory that the original response to returna URL to get the final result ?

vladbarosan: the nested if does ?

---
In reply to: [197263333](https://github.com/Azure/go-autorest/pull/293#discussion_r197263333) [](ancestors = 197263333)

jhendrixMSFT: LRO final result can come from either a "final GET URL" or as part of the final response body.  Usually LRO initial response should be a 201/202 however there are some older ones that return a 200 with a provisioning status in the initial response body.  For the case pointed out by Terraform the operation is marked as a LRO but doesn't really appear to be as it returns a 200 with the payload (i.e. no provisioning status/async headers).  I'm following up with the RP to understand why this operation is marked as LRO however we should try to handle this case gracefully.  This has also got me thinking that I should update the condition to ```lr != nil && f.pt.hasSucceeded()``` as we shouldn't be returning the latest response in non-terminal cases.

vladbarosan: Yes, compute has a bunch of operations that they say it could return a 200 with the final payload as the response. Just making sure that exactly this case you mentioned we are taking care of.

Issue #298, opened 2018-07-06
emit traces when env var set


joshgav:
Can we add a way (or two) to request traces via configuration?

Today, getting traces requires source modifications as described in [the SDK README](https://github.com/Azure/azure-sdk-for-go#inspecting-and-debugging). That is difficult and sometimes not possible, e.g. when we're embedded in a tool as in Azure/azure-sdk-for-go#1828.

The SDK should emit more raw trace data to stderr when an environment variable is set for the sake of support and troubleshooting. A potential scheme could be to check the env var `AZURE_GO_SDK_LOG_LEVEL=[0|1|2]`, with 0 meaning log nothing; 1 meaning log URLs and headers; and 2 meaning log URLs, headers, and bodies.

This should be alongside OpenTracing or OpenCensus spans if/when we add them, meaning I think we still need raw traces sometimes. We can use the same mechanism for both if that makes sense.

Thoughts?

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v10.15.0

PR #301, opened 2018-07-11
Add basic request/respopnse logging

jhendrixMSFT:
Added basic logging inspectors that are hooked up to the default client.
They are off by default but can be enabled by setting environment
variable AZURE_GO_AUTOREST_LOG_LEVEL to one of two values.
basic - includes URL and headers
full - everything in basic plus request/response bodies
The value of the Authorization header is redacted to avoid inadvertently
leaking access tokens.
Each request/response includes a timestamp.
Logging spew defaults to stderr but can be written to stdout or a log
file by setting AZURE_GO_AUTOREST_LOG_FILE with the desired destination.

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -22,8 +22,9 @@ import (
 	"log"
 	"net/http"
 	"net/http/cookiejar"
-	"runtime"
 	"time"
+
+	"github.com/Azure/go-autorest/version"
 )
 
 const (
@@ -41,15 +42,6 @@ const (
 )
 
 var (
-	// defaultUserAgent builds a string containing the Go version, system archityecture and OS,
-	// and the go-autorest version.
-	defaultUserAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
-		runtime.Version(),
-		runtime.GOARCH,
-		runtime.GOOS,
-		Version(),
-	)
-
 	// StatusCodesForRetry are a defined group of status code for which the client will retry
 	StatusCodesForRetry = []int{
 		http.StatusRequestTimeout,      // 408
@@ -179,7 +171,7 @@ func NewClientWithUserAgent(ua string) Client {
 		PollingDuration: DefaultPollingDuration,
 		RetryAttempts:   DefaultRetryAttempts,
 		RetryDuration:   DefaultRetryDuration,
-		UserAgent:       defaultUserAgent,
+		UserAgent:       version.UserAgent(),
 	}
 	c.Sender = c.sender()
 	c.AddToUserAgent(ua)
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -22,8 +22,11 @@ import (
 	"log"
 	"net/http"
 	"net/http/cookiejar"
+	"os"
+	"strings"
 	"time"
 
+	"github.com/Azure/go-autorest/autorest/logger"
 	"github.com/Azure/go-autorest/version"
 )
 
@@ -63,6 +66,36 @@ const (
 ===================================================== HTTP Response End
 `
 )
+
+var logWriter logger.Writer
+
+func init() {
+	llStr := strings.ToLower(os.Getenv("AZURE_GO_AUTOREST_LOG_LEVEL"))
+	if llStr == "" {
+		return
+	}
+	ll, err := logger.ParseLevel(llStr)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to parse log level: %s\n", err.Error())
+		return
+	}
+	if ll > logger.LogNone {
+		// default to stderr
+		dest := os.Stderr
+		lfStr := os.Getenv("AZURE_GO_AUTOREST_LOG_FILE")
+		if strings.EqualFold(lfStr, "stdout") {
+			dest = os.Stdout
+		} else if lfStr != "" {
+			lf, err := os.Create(lfStr)
+			if err == nil {
+				dest = lf
+			} else {
+				fmt.Fprintf(os.Stderr, "Failed to create log file, using stderr: %s\n", err.Error())
+			}
+		}
+		logWriter = logger.NewFileLogger(ll, dest)
+	}
+}
 
 // Response serves as the base for all responses from generated clients. It provides access to the
 // last http.Response.
@@ -208,8 +241,9 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		}
 		return resp, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
-
+	logWriter.WriteRequest(r)
 	resp, err := SendWithSender(c.sender(), r)
+	logWriter.WriteResponse(resp)
 	Respond(resp, c.ByInspecting())
 	return resp, err
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v10.13.0
+
+- Added support for additionalInfo in ServiceError type.
+
 ## v10.12.0
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+## v10.14.0
+
+### New Features
+
+- Added package version that contains version constants and user-agent data.
+
+### Bug Fixes
+
+- Add the user-agent to token requests.
+
 ## v10.13.0
 
 - Added support for additionalInfo in ServiceError type.
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	"github.com/Azure/go-autorest/autorest/date"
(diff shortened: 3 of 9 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/298

autorest/client.go
```diff
@@ -72,6 +74,52 @@ const (
...
+type logLevelType int
+
+const (
+	logLevelNone logLevelType = iota
+	logLevelBasic
+	logLevelFull
+)
+
+const (
+	logTimestamp = "2006-01-02T15:04:05.0000000Z07:00"
```
marstr: I'd add a comment here just to remind any new gophers spinning up on the team that this date/time is semantically important.

autorest/client.go
```diff
@@ -72,6 +74,52 @@ const (
 `
 )
 
+type logLevelType int
+
+const (
+	logLevelNone logLevelType = iota
+	logLevelBasic
+	logLevelFull
```
marstr: If it were me, I'd emulate popular tools like `github.com/sirupsen/logrus` and [loggly](https://www.loggly.com/blog/node-js-libraries-make-sophisticated-logging-simpler/) and have logging levels something like this, I think the difference in expected output is a lot easier to reason about than 'full' vs. 'basic':

- debug
- info
- warn
- error
- fatal
- panic

autorest/client.go
```diff
@@ -72,6 +74,52 @@ const (
...
+		logFile = os.Stderr
+		lfStr := strings.ToLower(os.Getenv("AZURE_GO_AUTOREST_LOG_FILE"))
+		if lfStr == "stdout" {
+			logFile = os.Stdout
+		} else if lfStr != "" {
+			lf, err := os.Create(lfStr)
+			if err == nil {
+				logFile = lf
+			} else {
+				os.Stderr.WriteString(fmt.Sprintf("Failed to create log file, using stderr: %s\n", err.Error()))
```
marstr: [nit] I'd simplify to:

``` Go
fmt.Fprintln(os.Stderr, "Failed to create log file. Using stderr: ", err)
```

autorest/client.go
```diff
@@ -72,6 +74,52 @@ const (
...
+	case "basic":
+		logLevel = logLevelBasic
+	case "full":
+		logLevel = logLevelFull
+	}
+	if logLevel > logLevelNone {
+		// default to stderr
+		logFile = os.Stderr
+		lfStr := strings.ToLower(os.Getenv("AZURE_GO_AUTOREST_LOG_FILE"))
+		if lfStr == "stdout" {
```
marstr: Given that this will need to run in case-sensitive environments, I'd refactor to this:

``` Go
lfStr := os.Getenv("AZURE_GO_AUTOREST_LOG_FILE")
if strings.EqualFold(lfStr, "stdout") {
    logFIle = os.Stdout
} else if lfStr != "" {
    // open/create file and truncate
}
```

autorest/client.go
```diff
@@ -262,3 +311,80 @@ func (c Client) ByInspecting() RespondDecorator {
...
 }
+
+func withLogging() PrepareDecorator {
+	return func(p Preparer) Preparer {
+		return PreparerFunc(func(r *http.Request) (*http.Request, error) {
+			// execute any remaining preparers before logging
+			r, err := p.Prepare(r)
+			if logLevel > logLevelNone {
+				b := &bytes.Buffer{}
+				fmt.Fprintf(b, "(%s) REQUEST: %s %s\n", time.Now().Format(logTimestamp), r.Method, r.URL.String())
```
marstr: I'd add a helper method to prefix a message with the timestamp for you.

autorest/client.go
```diff
@@ -262,3 +311,80 @@ func (c Client) ByInspecting() RespondDecorator {
...
+					fmt.Fprintf(b, "%s: %s\n", k, strings.Join(v, ","))
+				}
+				ct := r.Header.Get(headerContentType)
+				if logLevel == logLevelFull && r.Body != nil && strings.Index(ct, mimeTypeOctetStream) == -1 {
+					// dump body
+					var body []byte
+					body, err = ioutil.ReadAll(r.Body)
+					if err == nil {
+						fmt.Fprintln(b, string(body))
+						if nc, ok := r.Body.(NopClosingReadSeeker); ok {
```
marstr: Since you only need `Seek`, I'd assert just on [io.Seeker](https://godoc.org/io#Seeker).

autorest/client.go
```diff
@@ -262,3 +311,80 @@ func (c Client) ByInspecting() RespondDecorator {
...
+						} else {
+							// recreate the body
+							r.Body = NopClosingReadSeeker{bytes.NewReader(body)}
+						}
+					} else {
+						fmt.Fprintf(b, "failed to read body: %v\n", err)
+					}
+				}
+				logFile.WriteString(b.String())
+				logFile.Sync()
```
marstr: Other than this line, I'd recommend that you use the `log` package in the standard library.

autorest/client.go
```diff
@@ -262,3 +311,80 @@ func (c Client) ByInspecting() RespondDecorator {
...
+				b := &bytes.Buffer{}
+				fmt.Fprintf(b, "(%s) RESPONSE: %d %s\n", time.Now().Format(logTimestamp), resp.StatusCode, resp.Request.URL.String())
+				// dump headers
+				for k, v := range resp.Header {
+					fmt.Fprintf(b, "%s: %s\n", k, strings.Join(v, ","))
+				}
+				ct := resp.Header.Get(headerContentType)
+				if logLevel == logLevelFull && resp.Body != nil && strings.Index(ct, mimeTypeOctetStream) == -1 {
+					// dump body
+					defer resp.Body.Close()
```
marstr: I think that this will end up calling `Close()` on the wrong Body, as you'll want to close the original after the `ioutil.ReadAll` a few lines below. Using `defer`, Close will get called on the reassigned `NopClosingReadSeeker`.

autorest/client.go
jhendrixMSFT: I had originally used the log package but I didn't see any way to flush the underlying stream.  Can it be done?

autorest/client.go
jhendrixMSFT: "A deferred function's arguments are evaluated when the defer statement is evaluated."
See https://blog.golang.org/defer-panic-and-recover

autorest/client.go
marstr: I don't think so :/ I agree with your decision here. I was just trying to leave a comment here in case anybody else wondered why you didn't use it.

jhendrixMSFT: Good point, I'll add a comment explaining why I used *os.File instead of a logger.

autorest/client.go
marstr: Ah, my bad! I guess I had this one wrong :)

For anyone else who's curious I wrote up a little example: https://play.golang.org/p/nJD9JoAgDti

jhendrixMSFT: No worries I did pretty much the exact same exercise then found this link.

autorest/client.go
jhendrixMSFT: I do have a call to ```strings.ToLower``` to avoid case-sensitivity however I think ```strings.EqualFold``` is a better choice.

marstr: What I was getting at is that using `strings.ToLower` means that the original casing provided by the author gets lost when you `os.Create` using `lfStr` later. :)

jhendrixMSFT: Good point!

autorest/client.go
jhendrixMSFT: Derp, and I was wracking my brain as I really didn't want to add ```NopClosingReadSeeker```.  I guess the obvious answers are really hard to find sometimes.  :p

autorest/client.go
jhendrixMSFT: I like this idea.  How about we rename ```logLevelBasic``` to ```info```?  I'm debating if we should just include the request/response bodies in that or keep it separate, perhaps under ```debug```.  I feel like if I want to log requests/responses I'd want them in full although I can see how they could be quite large.  Thoughts?

jhendrixMSFT: Or we put all of this under ```debug```?

autorest/client.go
marstr: I've been there before, it's incredible what other people's eyes see. 😝

autorest/client.go
marstr: I like the way you're thinking. I'd keep bodies under `debug` to save space, and the fact that a request, response was seen under `info` unless you encounter some reason why that really doesn't make sense.

jhendrixMSFT: Agreed.  I'm tempted to include headers under ```debug``` as well.  This way if I just want to see request/response times I don't have to mentally tease out the header info I don't care about.

jhendrixMSFT: We should also follow the logging semantics defined in the [pipeline](https://github.com/Azure/azure-pipeline-go/blob/master/pipeline/core.go) so that in vNext it cleanly translates.

jhendrixMSFT:
Closing for now as I'm refactoring this.

Issue #300, opened 2018-07-09
add user-agent to IMDS/MSI calls


joshgav:
https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L645

Can we add a user-agent string identifying the Go SDK to the above ^ HTTP request? Is it included some other way? This is to help us review telemetry from the IMDS API. Thanks!

joshgav:
Let's use "go-autorest/version".

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v10.14.0

PR #302, opened 2018-07-19
Include user-agent info in token requests

jhendrixMSFT:
Add the standard go-autorest user-agent string to token requests.  This
required moving version info to its own package to avoid a circular
dependency.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -23,12 +23,15 @@ import (
 const Number = "v10.14.0"
 
 var (
-	// UserAgent builds a string containing the Go version, system archityecture and OS,
-	// and the go-autorest version.
-	UserAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
+	userAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/300

version/version.go
```diff
@@ -0,0 +1,34 @@
...
+	"runtime"
+)
+
+// Number contains the semantic version of this SDK.
+const Number = "v10.14.0"
+
+var (
+	// UserAgent builds a string containing the Go version, system archityecture and OS,
+	// and the go-autorest version.
+	UserAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",
```
jhendrixMSFT: Hmm I'm thinking that this should not be exported and instead expose the value via a func.

vladbarosan: so people can't change it ?

jhendrixMSFT: Yeah.  We already provide a facility for adding to the user-agent but we probably don't want folks to have the ability to muck about with the default info.

PR #303, opened 2018-07-20
Add support for request/response logging

jhendrixMSFT:
Request and response logging can now be enabled by setting the
environment variable AZURE_GO_SDK_LOG_LEVEL.  See the notes in the
change log for more information.
This logger is general-purpose so other data can be logged as required.

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -22,8 +22,10 @@ import (
 	"log"
 	"net/http"
 	"net/http/cookiejar"
+	"strings"
 	"time"
 
+	"github.com/Azure/go-autorest/logger"
 	"github.com/Azure/go-autorest/version"
 )
 
@@ -208,8 +210,17 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		}
 		return resp, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
-
+	logger.Instance.WriteRequest(r, logger.Filter{
+		Header: func(k string, v []string) (bool, []string) {
+			// remove the auth token from the log
+			if strings.EqualFold(k, "Authorization") || strings.EqualFold(k, "Ocp-Apim-Subscription-Key") {
+				v = []string{"**REDACTED**"}
+			}
+			return true, v
+		},
+	})
 	resp, err := SendWithSender(c.sender(), r)
+	logger.Instance.WriteResponse(resp, logger.Filter{})
 	Respond(resp, c.ByInspecting())
 	return resp, err
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,0 +1,328 @@
+package logger
+
+// Copyright 2017 Microsoft Corporation
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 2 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/298
This is a new iteration of https://github.com/Azure/go-autorest/pull/301

jhendrixMSFT:
I'm thinking about indenting the header/body content so that it's a easier to read the log file.  Thoughts?
```
(2018-07-25T13:43:23.1405564-07:00) INFO: REQUEST: PUT https://management.azure.com/subscriptions/6d3860f6-8a11-431d-b3fa-1b3c4a8b888a/resourcegroups/jhendrixmsirg1?api-version=2018-05-01
  Content-Type: application/json; charset=utf-8
  User-Agent: Go/go1.10 (amd64-windows) go-autorest/v10.15.0 Azure-SDK-For-Go/v19.0.0 resources/2018-05-01
  Authorization: **REDACTED**
  {"location":"WestUS"}
```

logger/logger.go
```diff
@@ -0,0 +1,332 @@
...
+
+func initDefaultLogger() {
+	// init with nilLogger so callers don't have to do a nil check on Default
+	Default = nilLogger{}
+	llStr := strings.ToLower(os.Getenv("AZURE_GO_AUTOREST_LOG_LEVEL"))
+	if llStr == "" {
+		return
+	}
+	ll, err := ParseLevel(llStr)
+	if err != nil {
```
vladbarosan: i wonder if its the best thing to just bail in case the log level  is malformed. or at least reported in the log file also if that is specified.

logger/logger.go
```diff
@@ -0,0 +1,332 @@
...
+		return
+	}
+	// default to stderr
+	dest := os.Stderr
+	lfStr := os.Getenv("AZURE_GO_AUTOREST_LOG_FILE")
+	if strings.EqualFold(lfStr, "stdout") {
+		dest = os.Stdout
+	} else if lfStr != "" {
+		lf, err := os.Create(lfStr)
+		if err == nil {
```
vladbarosan: this will truncate the file. maybe specify in the changelog that this is the case if they use a file

logger/logger.go
jhendrixMSFT: I wanted to avoid creating a file if specified if the log level is None.  We do write the error to stderr though.  Or are you suggesting we set it to something other than None if it's malformed?

logger/logger.go
jhendrixMSFT: Good point will fix.

autorest/client.go
```diff
@@ -208,8 +210,17 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 		}
 		return resp, NewErrorWithError(err, "autorest/Client", "Do", nil, "Preparing request failed")
 	}
-
+	logger.Default.WriteRequest(r, logger.Filter{
+		Header: func(k string, v []string) (bool, []string) {
+			// remove the auth token from the log
+			if strings.ToLower(k) == "authorization" {
```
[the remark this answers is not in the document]
vladbarosan: although that CS authorizer I think shouldn't be here, just api key authorizer would have been fine but that's that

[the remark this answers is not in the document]
jhendrixMSFT: But we still need to redact this header value right?

PR #305, opened 2018-08-03
Fixes for LROs

jhendrixMSFT:
If an LRO returns a failure in the initial response return an error
object at that point so the caller doesn't have to poll.
For failed LROs without an OData v4 error include the response body in
the error's AdditionalInfo field to aid in diagnosing the failure.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -348,6 +348,10 @@ func (pt *pollingTrackerBase) initializeState() error {
 	case http.StatusOK:
 		if ps := pt.getProvisioningState(); ps != nil {
 			pt.State = *ps
+			if pt.hasFailed() {
+				pt.updateErrorFromResponse()
+				return pt.pollingError()
+			}
 		} else {
 			pt.State = operationSucceeded
 		}
@@ -364,6 +368,7 @@ func (pt *pollingTrackerBase) initializeState() error {
 	default:
 		pt.State = operationFailed
 		pt.updateErrorFromResponse()
+		return pt.pollingError()
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Workaround to address https://github.com/Azure/azure-sdk-for-go/issues/2179 until the RP returns an OData v4 error.

autorest/azure/async.go
```diff
@@ -422,6 +427,7 @@ func (pt *pollingTrackerBase) pollForStatus(sender autorest.Sender) error {
 
 // attempts to unmarshal a ServiceError type from the response body.
 // if that fails then make a best attempt at creating something meaningful.
+// NOTE: this assumes that the async operation has failed.
 func (pt *pollingTrackerBase) updateErrorFromResponse() {
```
vladbarosan: >updateErrorFromResponse [](start = 30, length = 23)

can we return an error if it doesnt make sense to call this ?

autorest/azure/async.go
```diff
@@ -348,6 +348,10 @@ func (pt *pollingTrackerBase) initializeState() error {
 	case http.StatusOK:
 		if ps := pt.getProvisioningState(); ps != nil {
 			pt.State = *ps
+			if pt.hasFailed() {
+				pt.updateErrorFromResponse()
+				return pt.pollingError()
```
vladbarosan: was wondering is it possible to just have this part of pollingError() call ? i.e. are there situations where we odnt want to call `updateErrorFromResponse` when we call pollingError ? and if so is that discernable based on `pollingTrackerBase` state ?

jhendrixMSFT: It probably could be refactored (it's likely there was more differing state during the initial implementation) but given the finicky nature of the LRO state machine I'm a little hesitant to change it since things are working as is.

PR #311, opened 2018-08-30
Init LRO polling URL and method

jhendrixMSFT:
Always initialize the polling URL and method, favoring the
Azure-AsyncOperation header.  Each tracker will update the values as
required.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -119,6 +119,9 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 	if err := f.pt.updatePollingState(f.pt.provisioningStateApplicable()); err != nil {
 		return false, err
 	}
+	if err := f.pt.initHeaders(); err != nil {
+		return false, err
+	}
 	if err := f.pt.updateHeaders(); err != nil {
 		return false, err
 	}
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2179

autorest/azure/async.go
```diff
@@ -274,6 +277,10 @@ type pollingTracker interface {
 
 	// methods common to all trackers
 
+	// initializes a tracker's polling URL and method, called for each iteration.
+	// these values can be overridden by each polling tracker as required.
+	initHeaders() error
```
vladbarosan: >initHeaders [](start = 1, length = 11)

isn't this more like init form headers ?

jhendrixMSFT: Yep the naming isn't the best but since we already have ```updateHeaders{}``` I wanted to keep this matching as the two are kinda related.

vladbarosan: eh these are not exported, can we rename them ?

autorest/azure/async.go
```diff
@@ -119,6 +119,9 @@ func (f *Future) Done(sender autorest.Sender) (bool, error) {
 	if err := f.pt.updatePollingState(f.pt.provisioningStateApplicable()); err != nil {
 		return false, err
 	}
+	if err := f.pt.initHeaders(); err != nil {
+		return false, err
+	}
 	if err := f.pt.updateHeaders(); err != nil {
```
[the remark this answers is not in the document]
vladbarosan: hmm ok, I was hoping that acutally we can contain all the logic in that method so we dont have to duplicate the logic.

autorest/azure/async.go
jhendrixMSFT: Good point, fixed.

PR #313, opened 2018-09-04
Allow a new authorizer to be created from a configuration file by specifying a resource instead of a base url.

nphmuller:
This enables resource like KeyVault and Container Registry to use an authorizer configured from a configuration file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ _obj
 _test
 .DS_Store
 .idea/
+.vscode/
 
 # Architecture specific extensions/prefixes
 *.[568vq]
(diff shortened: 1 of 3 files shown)


marstr:
This is related to https://github.com/Azure/azure-sdk-for-go/issues/2610

nphmuller:
@marstr 
It's related to that issue, but it's not the solution to it.

This PR just allows the NewAuthorizerFromFile flow to be used with every resource. 

Before the changes in this PR the only way to use this flow was by providing a baseurl, which was used to [determine the resource url](https://github.com/Azure/go-autorest/blob/39469b0159db4a8540fd742779396d3884cb543c/autorest/azure/auth/auth.go#L232-L253). Downside was that some resources like KeyVault cannot be determined this way (afaik), so I've basically provided an overload where the user can provide the resource url directly. 

This overload already exists for the NewAuthorizerFromEnvironment flow.

PR #315, opened 2018-09-05
[WIP] Using the Context from the timeout if provided

tombuildsstuff:
- Using the timeout from the context if available
- Makes PollingDuration optional/nilable

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -164,9 +164,17 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
 // running operation has completed, the provided context is cancelled, or the client's
 // polling duration has been exceeded.  It will retry failed polling attempts based on
 // the retry value defined in the client up to the maximum retry attempts.
-func (f *Future) WaitForCompletionRef(ctx context.Context, client autorest.Client) error {
-	ctx, cancel := context.WithTimeout(ctx, client.PollingDuration)
-	defer cancel()
+func (f *Future) WaitForCompletionRef(inputCtx context.Context, client autorest.Client) error {
+
+	var ctx context.Context
+	if d := client.PollingDuration; d != nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(inputCtx, *d)
+		defer cancel()
+	} else {
+		ctx = inputCtx
+	}
+
 	done, err := f.Done(client)
 	for attempts := 0; !done; done, err = f.Done(client) {
 		if attempts >= client.RetryAttempts {
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -167,9 +167,9 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
 func (f *Future) WaitForCompletionRef(inputCtx context.Context, client autorest.Client) error {
 
 	var ctx context.Context
-	if d := client.PollingDuration; d != nil {
+	if d := client.PollingDuration; d != 0 {
 		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(inputCtx, *d)
+		ctx, cancel = context.WithTimeout(inputCtx, d)
 		defer cancel()
 	} else {
 		ctx = inputCtx
--- a/autorest/azure/rp.go
+++ b/autorest/azure/rp.go
@@ -141,7 +141,7 @@ func register(client autorest.Client, originalReq *http.Request, re RequestError
 
 	// poll for registered provisioning state
 	now := time.Now()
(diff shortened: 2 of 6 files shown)


autorest/client.go
```diff
@@ -153,7 +153,8 @@ type Client struct {
 	PollingDelay time.Duration
 
 	// PollingDuration sets the maximum polling time after which an error is returned.
-	PollingDuration time.Duration
+	// if nil, the timeout from the Context is used
+	PollingDuration *time.Duration
```
marstr: Instead of changing the type of PollingDuration to a pointer type, I'd recommend just checking for the zero value. That will prevent a breaking change to public surface area, and protect people from having an unhelpful instant timeout.

autorest/azure/async.go
```diff
@@ -164,9 +164,17 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
 // running operation has completed, the provided context is cancelled, or the client's
 // polling duration has been exceeded.  It will retry failed polling attempts based on
 // the retry value defined in the client up to the maximum retry attempts.
-func (f *Future) WaitForCompletionRef(ctx context.Context, client autorest.Client) error {
-	ctx, cancel := context.WithTimeout(ctx, client.PollingDuration)
-	defer cancel()
+func (f *Future) WaitForCompletionRef(inputCtx context.Context, client autorest.Client) error {
+
+	var ctx context.Context
+	if d := client.PollingDuration; d != nil {
```
marstr: Related to my comment below, suggesting that client.PollingDuration not become a pointer, this would become:
``` Go
var ctx context.Context
if d := client.PollingDuration; d != 0 {
    var cancel context.CancelFunc
    ctx, cancel = context.WithTimeout(inputCtx, d)
    defer cancel()
} else {
    ctx = inputCtx
}
```

autorest/client.go
marstr: On another note, it's not clear to me which value this runtime should assume, should PollingDuration not be provided:
1. [`time.Duration(math.MaxInt64) // Essentially indefinite, just over 290 years`](https://play.golang.org/p/V6QGlYzVMOv)
1. `autorest.DefaultPollingDuration // A reasonable amount of time to poll for the average Azure operation.`

Maybe you could accomplish much of the functionality of making this value optional by just making it easier to specify a polling duration that is essentially indefinite?

autorest/azure/rp.go
```diff
@@ -141,7 +141,7 @@ func register(client autorest.Client, originalReq *http.Request, re RequestError
 
 	// poll for registered provisioning state
 	now := time.Now()
-	for err == nil && time.Since(now) < client.PollingDuration {
+	for err == nil && (client.PollingDuration == nil || (client.PollingDuration != nil && time.Since(now) < *client.PollingDuration)) {
```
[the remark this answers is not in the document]
tombuildsstuff: sounds good 👍

autorest/client.go
tombuildsstuff: > Instead of changing the type of PollingDuration to a pointer type, I'd recommend just checking for the zero value. That will prevent a breaking change to public surface area, and protect people from having an unhelpful instant timeout.

It would - however I feel by making this a zero/default value this is exposing the internals behaviour of autorest to callers; rather than having this as a nil-pointer which (IMO) is clearer (but obviously would need to be made in future)?

autorest/azure/async.go
tombuildsstuff: (updated)

autorest/azure/async.go
```diff
@@ -164,9 +164,17 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
 // running operation has completed, the provided context is cancelled, or the client's
 // polling duration has been exceeded.  It will retry failed polling attempts based on
 // the retry value defined in the client up to the maximum retry attempts.
-func (f *Future) WaitForCompletionRef(ctx context.Context, client autorest.Client) error {
-	ctx, cancel := context.WithTimeout(ctx, client.PollingDuration)
-	defer cancel()
+func (f *Future) WaitForCompletionRef(inputCtx context.Context, client autorest.Client) error {
+
+	var ctx context.Context
+	if d := client.PollingDuration; d != 0 {
```
jhendrixMSFT: In hindsight this method should never have relied on these client settings, I should have only used them in the legacy implementation ```DoPollForAsynchronous```.  :(  While we could make such a change in a non-breaking way it would be a change in behavior that might cause unexpected problems.

jhendrixMSFT:
@marstr do you have any other comments on this?

marstr:
Nope, no further input, @jhendrixMSFT!

PR #316, opened 2018-09-06
Authorizer which uses Azure CLI to obtain the token

nphmuller:
This PR adds the `NewAuthorizerFromCli` Authorizer factory method, which uses Azure CLI to obtain an access token for the currently logged in user. The implementation is inspired by the [Azure SDK for .NET implementation](https://github.com/Azure/azure-sdk-for-net/blob/b38512ff94eb22446b925f6af2c2bc73207c55a4/src/SdkCommon/AppAuthentication/Azure.Services.AppAuthentication/TokenProviders/AzureCliAccessTokenProvider.cs). These changes would resolve https://github.com/Azure/azure-sdk-for-go/issues/2610.

I've tested the changes on Windows 10 (1803) and Ubuntu 16.04 (through WSL).

There are currently no additional test, because they would depend on an Azure CLI installation with a logged in user. I'm not sure how to mock this, so any pointers are really appreciated if the tests are necessary.

### DIFF
--- a/autorest/azure/auth/auth.go
+++ b/autorest/azure/auth/auth.go
@@ -31,6 +31,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/azure/cli"
 	"github.com/dimchansky/utfbom"
 	"golang.org/x/crypto/pkcs12"
 )
@@ -165,6 +166,35 @@ func NewAuthorizerFromFileWithResource(resource string) (autorest.Authorizer, er
 	}
 
 	return autorest.NewBearerAuthorizer(spToken), nil
+}
+
+// NewAuthorizerFromCLI creates an Authorizer configured from Azure CLI 2.0 for local development scenarios.
+func NewAuthorizerFromCLI() (autorest.Authorizer, error) {
+	settings, err := getAuthenticationSettings()
+	if err != nil {
+		return nil, err
+	}
+
+	if settings.resource == "" {
+		settings.resource = settings.environment.ResourceManagerEndpoint
+	}
+
+	return NewAuthorizerFromCLIWithResource(settings.resource)
+}
+
+// NewAuthorizerFromCLIWithResource creates an Authorizer configured from Azure CLI 2.0 for local development scenarios.
+func NewAuthorizerFromCLIWithResource(resource string) (autorest.Authorizer, error) {
+	token, err := cli.GetTokenFromCLI(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	adalToken, err := token.ToADALToken()
+	if err != nil {
+		return nil, err
+	}
+
+	return autorest.NewBearerAuthorizer(&adalToken), nil
 }
 
 func getAuthFile() (*file, error) {
--- a/autorest/azure/cli/token.go
+++ b/autorest/azure/cli/token.go
@@ -15,9 +15,13 @@ package cli
 //  limitations under the License.
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
+	"regexp"
+	"runtime"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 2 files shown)


nphmuller:
@marstr 
I know you're probably busy, but do you have any idea when you're able to take a look at this PR?
Any feedback is really appreciated! :)

jhendrixMSFT:
@nphmuller sorry for the delay on this, will review next Monday.

autorest/azure/auth/auth.go
```diff
@@ -167,6 +168,35 @@ func NewAuthorizerFromFileWithResource(resource string) (autorest.Authorizer, er
 	return autorest.NewBearerAuthorizer(spToken), nil
 }
 
+// NewAuthorizerFromCli creates an Authorizer configured from Azure CLI 2.0 for local development scenarios.
+func NewAuthorizerFromCli() (autorest.Authorizer, error) {
```
jhendrixMSFT: Nit, `Cli` should be `CLI` (comment applies to all instances.).

autorest/azure/cli/token.go
```diff
@@ -112,3 +116,55 @@ func LoadTokens(path string) ([]Token, error) {
...
 	return tokens, nil
 }
+
+// GetTokenFromCli gets a token using Azure CLI 2.0 for local development scenarios.
+func GetTokenFromCli(resource string) (*Token, error) {
+	// This is the path that a develop can set to tell this class what the install path for Azure CLI is.
+	const azureCliPath = "AzureCLIPath"
+
+	// The default install paths are used to find Azure CLI. This is for security, so that any path in the calling program's Path environment is not used to execute Azure CLI.
+	azureCliDefaultPathWindows := fmt.Sprintf("%s\\Microsoft SDKs\\Azure\\CLI2\\wbin; %s\\Microsoft SDKs\\Azure\\CLI2\\wbin", os.Getenv("ProgramFiles(x86)"), os.Getenv("ProgramFiles"))
```
jhendrixMSFT: Can users install the CLI to a non-default path?  If so then this would be broken.
CC @lmazuel

jhendrixMSFT: I guess the solution here is the `AzureCLIPath` environment var yes?

autorest/azure/cli/token.go
```diff
@@ -112,3 +116,55 @@ func LoadTokens(path string) ([]Token, error) {
 
 	return tokens, nil
 }
+
+// GetTokenFromCli gets a token using Azure CLI 2.0 for local development scenarios.
+func GetTokenFromCli(resource string) (*Token, error) {
+	// This is the path that a develop can set to tell this class what the install path for Azure CLI is.
+	const azureCliPath = "AzureCLIPath"
```
jhendrixMSFT: To be consistent with the other env vars this value should be `AZURE_CLI_PATH`.

autorest/azure/cli/token.go
yugangw-msft: CLI's doesn't let you select installation path. The logic here is fine. 
Because CLI's installer ensures global PATH env var gets updated for its executable, maybe the fallback would be just run "az" w/o the full path?

jhendrixMSFT: The idea here is to limit what's on the path so that something doesn't hijack the `az` command.

nphmuller: @jhendrixMSFT You're right. `AzureCLIPath` can be set to override the default path.

The .NET Azure SDK team considered PATH hijacking a sufficient risk to not use it.
Simply using PATH _would_ simplify a bit because there would no longer be any os-specific logic.

Should I leave it as is, or simplify the code to simply use the az binary in PATH?

autorest/azure/cli/token.go
nphmuller: The .NET SDK uses the `AzureCLIPath` env var, but I'd be happy to change this. Should I?

autorest/azure/cli/token.go
```diff
@@ -112,3 +116,55 @@ func LoadTokens(path string) ([]Token, error) {
...
+
+	// Execute Azure CLI to get token
+	var cliCmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cliCmd = exec.Command(fmt.Sprintf("%s\\system32\\cmd.exe", os.Getenv("windir")))
+		cliCmd.Env = os.Environ()
+		cliCmd.Env = append(cliCmd.Env, fmt.Sprintf("PATH=%s;%s", os.Getenv(azureCliPath), azureCliDefaultPathWindows))
+		cliCmd.Args = append(cliCmd.Args, "/c")
+	} else {
+		cliCmd = exec.Command("/bin/bash")
```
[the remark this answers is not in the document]
nphmuller: The .NET SDK uses bash instead of the default, but I'd be happy to change this. Should I?

autorest/azure/auth/auth.go
nphmuller: Done!

nphmuller:
Implemented some feedback and rebased.

autorest/azure/cli/token.go
jhendrixMSFT: I'd leave it as-is to avoid PATH hijacking.

autorest/azure/cli/token.go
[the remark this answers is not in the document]
vladbarosan: If you want consistency you can leave it as a first option but would add at least a fallback in case its not present.

autorest/azure/cli/token.go
jhendrixMSFT: I see, I wasn't aware of that.  Let's keep it as-is so it's consistent across the SDKs.

autorest/azure/cli/token.go
nphmuller: Ok! Marking this as resolved then.

autorest/azure/cli/token.go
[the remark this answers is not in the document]
nphmuller: Might as well only use $SHELL then. :)
Done!

nphmuller:
Implemented all feedback and manually retested on Windows 10 (1809) and Ubuntu 18.04 (through WSL).
Thanks for the feedback everyone!

PR #317, opened 2018-09-07
Handle ContentLength equals -1 case

seyadava:
Reference Issue: #318

When HTTP response status is NoContent (204) ContentLenght of the HTTP response is -1. Code checks if ContentLenght != 0 then tries to parse the response, which fails in case ContentLength is -1. 

I modified the code to parse response only if ContentLenght > 0

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -427,6 +427,10 @@ func (pt *pollingTrackerBase) updateRawBody() error {
 		if err != nil {
 			return autorest.NewErrorWithError(err, "pollingTrackerBase", "updateRawBody", nil, "failed to read response body")
 		}
+		// observed in 204 responses over HTTP/2.0; the content length is -1 but body is empty
+		if len(b) == 0 {
+			return nil
+		}
 		// put the body back so it's available to other callers
 		pt.resp.Body = ioutil.NopCloser(bytes.NewReader(b))
 		if err = json.Unmarshal(b, &pt.rawBody); err != nil {
@@ -471,7 +475,7 @@ func (pt *pollingTrackerBase) updateErrorFromResponse() {
 		re := respErr{}
 		defer pt.resp.Body.Close()
 		var b []byte
-		if b, err = ioutil.ReadAll(pt.resp.Body); err != nil {
+		if b, err = ioutil.ReadAll(pt.resp.Body); err != nil || len(b) == 0 {
 			goto Default
 		}
 		if err = json.Unmarshal(b, &re); err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v11.2.6
+
+### Bug Fixes
+
+- If zero bytes are read from a polling response body don't attempt to unmarshal them.
+
 ## v11.2.5
 
 ### Bug Fixes
--- a/version/version.go
+++ b/version/version.go
@@ -20,7 +20,7 @@ import (
 )
 
 // Number contains the semantic version of this SDK.
-const Number = "v11.2.5"
+const Number = "v11.2.6"
 
 var (
 	userAgent = fmt.Sprintf("Go/%s (%s-%s) go-autorest/%s",


knithinc:
@marstr @joshgav , could you guys please take a look at this PR? We had this issue #318 reported by Terraform.

marstr:
Howdy @seyadava,

Help me understand this scenario. To me, a negative content-length value is not clearly indicative of an empty body. To address #318 directly, why is it that ARM is returning a negative value in this situation?

seyadava:
Hey @marstr 
When I look at the raw response from ARM template I don't see content-length in their response header for 204 status: 

2018/09/06 16:26:00 [DEBUG] azurestack Response for https://management.westus.stackpoc.com/subscriptions/5d76acf0-3d6d-4e96-934f-69acfe079d65/operationresults/eyJqb2JJZCI6IkRlcGxveW1lbnREZWxldGlvbkpvYi1HTlMtQUNDVEVTVFJHOjJEMzQ1NjcwMjA4OTEyNzU2MzMwMC1BQ0NURVNUVEVNUExBVEU6MkQzNDU2NzAyMDg5MTI3NTYzMzAwLSIsImpvYkxvY2F0aW9uIjoid2VzdHVzIn0?api-version=2016-02-01: 

HTTP/2.0 204 No Content
Connection: close
Cache-Control: no-cache
Date: Thu, 06 Sep 2018 21:25:18 GMT
Expires: -1
Pragma: no-cache
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Content-Type-Options: nosniff
X-Ms-Correlation-Request-Id: 12090e43-e4d0-4639-8c8c-b4b2dde2b3d2
X-Ms-Ratelimit-Remaining-Subscription-Reads: 14973
X-Ms-Request-Id: 12090e43-e4d0-4639-8c8c-b4b2dde2b3d2
X-Ms-Routing-Request-Id: WESTUS:20180906T212519Z:12090e43-e4d0-4639-8c8c-b4b2dde2b3d2

However, when I look at the parsed response on Golang I can see ContentLength is set to -1. So I think somewhere in the code ContentLength value is defaulted to -1. Here is the parsed response:
image

I didn't find default value for ContentLength on ARM specs, but I saw in golang http response class (i.e. net/http/response.go) ContentLength -1 means the content length is unknown. So I guess maybe if ContentLength header is not set, golang defaults the value of ContentLength to -1. 

 Response.go class definition for content length: (https://golang.org/pkg/net/http/#Response)
// ContentLength records the length of the associated content. The
// value -1 indicates that the length is unknown. Unless Request.Method
// is "HEAD", values >= 0 indicate that the given number of bytes may
// be read from Body.
ContentLength int64

seyadava:
@marstr I noticed we do the same check (i.e. contentLength != 0) in some other places on the same file (async.go) . Since -1 is a legit value for contentLenght, shouldn't we handle it?

jhendrixMSFT:
I'm not convinced this is the correct fix.  Can you please provide a complete repro of this issue?  I suspect there's more to it.

marstr:
> Since -1 is a legit value for contentLenght, shouldn't we handle it?

Yes, I totally agree. I've spent a little more time looking at this, it looks like we need to go a little further than just updating the check. Could you please update your PR so that it still reads to completion the body of the request when `-1` is provided? It should read it into a buffer and update the Body of the Response, and unmarshal JSON only if there is contents to the response body.

If you don't have time for that, we can add this fix to our backlog and we can write-up the fix.

jhendrixMSFT:
What's in the response body in this case though?  Usually a -1 means "here's some content but I don't know how long it is".  I find it suspicious that we're getting some kind of response that isn't JSON.

jhendrixMSFT:
@seyadava is there any update here?

seyadava:
@jhendrixMSFT sorry for my delayed response to this PR. I was buried with higher priority tasks in the past weeks. I will update this PR to handle unknown contentLenght (-1) gracefully. I will send the update this week.

jhendrixMSFT:
@seyadava no worries on the delay, but before you update the PR we need more information as I'm not convinced this is the correct fix.  Do you have a repro available?

seyadava:
[gosample.txt](https://github.com/Azure/go-autorest/files/2435732/gosample.txt)

[repro.txt](https://github.com/Azure/go-autorest/files/2435733/repro.txt)

@jhendrixMSFT I could repro the issue (see the attached file). You can find the error at the end of the file. I have also attached my sample code that repro's the issue. 
The problem is that content-length is an optional header and if it is not populated by the endpoint then golang defaults it to -1 which means content length is unknown. In this case go-autorest tries to read the content and when the content is empty go-autorest cannot unmarshall the json and fails. Please let me know if you need more details. 
Once we agree on the problem, I will start coding the fix. My plan to fix the issue is what @marstr has suggested. When content length is unknown we need to figure out whether the content is empty or there is actually something in the content. We should try to unmarshall content only if it is not empty.

jhendrixMSFT:
In addition, since the response is a 204 (StatusNoContent) one could make the argument that we shouldn't even attempt to read the response body as there shouldn't be one; in that case I would expect the content length to be zero which is what I observe when running the repro.

marstr:
What's the current status of this PR?

seyadava:
Adding @knithinc 

I tried to repro this issue on public Azure and noticed this problem does not occur on public Azure. However, it repros on Azure Stack (on our test environments as well as customer's environment). 
The only difference I found between ARM endpoint responses on Azure and Azure Stack was HTTP version. HTTP version of ARM response on Azure is 1.1 but ARM endpoint on Azure Stack uses HTTP 2.0 version. I guess if content-length header is not populated, HTTP 2.0 defaults its value to (-1) but HTTP 1.1 defaults it to 0.

jhendrixMSFT:
Interesting.  If you read the body via `ioutil.ReadAll` what is the `len` of the bytes returned?  If it's zero (presumably it would be) then I think it would be best to make the fix based on a zero-length read.

seyadava:
ioutil.ReadAll length is zero. In fact because response body is empty json package fails to unmarshal response.

seyadava:
I updated this PR to check response length before unmarshaling response body.

autorest/azure/async.go
```diff
@@ -444,24 +446,26 @@ func (pt *pollingTrackerBase) updateErrorFromResponse() {
...
-			goto Default
-		}
-		if err = json.Unmarshal(b, &re); err != nil {
-			goto Default
-		}
-		// unmarshalling the error didn't yield anything, try unwrapped error
-		if re.ServiceError == nil {
-			err = json.Unmarshal(b, &re.ServiceError)
-			if err != nil {
+		if len(b) != 0 {
```
jhendrixMSFT: This is incorrect, you're checking the length before reading from the body.

jhendrixMSFT:
@seyadava can you also rebase this against the dev branch?

autorest/azure/async.go
seyadava: Oops. You are right. Moved if statement to the right place

PR #319, opened 2018-09-10
LRO fix for polling failure

jhendrixMSFT:
If a polling operation returns a failure status code return the
associated error.

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -428,6 +428,7 @@ func (pt *pollingTrackerBase) pollForStatus(sender autorest.Sender) error {
 	} else {
 		// check response body for error content
 		pt.updateErrorFromResponse()
+		err = pt.pollingError()
 	}
 	return err
 }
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2645

autorest/azure/async.go
```diff
@@ -428,6 +428,7 @@ func (pt *pollingTrackerBase) pollForStatus(sender autorest.Sender) error {
 	} else {
 		// check response body for error content
 		pt.updateErrorFromResponse()
+		err = pt.pollingError()
```
vladbarosan: to be clear, it is still valid for the polling to fail and not return an error ? is there any error format prescribed on polling ?

jhendrixMSFT: No, if the polling fails it should now always return a non-nil error.  If the service returns an error it should be in OData v4 format.

vladbarosan: aha ok. is this documented for polling errors somewhere ?

jhendrixMSFT: Yes see https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#operation-resource-format

PR #320, opened 2018-09-12
Pin utfbom and go-homedir to v1.0.0 and remove the explicit dependency on x/crypto master

ultimateboy:
I ran into some issues today with a build pipeline that has this project as a dependency. Some changes to `x/crypto` weren't valid with some of my other dependencies, and the explicit pinning to master here causes pain upstream (having to use dep `overrides`). I don't see any reason why go-autorest needs to always be pinned to master, so I've removed this - `dep` will use the most recent valid version it can.

While I was at it, I also added specific versions (`v1.0.0`) for utfbom and go-homedir.

None of these changes actually change the dependencies in any way, as seen with the simple diff on the `Gopkg.lock` file. This just helps other repos which import this project.

As for checking the boxes that this is an "urgent bug fix" and warrants its own release, that's up to you - I altered the changelog in a separate commit. But given this is dealing with how projects import go-autorest, I think it would be prudent to tag a patch release with these changes.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v10.15.5
+
+- Pin utfbom and go-homedir dependencies to v1.0.0 and remove the explicit dependency on x/crypto master.
+
 ## v10.15.4
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Sorry for the delay on this, I've rolled your changes into https://github.com/Azure/go-autorest/pull/326.

Issue #325, opened 2018-09-26
[AzureStack][ADFS] Unable to use go SDK against AzureStack ADFS environment as Token struct defines ExpiresIn as string


bganapa:
In ADFS, the value of expires_in is returned as integer, in case of AAD this is returned as string. The actual value of ExpiresIn is always numeral. So it makes sense to define this as int in SDK.  

The following line defines the ExpiresIn as string in go-autorest adal
https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L104

If we define the above as int, it works on ADFS.

Adal-dotnet defines this field as long 
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs#L94

and does long.Parse
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Internal/OAuth2/TokenResponse.cs#L143

Issue #314, opened 2018-09-05
Context is only partially supported through the Prepare->Send->Respond pipeline


tombuildsstuff:
👋 

I’ve been working through adding support for custom resource timeouts to Terraform - which means we’ve been adding Context to every field.

During this process I’ve noticed that Go-AutoRest doesn’t handle context properly - and only does so within the “Prepare” method (whereas Sender and Responder means this doesn’t work as expected). The current behaviour is to add the `PollingDuration` field (set at the Client level) to the passed in context and run with that. 

Unfortunately we’re unable to use the `PollingDuration` field as it’s scoped to the client, which means since we reuse the clients (to reuse auth tokens) - setting the `PollingDuration` at this level means it’s not possible to reuse clients, since there’s the potential for conflicts where different callers have different timeouts (and thus Caller B’s timeout impacts Caller A).

Taking a look at the code, I believe it’s possible to solve this by making the `PollingDuration` field on the client nilable (read: a pointer) - which means nil-checks could be used to either use the timeout from the context cancelling (if it’s nil) - or to use the current behaviour (if a timeout is specified). This allows users of the SDK to pass in a context that’s pre-configured (with a time-out and the “cancel” method deferred) which means the context is scoped to the given API call (which in most cases should be the `WaitForCompletionRef` method).

Since I was prototyping this anyway whilst investigating it - I'll open a PR for this shortly which includes this proposed change, but it’s as yet untested (I’ll handle that shortly) - but I figured I’d check for feedback before proceeding any further - what do you think? :)

Thanks!

PR #326, opened 2018-09-28
Fixes in adal.Token for ADFS

jhendrixMSFT:
Changed a few fields in adal.Token from string to json.Number to handle
differences between AAD and ADFS in how they send data over the wire.


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/325

autorest/adal/token.go
```diff
@@ -101,27 +100,35 @@ type Token struct {
 	AccessToken  string `json:"access_token"`
 	RefreshToken string `json:"refresh_token"`
 
-	ExpiresIn string `json:"expires_in"`
-	ExpiresOn string `json:"expires_on"`
-	NotBefore string `json:"not_before"`
+	ExpiresIn json.Number `json:"expires_in"`
```
marstr: It may be a good moment to add some documentation about the expected contents of these values as well.

For example, I believe this is a whole number of seconds. We should say so :)

autorest/adal/token_test.go
```diff
@@ -868,7 +862,7 @@ func expireToken(t *Token) *Token {
 
 func setTokenToExpireAt(t *Token, expireAt time.Time) *Token {
 	t.ExpiresIn = "3600"
-	t.ExpiresOn = strconv.Itoa(int(expireAt.Sub(date.UnixEpoch()).Seconds()))
+	t.ExpiresOn = json.Number(strconv.Itoa(int(expireAt.Sub(date.UnixEpoch()).Seconds())))
```
marstr: I'd change this to
``` Go
t.ExpiresOn = json.Number(fmt.Sprintf(int64(expireAt.Sub(date.UnixEpoch() / time.Seconds))))
```

While it's not any more readable, it has two advantages IMHO:
1. Not calling `.Seconds()` prevents the conversion from an `int` type to a `float` type.
1. Using `fmt.Sprintf()` instead of `strconv.Itoa()` prevents us from casting down from 64 bits to 32 bits.

marstr: Errr, I see this is just a test file so don't see this comment as too blocking or anything.

autorest/adal/token.go
jhendrixMSFT: I've added a link to the docs for this type.

autorest/adal/token_test.go
jhendrixMSFT: I tweaked this a bit.

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/314

PR #329, opened 2018-10-04
v11.1.0

jhendrixMSFT:

### DIFF
--- a/autorest/adal/config.go
+++ b/autorest/adal/config.go
@@ -19,10 +19,6 @@ import (
 	"net/url"
 )
 
-const (
-	activeDirectoryAPIVersion = "1.0"
-)
-
 // OAuthConfig represents the endpoints needed
 // in OAuth operations
 type OAuthConfig struct {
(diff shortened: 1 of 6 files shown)


autorest/adal/config.go
```diff
@@ -79,3 +79,50 @@ func NewOAuthConfig(activeDirectoryEndpoint, tenantID string) (*OAuthConfig, err
...
+func NewOAuthConfigWithAPIVersion(activeDirectoryEndpoint, tenantID string, apiVersion *string) (*OAuthConfig, error) {
+	if err := validateStringParam(activeDirectoryEndpoint, "activeDirectoryEndpoint"); err != nil {
+		return nil, err
+	}
+	// it's legal for tenantID to be empty so don't validate it
+	if apiVersion != nil {
+		if err := validateStringParam(*apiVersion, "apiVersion"); err != nil {
+			return nil, err
+		}
+		*apiVersion = fmt.Sprintf("?api-version=%s", *apiVersion)
```
vladbarosan: won't this mutate the string value that was passed to the function ?

jhendrixMSFT: D'oh you're correct, will fix.

autorest/adal/config.go
```diff
@@ -79,3 +79,50 @@ func NewOAuthConfig(activeDirectoryEndpoint, tenantID string) (*OAuthConfig, err
 		DeviceCodeEndpoint: *deviceCodeURL,
 	}, nil
 }
+
+// NewOAuthConfigWithAPIVersion returns an OAuthConfig with tenant specific urls.
+// If apiVersion is not nil the "api-version" query parameter will be appended to the endpoint URLs with the specified value.
```
vladbarosan: can we make the old function call this one to avoid duplicated code ?

vladbarosan: I presume we would call it with the 1.0 param

jhendrixMSFT: Derp yep will fix.

PR #335, opened 2018-10-16
Add tracing module

vladbarosan:
- add tracing support. Instrumentation is done through a custom RoundTripper on the http client and through SDK function calls with the goal of correlating sdk calls with the api calls. Also convenience tracing calls are added.
- remove usage of original request's context on LRO polling.


vladbarosan:
Initially wanted to place this inside azure path as it makes more sense, but in that case would be painful to use it for adal. In the end I decided to use it for adal calls also since that would provide value to customers.

autorest/tracing/tracing.go
```diff
@@ -0,0 +1,124 @@
...
+			Enable()
+		}
+	}
+}
+
+// Enable will start instrumentation for metrics and traces.
+func Enable() (err error) {
+	Enabled = true
+	sampler = nil
+	Transport = &ochttp.Transport{Propagation: &tracecontext.HTTPFormat{}}
```
vladbarosan: note we updated for the W3C's Trace Context for propagating across servers instead of the B3 default implementation. Since Trace Context is on its way of becoming the standard but still a draft a little torn here. @joshgav regarding the ARM tracing, for immediate practical effects using the Trace Context means having only 1 header with any custom key-value pairs as opposed to some predefined headers for the B3 propagation .

autorest/azure/async.go
```diff
@@ -86,7 +88,23 @@ func (f Future) PollingMethod() PollingMethodType {
...
+// Deprecated: Use DoneWithContext()
 func (f *Future) Done(sender autorest.Sender) (bool, error) {
+	return f.DoneWithContext(context.Background(), sender)
+}
+
+// DoneWithContext queries the service to see if the operation has completed.
+func (f *Future) DoneWithContext(ctx context.Context, sender autorest.Sender) (done bool, err error) {
+	ctx = tracing.StartSpan(ctx, "github.com/Azure/go-autorest/autorest/azure/async.DoneWithContext")
+	defer func() {
+		sc := -1
```
jhendrixMSFT: Why -1, is this the convention?

autorest/azure/async.go
```diff
@@ -164,15 +182,25 @@ func (f Future) WaitForCompletion(ctx context.Context, client autorest.Client) e
...
+	ctx = tracing.StartSpan(ctx, "github.com/Azure/go-autorest/autorest/azure/async.WaitForCompletionRef")
+	defer func() {
+		sc := -1
+		resp := f.Response()
+		if resp != nil {
+			sc = resp.StatusCode
+		}
+		tracing.EndSpan(ctx, sc, err)
+	}()
+	cancelCtx := ctx
```
jhendrixMSFT: I don't think this copy is necessary.  When the defer func captures `ctx` it will have its value at that time, not the changed value on line 198.

tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+import (
+	"context"
+	"net/http"
+	"os"
+
+	"contrib.go.opencensus.io/exporter/ocagent"
```
jhendrixMSFT: nit: these could be re-ordered to remove extra new-lines.

tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+		if ok {
+			EnableWithAIForwarding(agentEndpoint)
+		} else {
+			Enable()
+		}
+	}
+}
+
+// Enable will start instrumentation for metrics and traces.
+func Enable() (err error) {
```
jhendrixMSFT: Should there be a matching `Disable()`?

tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+	Enabled = true
+	sampler = nil
+	Transport = &ochttp.Transport{Propagation: &tracecontext.HTTPFormat{}}
```
jhendrixMSFT: So looks to me that all clients will share the same tracing transport.  Is this safe for concurrent use?

tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+	trace.RegisterExporter(exporter)
+	return
+}
+
+// initStats registers the views for the http metrics
+func initStats() (err error) {
+	clientViews := []*view.View{
```
jhendrixMSFT: Is this something that should be user-configurable?

autorest/azure/async.go
vladbarosan: now that you mentioned if we never get a reply from a server, what is going to be our status code. 0 ? if so we should use 0

tracing/tracing.go
vladbarosan: should be, I was even thinking of sharing the same http client but was worry that it might get changed randomly in places and have global affects. In the end the calls are independent between them. they could make concurrent calls within the same client

tracing/tracing.go
vladbarosan: well they can add more views if they want, these are the convenience views we enable. Didnt want to go in providing another layer on top of opencensus. This is just a convenience of having the sdk intrument these. if they have already some or want more they can add more directly

autorest/azure/async.go
vladbarosan: updated to use an http to tracing status function according to the spec. for any different http code it will report unknown so here we can pass anything except the valid supported http codes and it will be fine

tracing/tracing.go
```diff
@@ -0,0 +1,149 @@
...
+		return trace.StatusCodePermissionDenied
+	case http.StatusUnauthorized: // 401 is actually unauthenticated.
+		return trace.StatusCodeUnauthenticated
+	case http.StatusTooManyRequests:
+		return trace.StatusCodeResourceExhausted
+	case http.StatusNotImplemented:
+		return trace.StatusCodeUnimplemented
+	case http.StatusServiceUnavailable:
+		return trace.StatusCodeUnavailable
+	case http.StatusOK:
```
jhendrixMSFT: Is this correct?  According to the table (linked in the func's comment) 200-399 should be considered OK.

tracing/tracing.go
```diff
@@ -0,0 +1,149 @@
...
+		return
+	}
+
+	if err != nil {
+		span.SetStatus(trace.Status{Message: err.Error(), Code: toTraceStatusCode(httpStatusCode)})
+	}
+	span.End()
+}
+
+// TraceStatusCode converts HTTP Codes to OpenCensus codes as defined
```
jhendrixMSFT: nit: func name in comment doesn't match actual func name.

tracing/tracing.go
jhendrixMSFT: Got it.  Please be sure to add this info to the docs for this feature (probably add info about tracing in the README).

tracing/tracing.go
jhendrixMSFT: If adding `Disable()` can be done in a non-breaking way then we don't have to hold up this PR for it.

vladbarosan: well it could but still investigating if that might cause any issue. im still testing scenarios to see if anything breaks

tracing/tracing.go
vladbarosan: yes, its a shame the library doesnt provide this functionality, ill make a PR upstream to add it after this

tracing/tracing.go
```diff
@@ -0,0 +1,149 @@
...
+	"go.opencensus.io/trace"
+)
+
+var (
+	// Transport is the default tracing RoundTripper. If tracing is disabled
+	// the default RoundTripper is used which provides no instrumentation.
+	Transport = http.DefaultTransport
+
+	// Enabled is the flag for marking if tracing is enabled.
+	Enabled = false
```
jhendrixMSFT: Do you expect customers to use this value or is it purely for testing purposes?  If it's the latter then perhaps we can not export it.

jhendrixMSFT: Or make an `enabled()` func that predicates on `sampler` being nil or not.

vladbarosan: well, I was thinking they might use it to check if monitoring is enabled on the sdk. if we want we could leave it unexported and if there is a requirement we can export it.

jhendrixMSFT: I see.  It might be better to make it `Enabled()` then so it's effectively read-only (I could see somebody setting it to false thinking it would turn it off).

vladbarosan: fair enough

autorest/adal/token.go
```diff
@@ -386,7 +387,7 @@ func (spt *ServicePrincipalToken) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	spt.refreshLock = &sync.RWMutex{}
```
vladbarosan: not sure why we create new variables here in the UnmarshallJson. the constructor creates them and if someone creates the type by himself than he should provide them

vladbarosan: added a nil check in case they might be set

PR #338, opened 2018-10-19
The adal package reports its own user-agent

jhendrixMSFT:
Make the adal package report its own user-agent string so the dependency
on the version package can be removed (prep for modules).


jhendrixMSFT:
CI is acting up, close then re-open.

.travis.yml
```diff
@@ -6,8 +6,6 @@ go:
   - master
   - 1.11.x
   - 1.10.x
-  - 1.9.x
```
vladbarosan: maybe specify in the changelog that we are dropping officially support for 1.9 and 1.8 ? ( but they are still expected to work)

PR #339, opened 2018-10-19
fix few misspell and one small cleaning

dvrkps:


autorest/sender.go
```diff
@@ -40,7 +40,7 @@ func (sf SenderFunc) Do(r *http.Request) (*http.Response, error) {
 	return sf(r)
 }
 
-// SendDecorator takes and possibily decorates, by wrapping, a Sender. Decorators may affect the
+// SendDecorator takes and possibility decorates, by wrapping, a Sender. Decorators may affect the
```
jhendrixMSFT: I think this word is supposed to be `possibly`.

jhendrixMSFT:
One little tweak, and if you can rebase on top of tip should be good to go.

autorest/sender.go
dvrkps: Done.

vladbarosan:
@dvrkps I updated the branch for you with latest master. @jhendrixMSFT I guess we will do a patch release for this. Do you want a changelog although I would say we can go without that since there is no functional change.

jhendrixMSFT:
@vladbarosan agreed no changelog update required.

Issue #333, opened 2018-10-15
Race in DoRetryForStatusCodes sender


mxk:
`autorest.Preparer` and `autorest.Responder` interfaces both explicitly say that the implementations "may be shared and re-used." The same is not said for `autorest.Sender`, but it's not clear to me whether that's an oversight in documentation or if the rules for senders are different.

Currently, the sender returned by `DoRetryForStatusCodes` decorator increments the `attempts` argument, which causes a race if multiple goroutines share the same sender. I'm not sure if there is an actual reason why this increment needs to happen as part of the execution rather than in the outermost function:

https://github.com/Azure/go-autorest/blob/ea233b6412b0421a65dc6160e16c893364664a95/autorest/sender.go#L214-L219

Either way, the documentation for `autorest.Sender` should be updated to say whether the implementations can or cannot be shared and re-used. Ideally, this race should be fixed to allow more efficient operation by reusing senders.

jhendrixMSFT:
@mxk it's not clear from your description where the race is, can you please clarify?

mxk:
Sure, run the following with `go test -race`:

```go
package main

import (
	"net/http"
	"net/http/httptest"
	"sync"
	"testing"

	"github.com/Azure/go-autorest/autorest"
)

func TestDoRetryForStatusCodes(t *testing.T) {
	s := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
	defer s.Close()

	sender := autorest.DecorateSender(s.Client(),
		autorest.DoRetryForStatusCodes(0, 0, http.StatusRequestTimeout))

	var wg sync.WaitGroup
	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			req, _ := http.NewRequest(http.MethodGet, s.URL, nil)
			if _, err := sender.Do(req); err != nil {
				t.Fatal(err)
			}
		}()
	}
	wg.Wait()
}
```

The `attempts` argument to `DoRetryForStatusCodes` becomes shared state, preventing the `sender` in my example from being used concurrently or even reused for multiple requests.

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v11.2.5.

PR #347, opened 2018-10-29
Fix race in autorest.DoRetryForStatusCodes

jhendrixMSFT:
Removed incrementing shared state.
Added `-race` flag to `go test` in travis CI.


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/333

autorest/sender.go
```diff
@@ -218,8 +218,7 @@ func DoRetryForStatusCodes(attempts int, backoff time.Duration, codes ...int) Se
 		return SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
 			rr := NewRetriableRequest(r)
 			// Increment to add the first call (attempts denotes number of retries)
-			attempts++
-			for attempt := 0; attempt < attempts; {
+			for attempt := 0; attempt < attempts+1; {
```
vladbarosan: so we dont need to increment attempts here ?

vladbarosan: or is enough to increment it later ?

jhendrixMSFT: It's done this way so that the first attempt isn't counted as a "retry attempt", i.e. if attempts is 5 it means 5 retry attempts which makes a total of six requests.

PR #355, opened 2019-01-07
Add method to obtain SP token from device flow

jhendrixMSFT:
Added method `ServicePrincipalToken()` to `DeviceFlowConfig` so callers
can obtain the underlying token.

### DIFF
--- a/autorest/azure/auth/auth.go
+++ b/autorest/azure/auth/auth.go
@@ -410,6 +410,16 @@ type DeviceFlowConfig struct {
 
 // Authorizer gets the authorizer from device flow.
 func (dfc DeviceFlowConfig) Authorizer() (autorest.Authorizer, error) {
+	spToken, err := dfc.ServicePrincipalToken()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth token from device flow: %v", err)
+	}
+
+	return autorest.NewBearerAuthorizer(spToken), nil
+}
+
+// ServicePrincipalToken gets the service principal token from device flow.
+func (dfc DeviceFlowConfig) ServicePrincipalToken() (*adal.ServicePrincipalToken, error) {
 	oauthClient := &autorest.Client{}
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/3804

autorest/azure/auth/auth.go
```diff
@@ -410,6 +410,16 @@ type DeviceFlowConfig struct {
...
+	spToken, err := dfc.ServicePrincipalToken()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth token from device flow: %v", err)
+	}
+
+	return autorest.NewBearerAuthorizer(spToken), nil
+}
+
+// ServicePrincipalToken gets the service principal token from device flow.
+func (dfc DeviceFlowConfig) ServicePrincipalToken() (*adal.ServicePrincipalToken, error) {
```
vladbarosan: only thing is, will this make a network call all the time ? from the issue on the sdk repo it seemed that he might wanted to get the tokens that the authorizer has without necessarily making a network call

jhendrixMSFT: Not every time, only if the token is about to expire.  Code would look something like this.
```go
deviceConfig := auth.NewDeviceFlowConfig("appID", "tenantID")
spToken, _ := deviceConfig.ServicePrincipalToken()
spToken.EnsureFresh()
token := spToken.Token()
fmt.Println(token.AccessToken)
fmt.Println(token.RefreshToken)
```

jhendrixMSFT: The actual network call happens in `EnsureFresh()`.

vladbarosan: ok sounds good ( I think he might wanted having a method to get the Token directly without having to call EnsureFresh() etc  I would if you can check with him otherwise looks good, approved)

Issue #363, opened 2019-03-05
Investigate setting minimum TLS version to 1.2 by default


jhendrixMSFT:
In `Client.sender()` we could set the minimum TLS version to 1.2.

https://www.socketloop.com/tutorials/golang-use-tls-version-1-2-and-enforce-server-security-configuration-over-client

https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L230

Related to https://github.com/Azure/azure-sdk-for-go/issues/3942

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v11.5.1

PR #364, opened 2019-03-05
Set minimum version of TLS to 1.2

jhendrixMSFT:
When creating an HTTP client set the minimum version of TLS to 1.2 and
prefer server ciphers.

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -16,6 +16,7 @@ package autorest
 
 import (
 	"bytes"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/363

autorest/client.go
```diff
@@ -230,6 +231,12 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 func (c Client) sender() Sender {
 	if c.Sender == nil {
 		j, _ := cookiejar.New(nil)
+		tracing.Transport.Base = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				MinVersion:               tls.VersionTLS12,
+				PreferServerCipherSuites: true,
```
vladbarosan: hmm the `preferServerCipherSuites` has any effect on clients ? AFAIK this setting is usually for servers to use their own cipher suites as opposed to what they receive from clients.

jhendrixMSFT: Hmm it could be that I misread [this article](https://www.socketloop.com/tutorials/golang-use-tls-version-1-2-and-enforce-server-security-configuration-over-client).  I can remove this if you agree.

vladbarosan: yea i think there its saying that when its configured on the server the clients are forced to use the servers cipherSuites

jhendrixMSFT: Removed.

PR #365, opened 2019-03-06
Fix GetTokenFromCLI cannot work in zsh

laozc:
GetTokenFromCLI uses shell to launch az when obtaining tokens.
This approach may produce an error under zsh as zsh doesn't always
assume the shell script to be located in /usr/bin.

As the environment variable PATH is set when executing, it can
rely on the shell to locate az CLI script rather than explicitly
specifying the shell.

### DIFF
--- a/autorest/azure/cli/token.go
+++ b/autorest/azure/cli/token.go
@@ -126,7 +126,7 @@ func GetTokenFromCLI(resource string) (*Token, error) {
 	azureCLIDefaultPathWindows := fmt.Sprintf("%s\\Microsoft SDKs\\Azure\\CLI2\\wbin; %s\\Microsoft SDKs\\Azure\\CLI2\\wbin", os.Getenv("ProgramFiles(x86)"), os.Getenv("ProgramFiles"))
 
 	// Default path for non-Windows.
-	const azureCLIDefaultPath = "/usr/bin:/usr/local/bin"
+	const azureCLIDefaultPath = "/bin:/sbin:/usr/bin:/usr/local/bin"
 
 	// Validate resource, since it gets sent as a command line argument to Azure CLI
 	const invalidResourceErrorTemplate = "Resource %s is not in expected format. Only alphanumeric characters, [dot], [colon], [hyphen], and [forward slash] are allowed."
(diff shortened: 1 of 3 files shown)


laozc:
Updated the latest commit.
Verified in the following environments:
1) Windows
2) macOS + bash/sh/zsh
3) Ubuntu 18.04/Centos 7/Arch Linux + bash/sh/dash/zsh/csh
4) WSL + bash/sh/zsh

autorest/azure/cli/token.go
```diff
@@ -144,13 +144,13 @@ func GetTokenFromCLI(resource string) (*Token, error) {
 		cliCmd = exec.Command(fmt.Sprintf("%s\\system32\\cmd.exe", os.Getenv("windir")))
 		cliCmd.Env = os.Environ()
 		cliCmd.Env = append(cliCmd.Env, fmt.Sprintf("PATH=%s;%s", os.Getenv(azureCLIPath), azureCLIDefaultPathWindows))
-		cliCmd.Args = append(cliCmd.Args, "/c")
+		cliCmd.Args = append(cliCmd.Args, "/c", "az")
 	} else {
-		cliCmd = exec.Command(os.Getenv("SHELL"))
```
jhendrixMSFT: Sorry I'm not super-familiar with Linux environments.  Why don't we rely on $SHELL anymore?

laozc: Not an expect in Golang implementation on the exec calls but I guess it acts the same as like exec* functions in C and/or other languages.

This change basically changes the command from
"$SHELL az login ..." whereas $SHELL=/bin/bash, /bin/sh, /bin/zsh, ...
to
"az login".

As long as the command is executable (+x), the exec is able to launch the program
when the program can be found in the PATH
since exec has the ability to locate your executable with the PATH environment variable.
In our case, it is set to include all the possible locations that az executable (azure-cli) is installed.
So it should have no issue to obtain the token from azure CLI.

The format $SHELL az login is taking az as a file.
Suppose your shell is bash, /bin/bash az login will ask bash to load a script file named "az"
passing the login argument and the rest to the script.
In this case the az doesn't have to be executable, only need to be written in bash syntax.
It might act exactly like "az login" as
1) Bash locates the script with PATH. bash may locate where az executable is
2) az executable is written in bash syntax
However, it would fail if the shell is zsh, as these assumptions are not always true
in other shell (zsh would complain it cannot read the file)
To make it work in zsh, you have to change the command to bash az login
even when your shell is not bash.
If you took this way you need bash in your path
(which won't be an issue on most environment as bash is widely adopted and installed)

I didn't do it as I don't think of any reason to rely on $SHELL in this case.
The exec can do as expected.
The shebang line in az (#!/usr/bin/env bash) can tell the shell it needs to be loaded and run in bash.
The az executable will always have the +x permission.

PR #367, opened 2019-03-08
Add KeyVaultResourceURL

karataliu:
Fix for https://github.com/Azure/azure-sdk-for-go/issues/4234

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -22,9 +22,14 @@ import (
 	"strings"
 )
 
-// EnvironmentFilepathName captures the name of the environment variable containing the path to the file
-// to be used while populating the Azure Environment.
-const EnvironmentFilepathName = "AZURE_ENVIRONMENT_FILEPATH"
+const (
+	// EnvironmentFilepathName captures the name of the environment variable containing the path to the file
+	// to be used while populating the Azure Environment.
+	EnvironmentFilepathName = "AZURE_ENVIRONMENT_FILEPATH"
+
+	// NotAvailable is used for endpoints and resource IDs that are not available for a given cloud.
+	NotAvailable = "N/A"
+)
 
 var environments = map[string]Environment{
 	"AZURECHINACLOUD":        ChinaCloud,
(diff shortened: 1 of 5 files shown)


autorest/azure/environments.go
```diff
@@ -43,6 +43,7 @@ type Environment struct {
 	ActiveDirectoryEndpoint      string `json:"activeDirectoryEndpoint"`
 	GalleryEndpoint              string `json:"galleryEndpoint"`
 	KeyVaultEndpoint             string `json:"keyVaultEndpoint"`
+	KeyVaultResourceID           string `json:"keyVaultResourceID"`
```
yangl900: since we are here, do you mind add some of these well-known resource IDs too?

| Resource | ResourceID |
|---|---|
| ARMResourceID |  "https://management.azure.com/" |
| GraphResourceID | "https://graph.windows.net/" |
| KeyVaultResourceID | "https://vault.azure.net" |
| DatalakeResourceID  | "https://datalake.azure.net/" |
| MicrosoftGraphResourceID | "https://graph.microsoft.com/" |
| BatchResourceID | "https://batch.core.windows.net/" |
| StorageResourceID | "https://storage.azure.com/" |

karataliu: Looks like existing property 'TokenAudience' is already for 'ARMResourceID'.

I'm fine about adding those, but since there are bunch of items, shall we separate them from existing configs. What about adding a sub property 'TokenAudiences' for all the ids?

e.g.
```
PublicCloud = Environment{
		Name:                         "AzurePublicCloud",
		ManagementPortalURL:          "https://manage.windowsazure.com/",
		PublishSettingsURL:           "https://manage.windowsazure.com/publishsettings/index",
                ...
		TokenAudiences: {
                  ResourceManager = "https://management.azure.com/",
                  Graph = "https://graph.windows.net/",
                  KeyVault= "https://vault.azure.net"
                  ...
                }
	}
```

@yangl900  @jhendrixMSFT

jhendrixMSFT: Agreed on separating into sub properties.

autorest/azure/environments.go
```diff
@@ -132,6 +164,14 @@ var (
 		ContainerRegistryDNSSuffix:   "azurecr.cn",
 		CosmosDBDNSSuffix:            "documents.azure.cn",
 		TokenAudience:                "https://management.chinacloudapi.cn/",
+		ResourceIdentifiers: ResourceIdentifier{
+			Graph:          "https://graph.chinacloudapi.cn/",
+			KeyVault:       "https://vault.azure.cn",
+			Datalake:       NotAvailable,
+			MicrosoftGraph: NotAvailable,
+			Batch:          "https://batch.chinacloudapi.cn/",
+			Storage:        NotAvailable,
```
vladbarosan: storage is not available in all clouds ?

jhendrixMSFT: I suspect it is but couldn't find it anywhere.  Hoping @yangl900 or somebody has the answer.

vladbarosan: ok. Im worried than that saying `NotAvailable` might send the wrong message that storage is not available in that cloud.

karataliu: this indicates storage-aad support is not available in all clouds.

jhendrixMSFT: Is it?  I can't find any documentation on what the resource ID should be.

PR #369, opened 2019-03-09
Add basic authorization

mikaelkrief:
Add Basic Authorization
+ associate tests

### DIFF
--- a/autorest/authorization.go
+++ b/autorest/authorization.go
@@ -15,6 +15,7 @@ package autorest
 //  limitations under the License.
 
 import (
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -31,6 +32,8 @@ const (
 	apiKeyAuthorizerHeader      = "Ocp-Apim-Subscription-Key"
 	bingAPISdkHeader            = "X-BingApis-SDK-Client"
 	golangBingAPISdkHeaderValue = "Go-SDK"
+	authorization               = "Authorization"
+	basic                       = "Basic"
 )
 
 // Authorizer is the interface that provides a PrepareDecorator used to supply request
@@ -257,4 +260,26 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
 		"aeg-sas-key": egta.topicKey,
 	}
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
+}
+
+// BasicAuthorizer implements authorization for Basic authentication.
+type BasicAuthorizer struct {
+	token string
+}
+
+// NewBasicAuthorizer is
+func NewBasicAuthorizer(token string) *BasicAuthorizer {
+	return &BasicAuthorizer{token: token}
+}
+
+// WithAuthorization is
+func (ba *BasicAuthorizer) WithAuthorization() PrepareDecorator {
+	headers := make(map[string]interface{})
+	headers[authorization] = basic + " " + basicAuth(":"+ba.token)
+
+	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
+}
+
+func basicAuth(token string) string {
+	return base64.StdEncoding.EncodeToString([]byte(token))
 }
--- a/autorest/authorization.go
+++ b/autorest/authorization.go
@@ -262,24 +262,26 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
 }
 
-// BasicAuthorizer implements authorization for Basic authentication.
+// BasicAuthorizer implements basic HTTP authorization by adding the Authorization HTTP header
+// with the value "Basic <TOKEN>" where <TOKEN> is a base64-encoded username:password tuple.
 type BasicAuthorizer struct {
-	token string
+	userName string
+	password string
 }
 
-// NewBasicAuthorizer is
-func NewBasicAuthorizer(token string) *BasicAuthorizer {
-	return &BasicAuthorizer{token: token}
+// NewBasicAuthorizer creates a new BasicAuthorizer with the specified username and password.
+func NewBasicAuthorizer(userName, password string) *BasicAuthorizer {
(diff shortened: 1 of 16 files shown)


autorest/authorization.go
```diff
@@ -258,3 +261,25 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
 	}
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
 }
+
+// BasicAuthorizer implements authorization for Basic authentication.
+type BasicAuthorizer struct {
+	token string
+}
+
+// NewBasicAuthorizer is
```
jhendrixMSFT: Finish comment.

autorest/authorization.go
```diff
@@ -258,3 +261,25 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
...
+func NewBasicAuthorizer(token string) *BasicAuthorizer {
+	return &BasicAuthorizer{token: token}
+}
+
+// WithAuthorization is
```
jhendrixMSFT: Finish comment.

autorest/authorization.go
```diff
@@ -258,3 +261,25 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
...
+func (ba *BasicAuthorizer) WithAuthorization() PrepareDecorator {
+	headers := make(map[string]interface{})
+	headers[authorization] = basic + " " + basicAuth(":"+ba.token)
+
+	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
+}
+
+func basicAuth(token string) string {
```
jhendrixMSFT: Since this is only called in one place I'd prefer it to be folded into `WithAuthorization()`.

mikaelkrief:
Hi @jhendrixMSFT ,
Azure DevOps Rest API use Basic authentication.

autorest/authorization.go
```diff
@@ -258,3 +261,21 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
...
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
 }
+
+// BasicAuthorizer implements authorization for Basic authentication.
+type BasicAuthorizer struct {
+	token string
+}
+
+// NewBasicAuthorizer creates a new BasicAuthorizer with the specified token.
+func NewBasicAuthorizer(token string) *BasicAuthorizer {
```
vladbarosan: to avoid confusion, can we name this BasicAccessAuthorizer or BasicHttpAuthorizer ? to be more obvious it refers to the http basic auth and not some generic BasicAuthorizer

autorest/authorization.go
```diff
@@ -258,3 +261,21 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
 	}
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
 }
+
+// BasicAuthorizer implements authorization for Basic authentication.
+type BasicAuthorizer struct {
+	token string
+}
+
+// NewBasicAuthorizer creates a new BasicAuthorizer with the specified token.
```
vladbarosan: dont we need a username also on top of the token/password ?

jhendrixMSFT: From the [docs](https://docs.microsoft.com/en-us/rest/api/azure/devops/?view=azure-devops-rest-5.0) it looks like all that's required is the token.

autorest/authorization.go
jhendrixMSFT: We could but it wouldn't follow the pattern we have for other authorizers, e.g. `BearerAuthorizer`.  I will update the comment to be more clear.

vladbarosan: yea fair enough , to be fair the auth schemes are called "Basic" , "Bearer" ... . so yea maybe have the comment more explicit that it creates an authorizer using the Basic Http Authorization Scheme

autorest/authorization.go
```diff
@@ -258,3 +261,23 @@ func (egta EventGridKeyAuthorizer) WithAuthorization() PrepareDecorator {
 	}
 	return NewAPIKeyAuthorizerWithHeaders(headers).WithAuthorization()
 }
+
+// BasicAuthorizer implements authorization by adding the Authorization HTTP header
+// with the value "Basic <TOKEN>" where <TOKEN> is a base64-encoded personal access token.
+type BasicAuthorizer struct {
```
vladbarosan: so it seems we are tying this too much with the VSTS requirement. THe Basic Http Scheme is more generic, accepts a base64 encoded username:password token.  https://tools.ietf.org/html/rfc7617

vladbarosan: so I guess the params here should be username, password ?

jhendrixMSFT: Hmm well if this is specific to VSTS should we rename the type to `VSTSAuthorizer`, similar to what we have for cognitive services etc?

mikaelkrief: It's not specific to VSTS.
https://tools.ietf.org/html/rfc7617

jhendrixMSFT: Well according to the RFC the token should be a tuple `userid:password` however in the ctor all we're asking for is a generic `token` parameter which from the VSTS docs looks like a personal access token.  Am I missing something?

mikaelkrief: for VSTS the token passed to the Autorization header is "Basic <any text>:encodBase64(PAT)"
https://docs.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/pats?view=azure-devops

jhendrixMSFT: I see.  So seems like we should update the ctor to take user-name and password params, and if user-name is empty string we base64 encode `":<password>"` then.  Does everybody agree?

jhendrixMSFT: @mikaelkrief I've refactored this to take a username/password.  Can you confirm this is working for you?  If so I will merge and tag.

mikaelkrief: @jhendrixMSFT , Yes that working fine for me. You can merge and tag.

PR #370, opened 2019-03-11
Commit vendored dependencies

jhendrixMSFT:
Added pruned, vendored dependencies
Removed all traces of dep from CI
Remove vendor directory from .gitignore


aviaoh:
This PR breaks latest `azure-sdk-for-go`. we are now getting the following error:
> cannot use env (type "github.com/Azure/go-autorest/autorest/azure".Environment) as type "github.com/Azure/azure-sdk-for-go/vendor/github.com/Azure/go-autorest/autorest/azure".Environment in argument to storage.NewAccountSASClient

Issue #375, opened 2019-03-22
`dep` dependencies broken for github.com/census-instrumentation/opencensus-proto


DrEsteban:
The latest version of this package fails to build when pulled in via `dep ensure` due to a breaking change pushed by one of your dependencies.

**Example Repro**
https://github.com/DrEsteban/Azure-go-autorest-repro
(Repro steps included in README)

**Workaround**
I needed to manually specify a dep [[override]] as follows:
```
[[override]]
  name = "github.com/census-instrumentation/opencensus-proto"
  version = "0.1.0"
```

jhendrixMSFT:
Fixed in https://github.com/Azure/go-autorest/releases/tag/v11.6.1.  Thanks for reporting this issue.

PR #377, opened 2019-03-25
Add CosmosDb endpoint and fix ACR endpoints for Gov Clouds

### LINKED ISSUE #366: Expose 'CosmosDBDNSSuffix' in the Environments struct
👋 

We're working through scoping out what's needed to add support for CosmosDB Collections & Databases into Terraform. As a part of the investigation here we've noticed that there's no suffix available as part of the Environment struct for the CosmosDB DNS Suffix, [as there is for example for Container Registries etc](https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L37-L57).

Would it be possible to add a new property to the Environments struct which exposes the CosmosDB DNS Suffix?

Thanks!

vladbarosan:
- fix DNS endpoint for ACR
- Add Cosmos DB endpoint

Fixes #366

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -54,6 +54,7 @@ type Environment struct {
 	ServiceManagementVMDNSSuffix string `json:"serviceManagementVMDNSSuffix"`
 	ResourceManagerVMDNSSuffix   string `json:"resourceManagerVMDNSSuffix"`
 	ContainerRegistryDNSSuffix   string `json:"containerRegistryDNSSuffix"`
+	CosmosDBDNSSuffix            string `json:"cosmosDBDNSSuffix"`
 	TokenAudience                string `json:"tokenAudience"`
 }
 
@@ -79,6 +80,7 @@ var (
 		ServiceManagementVMDNSSuffix: "cloudapp.net",
 		ResourceManagerVMDNSSuffix:   "cloudapp.azure.com",
 		ContainerRegistryDNSSuffix:   "azurecr.io",
+		CosmosDBDNSSuffix:            "documents.azure.com",
 		TokenAudience:                "https://management.azure.com/",
 	}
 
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Can you merge https://github.com/Azure/go-autorest/pull/376 into this PR?

autorest/azure/environments.go
```diff
@@ -150,7 +154,8 @@ var (
 		ServiceBusEndpointSuffix:     "servicebus.cloudapi.de",
 		ServiceManagementVMDNSSuffix: "azurecloudapp.de",
 		ResourceManagerVMDNSSuffix:   "cloudapp.microsoftazure.de",
-		ContainerRegistryDNSSuffix:   "azurecr.io",
+		ContainerRegistryDNSSuffix:   "",
```
jhendrixMSFT: This is supposed to be empty?

vladbarosan:
@jhendrixMSFT  it seems ACR is not present in the German Cloud . do we want to leave it blank ?

autorest/azure/environments.go
vladbarosan: beat me to it, there is no endpoint in [German Cloud]( https://azure.microsoft.com/en-us/global-infrastructure/services/?regions=china-non-regional,china-east,china-east-2,china-north,china-north-2,germany-non-regional,germany-central,germany-northeast,us-central,us-east,us-east-2,us-north-central,us-south-central,us-west-central,us-west,us-west-2&products=all )  ( also searched internally )

jhendrixMSFT: So if we omit it entirely and somebody tried to use this for German configuration will they get a compile time error?  I would prefer that to possible weird behavior due to empty string.  Or should the value be the one for public?

vladbarosan: removed per our discussion

vladbarosan:
Fixes #375

PR #381, opened 2019-04-01
tracing: fix proxy support for default sender (#379)

### LINKED ISSUE #379: tracing: default sender proxy support broken
The autorest package can currently not make use of proxies (specified via lower/upper-case http(s)_proxy environment variables) unless explicitly filling in the `Client.Sender` field.

## Problem description

The [tracing module](https://godoc.org/go.opencensus.io/plugin/ochttp) transparently allows proxy support by virtue of using the [`http.DefaultTransport`](https://golang.org/pkg/net/http/#RoundTripper), which uses [`ProxyFromEnvironment`](https://golang.org/pkg/net/http/#ProxyFromEnvironment) to re-route traffic via a Http(s) proxy.

Where no `Sender` is provided explicitly to the[`Client`](https://godoc.org/github.com/Azure/go-autorest/autorest#Client), _transparent proxy support is no longer possible_, due to the initialization in [`sender()`](https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L234-L237):
```go
// client.go

// sender returns the Sender to which to send requests.
func (c Client) sender() Sender {
    if c.Sender == nil {
       j, _ := cookiejar.New(nil)
       tracing.Transport.Base = &http.Transport{
                TLSClientConfig: &tls.Config{
                MinVersion: tls.VersionTLS12,
               },
       }
       client := &http.Client{Jar: j, Transport: tracing.Transport}
       return client
    }
    return c.Sender
}
```

The _cause_ is that, since the `Proxy` field uses the zero value, the `ProxyFromEnvironment`, which the http.DefaultTransport uses, no longer applies.

A _second_, related, problem is that the [_default timeouts_](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/) no longer apply.

As a result, _proxied requests_ of code that relies on the default sender _currently time out_:
```
Azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/path/to/resource?api-version=2018-04-01: |
StatusCode=0 -- Original Error: adal: Failed to execute the refresh request.
Error = 'Post https://login.microsoftonline.com/<uuid here>/oauth2/token?api-version=1.0:
dial tcp 40.126.2.38:443: connect: connection timed out
```

### Fix for first problem

We were able to bring back the http.DefaultTransport default using
```go
   tracing.Transport.Base = http.DefaultTransport
   tracing.Transport.Base
    if c.Sender == nil {
       // Use behaviour compatible with DefaultTransport, but require TLS minimum version.
       var baseTransport = *http.DefaultTransport.(*http.Transport)

       baseTransport.TLSClientConfig = &tls.Config{
                MinVersion: tls.VersionTLS12,
       }

       tracing.Transport.Base = &baseTransport
       j, _ := cookiejar.New(nil)
       // ...
     }
```

However, we then ran into another error (parts redacted as above):
```
azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/<uuid here>/resourceGroups/COSMIN/providers/Microsoft.Compute/disks/<diskName>?api-version=2018-04-01: StatusCode=0 -- Original Error: adal: Failed to execute the refresh request. Error = 'Post https://login.microsoftonline.com/<uuid here>/oauth2/token?api-version=1.0: proxyconnect tcp: tls: oversized record received with length 20527
```

This second error is mysterious - it fails on a standard SQUID proxy configuration using `https_proxy=https://proxy:port`.

grrtrr:
This resolves #379 by using the defaults of http.DefaultTransport without modifying the global object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v11.7.1
+
+### Bug Fixes
+
+- Fix missing support for http(s) proxy when using the default sender.
+
 ## v11.7.0
 
 ## New Features
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -230,14 +230,23 @@ func (c Client) Do(r *http.Request) (*http.Response, error) {
 // sender returns the Sender to which to send requests.
 func (c Client) sender() Sender {
 	if c.Sender == nil {
-		j, _ := cookiejar.New(nil)
+		// Use behaviour compatible with DefaultTransport, but require TLS minimum version.
+		var defaultTransport = http.DefaultTransport.(*http.Transport)
+
 		tracing.Transport.Base = &http.Transport{
+			Proxy:                 defaultTransport.Proxy,
+			DialContext:           defaultTransport.DialContext,
+			MaxIdleConns:          defaultTransport.MaxIdleConns,
+			IdleConnTimeout:       defaultTransport.IdleConnTimeout,
+			TLSHandshakeTimeout:   defaultTransport.TLSHandshakeTimeout,
+			ExpectContinueTimeout: defaultTransport.ExpectContinueTimeout,
 			TLSClientConfig: &tls.Config{
 				MinVersion: tls.VersionTLS12,
 			},
 		}
-		client := &http.Client{Jar: j, Transport: tracing.Transport}
-		return client
+
+		j, _ := cookiejar.New(nil)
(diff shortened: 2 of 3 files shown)


grrtrr:
@jhendrixMSFT - yes, will update PR and force-push branch changes.

PR #383, opened 2019-04-15
add support for TLS Renegotiation

tiraffe:
Related to https://github.com/Azure/go-autorest/issues/384

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New Features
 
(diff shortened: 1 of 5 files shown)


tiraffe:
> Can you please provide more context on why this is needed, specifically which cognitive services APIs require this to be enabled? It's a bit surprising that after 2+ years of the SDK this is only showing up now.

Sure, I have sent you an email to provide more details.

PR #390, opened 2019-04-19
create adal module

jhendrixMSFT:


autorest/adal/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/go-autorest/autorest/adal
+
+go 1.12
+
+require (
+	github.com/Azure/go-autorest/autorest/date v0.1.0
+	github.com/Azure/go-autorest/autorest/mocks v0.1.0
+	github.com/Azure/go-autorest/tracing v0.1.0
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
vladbarosan: what does incompatible mean ?

jhendrixMSFT: See https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-v2-package-that-has-not-opted-into-modules-what-does-incompatible-mean, highlight is as follows.
"When the go tool is in module mode, it will assume a non-module v2+ package has no awareness of Semantic Import Versioning and treat it as an (incompatible) extension of the v1 version series of the package (and the +incompatible suffix is an indication that the go tool is doing so)."

PR #393, opened 2019-04-19
update auth module

jhendrixMSFT:


autorest/azure/auth/go.mod
```diff
@@ -1,3 +1,11 @@
 module github.com/Azure/go-autorest/autorest/azure/auth
 
 go 1.12
+
+require (
+	github.com/Azure/go-autorest/autorest v0.1.0
```
vladbarosan: ah this is funky a little , it wont cause any issues ?

jhendrixMSFT: nope since we already carved the `auth` module out of the `autorest` module in the previous commit (that's why I couldn't include these dependencies in that commit).

PR #395, opened 2019-04-23
Migrate to go modules

syndbg:
# What/Why

Migrate to golang modules.

To ease everyone in the Golang community in using this package, golang modules are a must-have.

While gomod is somewhat capable of automatically dealing with cross-compatibility between `dep`, `glide`, etc, having multiple modules inside a mono repo requires some `gomod` "hacks".

# Must-have changes for the root (github.com/Azure/go-autorest)

Added golang module for the root of the repository.

I assume it was left out, cause `go build` (gomod) was complaining that you don't have any Golang files to compose a "module". However as many others have done, you can create a `doc.go`, which can serve a golang file enough to create a compose a golang module.

I've taken the pretty good documentation from the `README.md` and added as part of `doc.go`.

This will enable us to have a golang module and also improve the developer experience by having a good documentation in the source code itself.

# Nice-to-have changes for the other modules (github.com/Azure/go-autorest/{autorest,logger,tracing}

It is a great starting point that they're Golang modules already.

Since the root is now a golang module itself (from the previous point) we can now replace the import paths to make sure that the source code we're downloading does not cross-depend on packages that are already present.

Meaning that it prevents the scenarios where you have `github.com/Azure/go-autorest/autorest` that depends on `github.com/Azure/go-autorest/logger` and you go over the network to download a version of the code you already have locally. This is applicable for all other scenarios too.

# Must-have changes for CI

Updated TravisCI to support Golang 1.12.x

# Nice-to-have-changes for CI

I dropped support for Golang 1.10.x and 1.11.x. I assume that they weren't that much used anyway, 

since **every other** package was already a golang module. It was probably kept until the root module is migrated itself.

**I can retain support for Golang 1.10.x and 1.11.x** by keeping `Gopkg.{toml,lock`** and making the CI pipeline support both, but I assume that you don't want that, right?

## PR workflow question/confusion

Are you still submitting PRs to `dev`? From the git history I see that `dev` is pretty much abandonware.

Should I submit this PR to `dev` or `master` is ok with you?

## PR checklist


syndbg:
CI passes, ready for review.

Not sure who to mention for review, so.. wdyt @jhendrixMSFT ?

jhendrixMSFT:
I don't think we can take this quite yet since azure-sdk-for-go has not yet migrated to modules (work in progress).  I want to keep this open though for when the right time comes.

jhendrixMSFT:
This PR is no longer applicable given our multi-module strategy.

PR #408, opened 2019-06-19
Environment: adding the Storage Management Endpoint for AzureAD auth

tombuildsstuff:
It's possible to authenticate to Azure Storage [using Azure Active Directory authentication](https://docs.microsoft.com/en-us/rest/api/storageservices/authenticate-with-azure-active-directory), which requires generating an Authentication Token against the Storage Management Endpoint.

This PR adds a new variable to the Environment struct which is the Storage Management Endpoint; I can confirm in both Azure Public and Azure Germany that these endpoints are valid, and return tokens via the Azure CLI command:

```bash
$ az account get-access-token --resource "https://storage.azure.com/" -o=json
```

I can confirm this endpoint works in the Azure Storage SDK that we're building using go-autorest as a base for the items above, however I'm unable to confirm the following (but have asked around):

* the Azure China Endpoint
* the Azure Government Endpoint
* if the Endpoint is returned from the MetaData service ([it appears not](http://management.azure.com/metadata/endpoints?api-version=1.0) - but I'm not sure if that fields not returned)

---

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -49,12 +49,13 @@ type ResourceIdentifier struct {
 
 // Environment represents a set of endpoints for each of Azure's Clouds.
 type Environment struct {
-	Name                         string             `json:"name"`
-	ManagementPortalURL          string             `json:"managementPortalURL"`
-	PublishSettingsURL           string             `json:"publishSettingsURL"`
-	ServiceManagementEndpoint    string             `json:"serviceManagementEndpoint"`
-	ResourceManagerEndpoint      string             `json:"resourceManagerEndpoint"`
-	ActiveDirectoryEndpoint      string             `json:"activeDirectoryEndpoint"`
+	Name                         string `json:"name"`
+	ManagementPortalURL          string `json:"managementPortalURL"`
+	PublishSettingsURL           string `json:"publishSettingsURL"`
+	ServiceManagementEndpoint    string `json:"serviceManagementEndpoint"`
+	ResourceManagerEndpoint      string `json:"resourceManagerEndpoint"`
+	ActiveDirectoryEndpoint      string `json:"activeDirectoryEndpoint"`
+	StorageEndpoint              string
 	GalleryEndpoint              string             `json:"galleryEndpoint"`
 	KeyVaultEndpoint             string             `json:"keyVaultEndpoint"`
 	GraphEndpoint                string             `json:"graphEndpoint"`
@@ -82,6 +83,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.windows.net/",
 		ResourceManagerEndpoint:      "https://management.azure.com/",
 		ActiveDirectoryEndpoint:      "https://login.microsoftonline.com/",
+		StorageEndpoint:              "https://storage.azure.com/",
 		GalleryEndpoint:              "https://gallery.azure.com/",
 		KeyVaultEndpoint:             "https://vault.azure.net/",
 		GraphEndpoint:                "https://graph.windows.net/",
@@ -114,6 +116,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.usgovcloudapi.net/",
 		ResourceManagerEndpoint:      "https://management.usgovcloudapi.net/",
 		ActiveDirectoryEndpoint:      "https://login.microsoftonline.us/",
+		StorageEndpoint:              "https://storage.azure.com/",
 		GalleryEndpoint:              "https://gallery.usgovcloudapi.net/",
 		KeyVaultEndpoint:             "https://vault.usgovcloudapi.net/",
 		GraphEndpoint:                "https://graph.windows.net/",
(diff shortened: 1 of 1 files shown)


autorest/azure/environments.go
```diff
@@ -49,12 +49,13 @@ type ResourceIdentifier struct {
...
-	ServiceManagementEndpoint    string             `json:"serviceManagementEndpoint"`
-	ResourceManagerEndpoint      string             `json:"resourceManagerEndpoint"`
-	ActiveDirectoryEndpoint      string             `json:"activeDirectoryEndpoint"`
+	Name                         string `json:"name"`
+	ManagementPortalURL          string `json:"managementPortalURL"`
+	PublishSettingsURL           string `json:"publishSettingsURL"`
+	ServiceManagementEndpoint    string `json:"serviceManagementEndpoint"`
+	ResourceManagerEndpoint      string `json:"resourceManagerEndpoint"`
+	ActiveDirectoryEndpoint      string `json:"activeDirectoryEndpoint"`
+	StorageEndpoint              string
```
tombuildsstuff: unfortunately this isn't returned from a MetaData endpoint right now ([that I can see](https://management.azure.com/metadata/endpoints?api-version=1.0)), but I may be wrong

autorest/azure/environments.go
```diff
@@ -82,6 +83,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.windows.net/",
 		ResourceManagerEndpoint:      "https://management.azure.com/",
 		ActiveDirectoryEndpoint:      "https://login.microsoftonline.com/",
+		StorageEndpoint:              "https://storage.azure.com/",
```
tombuildsstuff: ✅ I can confirm this one works

autorest/azure/environments.go
```diff
@@ -178,6 +182,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.cloudapi.de/",
 		ResourceManagerEndpoint:      "https://management.microsoftazure.de/",
 		ActiveDirectoryEndpoint:      "https://login.microsoftonline.de/",
+		StorageEndpoint:              "https://storage.azure.com/",
```
tombuildsstuff: ✅ I can confirm this one works

autorest/azure/environments.go
```diff
@@ -114,6 +116,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.usgovcloudapi.net/",
 		ResourceManagerEndpoint:      "https://management.usgovcloudapi.net/",
 		ActiveDirectoryEndpoint:      "https://login.microsoftonline.us/",
+		StorageEndpoint:              "https://storage.azure.com/",
```
tombuildsstuff: 🚫 I can't confirm this one personally, but I'm asking around

autorest/azure/environments.go
```diff
@@ -146,6 +149,7 @@ var (
 		ServiceManagementEndpoint:    "https://management.core.chinacloudapi.cn/",
 		ResourceManagerEndpoint:      "https://management.chinacloudapi.cn/",
 		ActiveDirectoryEndpoint:      "https://login.chinacloudapi.cn/",
+		StorageEndpoint:              "https://storage.azure.com/",
```
tombuildsstuff: 🚫 I can't confirm this one personally, but I'm asking around

jhendrixMSFT:
@feiyushi do you happen to know if the storage endpoints in question are good?

feiyushi:
> @feiyushi do you happen to know if the storage endpoints in question are good?

Sorry I'm not sure since I've only tested in the AzurePublicCloud.

@jhendrixMSFT, @tombuildsstuff,  I thought `Environment.ResourceIdentifiers` is used as the token audience (resource id) for the purpose of generating a token from AAD?

tombuildsstuff:
@feiyushi [it's not what we (Terraform) use today](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go#L353-L367) but the values appear to be the same as what we're using - so it could well make sense in that struct?

feiyushi:
> @feiyushi [it's not what we (Terraform) use today](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go#L353-L367) but the values appear to be the same as what we're using - so it could well make sense in that struct?

@jhendrixMSFT knows the full history of that struct (fairly new one). 
As far as I know, `TokenAudience` is the  resource id for the ARM. And people use the endpoint values for the resource ids other than ARM. However, there're some exceptions like key vault (no ending `/` for the resource id). I think that's the reason for the new struct.

tombuildsstuff:
Thinking about this a little more just in-case this ends up being a pain to track down if these endpoints are valid; since Germany allows generating an access token for `storage.azure.com` endpoint (and that's what the docs refer to) perhaps that's a safe assumption for Government/China too?

jhendrixMSFT:
Looking back at your initial comment, in your example this endpoint is a resource ID?  If that's the case then we already have this for storage, see https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L107

tombuildsstuff:
@jhendrixMSFT good spot 🤦‍♂ maybe we can just use that instead them

FWIW @jeffreyCline has confirmed the China and Government endpoints respond correctly:

```
AzureChinaCloud:

C:\WINDOWS\system32>az account get-access-token --resource "https://storage.azure.com/" -o=json
{
  "accessToken": "*****",
  "expiresOn": "2019-06-20 16:36:37.571957",
  "subscription": "*****",
  "tenant": "*****",
  "tokenType": "Bearer"
}
```

```
AzureUSGovernment:

C:\WINDOWS\system32>az account get-access-token --resource "https://storage.azure.com/" -o=json
{
  "accessToken": "*****",
  "expiresOn": "2019-06-20 17:54:56.877611",
  "subscription": "*****",
  "tenant": "*****",
  "tokenType": "Bearer"
}
```

so I think we're good here and I can close this PR 👍 

Thanks!

PR #412, opened 2019-06-24
Add support for multi-tenant authentication

jhendrixMSFT:
Support for multi-tenant via x-ms-authorization-auxiliary header has
been added for client credentials with secret scenario; this basically
bundles multiple OAuthConfig and ServicePrincipalToken types into
corresponding MultiTenant* types along with a new authorizer that adds
the primary and auxiliary token headers to the reqest.
The authenticaion helpers have been updated to support this scenario; if
environment var AZURE_AUXILIARY_TENANT_IDS is set with a semicolon
delimited list of tenants the multi-tenant codepath will kick in to
create the appropriate authorizer.

### DIFF
--- a/autorest/adal/config.go
+++ b/autorest/adal/config.go
@@ -15,8 +15,13 @@ package adal
 //  limitations under the License.
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
+)
+
+const (
+	activeDirectoryEndpointTemplate = "%s/oauth2/%s%s"
 )
 
 // OAuthConfig represents the endpoints needed
@@ -60,7 +65,6 @@ func NewOAuthConfigWithAPIVersion(activeDirectoryEndpoint, tenantID string, apiV
 		}
 		api = fmt.Sprintf("?api-version=%s", *apiVersion)
 	}
-	const activeDirectoryEndpointTemplate = "%s/oauth2/%s%s"
 	u, err := url.Parse(activeDirectoryEndpoint)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


autorest/adal/config.go
```diff
@@ -89,3 +93,58 @@ func NewOAuthConfigWithAPIVersion(activeDirectoryEndpoint, tenantID string, apiV
...
 }
+
+// MultiTenantOAuthConfig provides endpoints for primary and aulixiary tenant IDs.
+type MultiTenantOAuthConfig interface {
+	PrimaryTenant() *OAuthConfig
+	AuxiliaryTenants() []*OAuthConfig
+}
+
+// Config contains optional OAuthConfig creation arguments.
+type Config struct {
```
jhendrixMSFT: I think this type needs a better name.

vladbarosan: >Config [](start = 5, length = 6)

well is this meant to be just the config for AAD or more generic ?

autorest/adal/config.go
```diff
@@ -89,3 +93,58 @@ func NewOAuthConfigWithAPIVersion(activeDirectoryEndpoint, tenantID string, apiV
...
+	if c.APIVersion != "" {
+		return fmt.Sprintf("?api-version=%s", c.APIVersion)
+	}
+	return "1.0"
+}
+
+// NewMultiTenantOAuthConfig creates an object that support multitenant OAuth configuration.
+func NewMultiTenantOAuthConfig(activeDirectoryEndpoint, primaryTenantID string, auxiliaryTenantIDs []string, cfg Config) (MultiTenantOAuthConfig, error) {
+	if len(auxiliaryTenantIDs) == 0 || len(auxiliaryTenantIDs) > 3 {
+		return nil, errors.New("must specify one to three auxiliary tenants")
```
vladbarosan: >three auxiliary ten [](start = 46, length = 19)

is there any documentation pointing to this ?

jhendrixMSFT: Yes see https://docs.microsoft.com/en-us/azure/azure-resource-manager/authenticate-multi-tenant

autorest/adal/config.go
jhendrixMSFT: It's meant as optional config params for `NewMultiTenantOAuthConfig()` so that we don't end up in the same situation as `NewOAuthConfig()` and `NewOAuthConfigWithAPIVersion()`.

autorest/adal/config.go
vladbarosan: should we point in the comment to that page ?

---
In reply to: [297407597](https://github.com/Azure/go-autorest/pull/412#discussion_r297407597) [](ancestors = 297407597)

jhendrixMSFT: Yep was thinking the same, will add.

autorest/adal/config.go
jhendrixMSFT: Maybe `Options`?

PR #416, opened 2019-07-01
New Authorizers for Azure Storage

tombuildsstuff:
👋 

This PR introduces a couple of new Authorizers for Azure Storage - the logic for which is lifted/amended from [the existing (deprecated) Azure Storage SDK for Go](https://github.com/Azure/azure-sdk-for-go/tree/master/storage) - which allows SDK's based on this library to talk to Azure Storage.

The first Authorizer allows authenticating to the Blob, File and Queue API's using a Shared Key Lite (`SharedKeyLiteAuthorizer`) and the other (`SharedKeyLiteTableAuthorizer`) allows authenticating to the Table Storage API's using a Shared Key Lite - both of which are based off the Account Name and Account Key.

We've been using these in the Storage SDK we're building - and I can confirm these work:

```
$ go build . && envchain azurerm ./example
2019/07/01 19:10:28 Testing if Service Principal / Client Certificate is applicable for Authentication..
2019/07/01 19:10:28 Testing if Service Principal / Client Secret is applicable for Authentication..
2019/07/01 19:10:28 Using Service Principal / Client Secret for Authentication
2019/07/01 19:10:28 [DEBUG] AzureRM Request:
POST /Tables HTTP/1.1
Host: tomdevsa29.table.core.windows.net
User-Agent: Go/go1.12.6 (amd64-darwin) go-autorest/v11.7.0 tombuildsstuff/giovanni/v0.0.1 storage/2018-11-09
Content-Length: 23
Accept: application/json;odata=nometadata
Content-Type: application/json
Date: Mon, 01 Jul 2019 17:10:28 GMT
Prefer: return-no-content
X-Ms-Date: Mon, 01 Jul 2019 17:10:28 GMT
X-Ms-Version: 2018-11-09
Accept-Encoding: gzip

{"TableName":"sdktest"}
2019/07/01 19:10:29 [DEBUG] AzureRM Response for https://tomdevsa29.table.core.windows.net/Tables:
HTTP/1.1 204 No Content
Content-Length: 0
Cache-Control: no-cache
Dataserviceid: https://tomdevsa29.table.core.windows.net/Tables('sdktest')
Date: Mon, 01 Jul 2019 17:10:29 GMT
Location: https://tomdevsa29.table.core.windows.net/Tables('sdktest')
Preference-Applied: return-no-content
Server: Windows-Azure-Table/1.0 Microsoft-HTTPAPI/2.0
X-Content-Type-Options: nosniff
X-Ms-Request-Id: 54dbae8a-7002-007a-2f2f-30250d000000
X-Ms-Version: 2018-11-09
```

using a client based on Azure/go-autorest

```
storageKeyLiteAuthorizer := auth.NewSharedKeyLiteTableAuthorizer(accountName, accountKey)
client := tables.New()
client.Client = prepareClient(client.Client, storageKeyLiteAuthorizer)

_, err := blobClient.Create(ctx, "tomdevsa29", "sdktest")
if err != nil {
	return err
}
```

---


tombuildsstuff:
@jhendrixMSFT I've split the storage authorizers out into their own files since there's a /lot/ in the existing ones - perhaps it's worth splitting the authorizers out into their own package/files?

tombuildsstuff:
Heads up that the SDK we've been using to build/test this is now public - if it's helpful to test this works as expected: https://github.com/tombuildsstuff/giovanni

autorest/authorization_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"net/url"
+	"sort"
+	"strings"
+	"time"
+)
+
+var (
```
jhendrixMSFT: Const?

tombuildsstuff: yes these should be - I'll push a commit to fix

jhendrixMSFT:
@tombuildsstuff is the guts of the implementation a copy of what's in the `storage` package?

tombuildsstuff:
@jhendrixMSFT that was the starting point for this, yeah - I'm not sure if that wanted a different license on it or something to state that (sorry!)?

kahing:
From a quick glance this doesn't sign the entire query string? That's not going to work for things like datalake gen 2. Also want to fix bugs like https://github.com/Azure/azure-storage-blob-go/issues/146 while you port that over

tombuildsstuff:
@kahing at the moment this authorizer only supports [SharedKeyLite](https://github.com/Azure/go-autorest/pull/416/files#diff-d980313d787e9b5d7d676d7ee794038bR137) rather than SharedKey authorization - since that's what we needed when [building a Storage SDK based on go-autorest](https://github.com/tombuildsstuff/giovanni).

That said - I believe it should be possible to extend this to a SharedKey authorizer if that's needed, since the Storage API's should work with either (I guess really both should be supported?)

invidian:
It looks like Premium Tier for storage accounts does not work with SharedKeyLite (see https://github.com/terraform-providers/terraform-provider-azurerm/issues/3939#issuecomment-524401721), so I also suggest adding SharedKey support here, especially that the basic implementation is not much different from SharedKeyLite.

tombuildsstuff:
@invidian thanks for the heads up - so that we can get both the SharedKey and the SharedKeyLite Authorizers into this PR I'm going to pull that patch into AzureRM, run the tests and then update this PR with you as a co-author in both cases, I hope you don't mind?

invidian:
>@invidian thanks for the heads up - so that we can get both the SharedKey and the SharedKeyLite Authorizers into this PR I'm going to pull that patch into AzureRM, run the tests and then update this PR with you as a co-author in both cases, I hope you don't mind?

I don't mind at all!

I have a testing patch for that, available here: https://github.com/kinvolk/terraform-provider-azurerm/commit/f9676819c28b236967b191d8858c04f118135095, where I tested that Terraform provider behaves as expected. It requires some rework of course, which I was going to do today and push it to AzureRM provider. Like:
- move shared code between SharedKeyLite and SharedKey to helpers
- clarify on `buildCanonicalizedResource` interface.

@tombuildsstuff what would be your timeline on that? I'm asking, because if you work on that next week, then I will do it today, so we can move things forward faster, as this is important for me that this issue is resolved.

Also feel free to pull me for any related code-reviews if that's helpful.

tombuildsstuff:
@invidian 

> what would be your timeline on that? 

I'm looking into this [at the moment](https://github.com/terraform-providers/terraform-provider-azurerm/compare/f/shared-key-auth), so once I've done some testing to determine if we also need a SharedKey authorizer for Table storage we should be good to move forward :)

autorest/authorization_storage.go
```diff
@@ -19,6 +19,101 @@ import (
...
+
+// WithAuthorization returns a PrepareDecorator that adds an HTTP Authorization header whose
+// value is "SharedKey " followed by the computed key.
+// This can be used for the Blob, Queue, and File Services
+//
+// from: https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key
+// You may use Shared Key Lite authorization to authorize a request made against the
+// 2009-09-19 version and later of the Blob and Queue services,
+// and version 2014-02-14 and later of the File services.
+func (skl *SharedKeyAuthorizer) WithAuthorization() PrepareDecorator {
```
invidian: does `skl` here refer to `SharedKeyLite`?

autorest/authorization_storage.go
```diff
@@ -19,6 +19,101 @@ import (
...
+		storageAccountKey:  accountKey,
+	}
+}
+
+// WithAuthorization returns a PrepareDecorator that adds an HTTP Authorization header whose
+// value is "SharedKey " followed by the computed key.
+// This can be used for the Blob, Queue, and File Services
+//
+// from: https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key
+// You may use Shared Key Lite authorization to authorize a request made against the
```
invidian: `s/Shared Key Lite/Shared Key/g`?

autorest/authorization_storage.go
```diff
@@ -19,6 +19,101 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	// we then need to HMAC that value
+	hmacdValue := hmacValue(storageAccountKey, *sharedKey)
+	return &hmacdValue, nil
+}
+
+// computeSharedKey computes the Shared Key Lite required for Storage Authentication
```
invidian: `s/Shared Key Lite/Shared Key/g`?

jhendrixMSFT:
@tombuildsstuff I refactored this to pretty much be a copy of what's in the /storage package in the azure-sdk-for-go repo.  There's just one SharedKeyAuthorizer type now that takes an enum value for the shared key type.  PTAL

autorest/authorization_storage.go
```diff
@@ -0,0 +1,298 @@
...
+// SharedKeyAuthorizer implements an authorization for Shared Key
+// this can be used for interaction with Blob, File and Queue Storage Endpoints
+type SharedKeyAuthorizer struct {
+	accountName string
+	accountKey  []byte
+	keyType     SharedKeyType
+}
+
+// NewSharedKeyAuthorizer creates a SharedKeyAuthorizer using the provided credentials and shared key type.
+func NewSharedKeyAuthorizer(accountName, accountKey string, keyType SharedKeyType) (SharedKeyAuthorizer, error) {
```
tombuildsstuff: I'd argue this should be:

```suggestion
func NewSharedKeyAuthorizer(accountName, accountKey string, keyType SharedKeyType) (*SharedKeyAuthorizer, error) {
```

since if there's an error there shouldn't be an authorizer returned?

autorest/authorization_storage_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Fatalf("create shared key authorizer: %v", err)
+	}
+	req, err := http.NewRequest(http.MethodGet, "https://golangrocksonazure.blob.core.windows.net/some/blob.dat", nil)
+	if err != nil {
+		t.Fatalf("create HTTP request: %v", err)
+	}
+	req.Header.Add("Accept-Charset", "UTF-8")
+	req.Header.Add(headerContentType, "application/json")
+	req.Header.Add(headerXMSDate, "Wed, 23 Sep 2015 16:40:05 GMT")
+	req.Header.Add(headerContentLength, "0")
```
tombuildsstuff: potentially pedantry: but I'd suggest this might be better to hard-code these keys to account for potential bugs in the key names?

autorest/authorization_storage_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Fatalf("create shared key authorizer: %v", err)
+	}
+	req, err := http.NewRequest(http.MethodGet, "https://golangrocksonazure.table.core.windows.net/tquery()", nil)
+	if err != nil {
+		t.Fatalf("create HTTP request: %v", err)
+	}
+	req.Header.Add("Accept-Charset", "UTF-8")
+	req.Header.Add(headerContentType, "application/json")
+	req.Header.Add(headerXMSDate, "Wed, 23 Sep 2015 16:40:05 GMT")
+	req.Header.Add(headerContentLength, "0")
```
tombuildsstuff: (same here)

autorest/authorization_storage_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	req, err := http.NewRequest(http.MethodGet, "https://golangrocksonazure.table.core.windows.net/tquery()", nil)
+	if err != nil {
+		t.Fatalf("create HTTP request: %v", err)
+	}
+	req.Header.Add("Accept-Charset", "UTF-8")
+	req.Header.Add(headerContentType, "application/json")
+	req.Header.Add(headerXMSDate, "Wed, 23 Sep 2015 16:40:05 GMT")
+	req.Header.Add(headerContentLength, "0")
```
tombuildsstuff: (same here)

autorest/authorization_storage_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	req, err := http.NewRequest(http.MethodGet, "https://golangrocksonazure.file.core.windows.net/some/file.dat", nil)
+	if err != nil {
+		t.Fatalf("create HTTP request: %v", err)
+	}
+	req.Header.Add("Accept-Charset", "UTF-8")
+	req.Header.Add(headerContentType, "application/json")
+	req.Header.Add(headerXMSDate, "Wed, 23 Sep 2015 16:40:05 GMT")
+	req.Header.Add(headerContentLength, "0")
```
tombuildsstuff: (same here)

autorest/authorization_storage_test.go
jhendrixMSFT: Good point I'll remove the strings, replacing them with constants.

autorest/authorization_storage.go
jhendrixMSFT: I have it returning by value as the method set for this type is also by value (SharedKeyAuthorizer is immutable and copy semantics are ok).  And if error is non-nil the caller shouldn't be touching any other return values.

invidian: But do you trust the caller? :) Returning nil pointer here is safer here in my opinion.

jhendrixMSFT: We discussed this and will change to by-ref but not because it's safer (ignoring the error breaks the method's contract).  A `SharedKeyAuthorizer` is unique for a set of inputs so should be shared; in the future we could also add support for changing the account key so copy-semantics don't work there.  Also, the zero-value for `SharedKeyAuthorizer` is useless.

PR #463, opened 2019-08-21
Add MSI for App Service / Functions support

### LINKED ISSUE #447: Add support for MSI in Azure App Service
Please add support for the different MSI endpoint and flow used in the authentication of Azure App Service. This flow uses the `MSI_ENDPONT` and `MSI_SECRET` environment variables.

https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity#using-the-rest-protocol

mbrancato:
This adds support for the different method of MSI token acquisition on Azure App Service and Azure Functions.

Fixes #447 

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -667,17 +667,17 @@ func GetMSIEndpoint() (string, error) {
 
 // NewServicePrincipalTokenFromMSI creates a ServicePrincipalToken via the MSI VM Extension.
 // It will use the system assigned identity when creating the token.
-func NewServicePrincipalTokenFromMSI(msiEndpoint string, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
+func NewServicePrincipalTokenFromMSI(msiEndpoint, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
 	return newServicePrincipalTokenFromMSI(msiEndpoint, resource, nil, callbacks...)
 }
 
 // NewServicePrincipalTokenFromMSIWithUserAssignedID creates a ServicePrincipalToken via the MSI VM Extension.
 // It will use the specified user assigned identity when creating the token.
-func NewServicePrincipalTokenFromMSIWithUserAssignedID(msiEndpoint string, resource string, userAssignedID string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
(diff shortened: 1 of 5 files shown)


autorest/adal/token.go
```diff
@@ -634,6 +635,31 @@ func GetMSIVMEndpoint() (string, error) {
 	return msiEndpoint, nil
 }
 
+func isAppService() bool {
+	_, endpointEnvExists := os.LookupEnv("MSI_ENDPOINT")
```
jhendrixMSFT: Can you make MSI_ENDPOINT and MSI_SECRET unexported constants to avoid the duplication.

autorest/adal/token_test.go
```diff
@@ -722,6 +723,22 @@ func TestGetVMEndpoint(t *testing.T) {
 	}
 }
 
+func TestGetAppServiceEndpoint(t *testing.T) {
```
jhendrixMSFT: Can you also add a test (or update this one, whatever makes sense) to cover `GetMSIEndpoint()`.

mbrancato: I've added a tests for both VM and App Service results in the main `GetMSIEndpoint()` function. I also cleaned up the environment variable handling in tests.

autorest/adal/token.go
mbrancato: I moved these to constants and replaced all the static string use.

mbrancato:
I think I've completed the changes @jhendrixMSFT. Let me know if anything else needs to be updated.

There is probably a need to perform retries as there is some race condition before the MSI endpoint proxy is available in App Service. I'm not certain if this should be added to this PR or add as an enhancement in a new PR.

jhendrixMSFT:
@mbrancato thanks!  I also rebased on top of master.  Good point about the retries, in token.go `refreshInternal()` we call `isIMDS()` to determine which retry algorithm to use.  For this change I think `isIMDS()` will return `false` so no retries will happen.  I think `isIMDS()` should be updated to handle this scenario so the retry logic will apply.

jhendrixMSFT:
@mbrancato now that I reworked https://github.com/Azure/go-autorest/pull/464 does it resolve the problem you were having?

mbrancato:
@jhendrixMSFT I was able to take some time and test this latest branch merged with the upstream master. It is working in my testing and I'm not seeing any issues because the retry logic.

Issue #450, opened 2019-08-14
ECONNRESET on read is not a transient network error


mgaurav:
In DoRetryWithStatusCodes, we retry for a list of status codes and also retry on temporary network errors. However, I am seeing transient network errors not being retried.

E.g.
`compute.DisksClient#List: Failure sending request: StatusCode=0 -- Original Error: Get https://management.azure.com/subscriptions/13979c32-3382-46f7-8f3a-ef86b8270227/providers/Microsoft.Compute/disks?api-version=2018-09-30: read tcp 10.28.1.65:56094-u003e13.91.242.34:443: read: connection reset by peer`

It looks to me that "connection reset by peer" error is not treated as temporary by `IsTemporaryNetworkError`. Is there anyway I can enable retry on it for all autorest call (other than doing it in my application code itself)?

jhendrixMSFT:
I dug into this, the behavior is by design; i.e. the standard library doesn't treat ECONNRESET as a transient network error, see test `TestNotTemporaryRead` in [net_test.go](https://golang.org/src/net/net_test.go).  Do you know why the connection is being reset and does retrying resolve the issue?  I'm hesitant to treat this case as transient if the standard library doesn't (presumably they have a good reason).

tombuildsstuff:
@jhendrixMSFT 

> Do you know why the connection is being reset and does retrying resolve the issue?

FWIW we also see this through our test suite most days across a random selection of tests, presumably because of networking issues between AWS (where our CI runs) and Azure (where the ARM API's are)

mgaurav:
Thanks for looking. I am not sure why connection is being reset by Azure as there is no other information in the error logs. Retries fixes this issue, I suspect it might have to do with the load on the server side leading to connection being reset or transient network issue between application and Azure servers.

mgaurav:
azure-storage-node API also seem to treat ECONNRESET as retryable error: https://github.com/Azure/azure-storage-node/blob/master/ChangeLog.md

> ECONNRESET is handled in the retry policy filter.

> Fixed an issue which prevented transient server errors like ECONNRESET, ESOCKETTIMEDOUT and ETIMEDOUT from being retried.

jhendrixMSFT:
Thanks for the info, I will update the retry logic to include these conditions.

jhendrixMSFT:
[...]

PR #464, opened 2019-08-21
Always retry a request even if the sender returns a non-nil error

jhendrixMSFT:
Minor CHANGELOG fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v13.0.2
+
+### Bug Fixes
+
+- Always retry a request even if the sender returns a non-nil error.
+
 ## v13.0.1
 
 ## Bug Fixes
@@ -12,9 +18,9 @@
 
 The `tracing` package has been rewritten to provide a common interface for consumers to wire in the tracing package of their choice.
 What this means is that by default no tracing provider will be compiled into your program and setting the `AZURE_SDK_TRACING_ENABLED`
-environment variable will have no effect.  To enable this previous behavior you must now add the following include to your source file.
+environment variable will have no effect.  To enable this previous behavior you must now add the following import to your source file.
 ```go
-  include _ "github.com/Azure/go-autorest/tracing/opencensus"
+  import _ "github.com/Azure/go-autorest/tracing/opencensus"
 ```
 The APIs required by autorest-generated code have remained but some APIs have been removed and new ones added.
 The following APIs and variables have been removed (the majority of them were moved to the `opencensus` package).
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -26,7 +26,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
-	"net"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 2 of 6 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/450

internal/internal_test.go
```diff
@@ -0,0 +1,81 @@
...
+func (s someTempError) Timeout() bool {
+	return true
+}
+func (s someTempError) Temporary() bool {
+	return true
+}
+
+func TestIsTemporaryNetworkErrorTrue(t *testing.T) {
+	if !IsTemporaryNetworkError(someTempError{}) {
+		t.Fatal("expected someTempError to be a temporary network error")
```
devigned: You really want to fail fast :)

kahing:
I've been running this on a branch in travis and just saw another transient failure: `url.Err: *net.OpError: dial tcp 52.239.160.162:443: connect: connection refused`

Maybe instead of having a list of errno to retry, it should keep a list of errno not to retry?

kahing:
retrying on non-transient failure is better than not retrying on transient failures

kahing:
also, seems like with this PR I still see connection reset failures: https://travis-ci.org/kahing/goofys/jobs/585257461

```go
			if detailedError, ok := err.(autorest.DetailedError); ok {
				if urlErr, ok := detailedError.Original.(*url.Error); ok {
					adl2Log.Errorf("url.Err: %T: %v %v %v", urlErr.Err, urlErr.Err, urlErr.Temporary(), urlErr.Timeout())
```

that code produced this log line:

```
2019/09/15 17:42:09.184467 adlv2.ERROR url.Err: *net.OpError: read tcp 10.20.1.217:42936->52.239.160.162:443: read: connection reset by peer false false
```

mbrancato:
I agree on retrying is better than not. 

> It's an interesting idea, then the question becomes what's the list of error codes we shouldn't retry? EHOSTDOWN and EHOSTUNREACH seem like good candidates although couldn't these too be transient in the face of network connectivity problems? Seems like getting all the conditions right might be tricky (same as the changes in this PR). I'm not against the idea, just wondering which one is easier to maintain (having a list of errno to not retry kinda feels like trying to prove a negative). Of course we could just always retry on error which is what we originally used to do; it's the simplest solution but runs the risk of retrying on non-transient failures.

I think this is addressed in the retry guidance:
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#error-handling

My read of 'timeout' is to include all connection failures and a complete failure after 5 recommended attempts in the backoff process.

jhendrixMSFT:
@kahing @mbrancato I've reworked this to always retry failed requests.

kahing:
still got some errors:

```
... value autorest.DetailedError = autorest.DetailedError{Original:(*errors.errorString)(0xc0038d7930), PackageType:"storagedatalake.adl2PathClient", Method:"List", StatusCode:200, Message:"Failure responding to request", ServiceError:[]uint8(nil), Response:(*http.Response)(0xc000325b90)} ("storagedatalake.adl2PathClient#List: Failure responding to request: StatusCode=200 -- Original Error: Error occurred reading http.Response#Body - Error = 'read tcp 10.20.0.122:53872->52.239.193.98:443: read: connection reset by peer'")
```

and

```
2019/09/28 03:14:41.831574 adlv2.ERROR azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to [secure]goofys-test-ac93v4y6q13janv8?resource=filesystem: StatusCode=0 -- Original Error: adal: Failed to execute the refresh request. Error = 'Post https://login.microsoftonline.com/[secure]/oauth2/token?api-version=1.0: read tcp 10.20.0.99:55356->40.126.5.98:443: read: connection reset by peer'
```

jhendrixMSFT:
Thanks for the info.  For the first case we don't, at present, have any retry logic when reading a response body.  The retry logic for this PR is only for calling the REST API.  Can you please open a new issue to track adding retries for reading responses?  Please note that this is likely a significant design change so I don't know how fast a fix will be forthcoming.
For the second case, looking at `ServicePrincipalToken.refreshInternal()` I don't believe we actually retry the request if it fails (only when authenticating via IMDS).  I will update this PR to retry that case.

kahing:
pass rate is better but still get #470 quite often. This is definitely a step forward though

jhendrixMSFT:
CC @tombuildsstuff @katbyte

PR #476, opened 2019-10-18
token: support for a custom refresh func

tombuildsstuff:
This PR adds support for specifying a custom refresh function which allows users defining manual refresh tokens to refresh the token via a callback

Fixes https://github.com/hashicorp/go-azure-helpers/issues/22

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -350,7 +350,7 @@ type ServicePrincipalToken struct {
 	inner             servicePrincipalToken
 	refreshLock       *sync.RWMutex
 	sender            Sender
-	customRefreshFunc *TokenRefresh
+	customRefreshFunc TokenRefresh
 	refreshCallbacks  []TokenRefreshCallback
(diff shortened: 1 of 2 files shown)


autorest/adal/token.go
```diff
@@ -344,10 +347,11 @@ func (secret ServicePrincipalAuthorizationCodeSecret) MarshalJSON() ([]byte, err
...
 // ServicePrincipalToken encapsulates a Token created for a Service Principal.
 type ServicePrincipalToken struct {
-	inner            servicePrincipalToken
-	refreshLock      *sync.RWMutex
-	sender           Sender
-	refreshCallbacks []TokenRefreshCallback
+	inner             servicePrincipalToken
+	refreshLock       *sync.RWMutex
+	sender            Sender
+	customRefreshFunc *TokenRefresh
```
jhendrixMSFT: I don't think this needs * as func types can implicitly be nil.

tombuildsstuff:
@jhendrixMSFT thanks for pushing a fix for this - LGTM 👍

PR #498, opened 2020-01-09
Improve the code to support int and other types and cast them to string

thesunwave:
Hey, how about this solution @jhendrixMSFT? This code accepts interface but casts types not always right. I mean isn't this more correct to cast not only string arrays to string if that function accepts not only strings? If we accept it then we don't need to fix a swagger spec.

realates with https://github.com/Azure/azure-sdk-for-go/issues/6586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v13.3.1
+
+### Bug Fixes
+
+- Updated `autorest.AsStringSlice()` to convert slice elements to their string representation.
+
 ## v13.3.0
 
 ### New Features
--- a/autorest/utility.go
+++ b/autorest/utility.go
@@ -140,13 +140,13 @@ func MapToValues(m map[string]interface{}) url.Values {
 	return v
 }
 
-// AsStringSlice method converts interface{} to []string. This expects a
-//that the parameter passed to be a slice or array of a type that has the underlying
-//type a string.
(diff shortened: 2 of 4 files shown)


thesunwave:
is anybody here?

jhendrixMSFT:
Sorry for the delay.

While the change looks fine, the code comment states the following.
```go
// AsStringSlice method converts interface{} to []string. This expects a
// AsStringSlice method converts interface{} to []string. This expects a
//that the parameter passed to be a slice or array of a type that has the underlying
//that the parameter passed to be a slice or array of a type that has the underlying
//type a string.
```
It could be the case that the comment is incorrect, but we need to be sure this isn't going to break anything before it can be merged.  I will investigate further later in the week.

thesunwave:
Thank you @jhendrixMSFT.

jhendrixMSFT:
I dug into this a bit more.  Per issue https://github.com/Azure/go-autorest/issues/207, `AsStringSlice()` was introduced to support cognitive services.  While the issue doesn't mention the specifics, I believe this was to fix cases where collections of enums are used as parameters.  The original code simply would type-assert `interface{}` to `[]string` which would panic for slices of enums.  

Since the underlying type of enums is string, this function was introduced to handle that case.  Unfortunately collections of integers were never considered, either they didn't exist at the time or their failure was obscured by the enum case; and since this change "fixed" integers we never noticed it was spewing out bad values.

I've updated the code comment and CHANGELOG with info on the fix.

PR #503, opened 2020-02-04
Enable exponential back-off when retrying on 429

jhendrixMSFT:

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -295,7 +295,8 @@ func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempt
 			return resp, err
 		}
 		delayed := DelayWithRetryAfter(resp, r.Context().Done())
-		// enforce a 2 minute cap when not counting 429 if there isn't one.
+		// enforce a 2 minute cap between requests when 429 status codes are
+		// not going to be counted as an attempt and when the cap is 0.
 		// this should only happen in the absense of a retry-after header.
 		if !count429 && cap == 0 {
 			cap = 2 * time.Minute
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -297,7 +297,7 @@ func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempt
 		delayed := DelayWithRetryAfter(resp, r.Context().Done())
 		// enforce a 2 minute cap between requests when 429 status codes are
 		// not going to be counted as an attempt and when the cap is 0.
-		// this should only happen in the absense of a retry-after header.
+		// this should only happen in the absence of a retry-after header.
 		if !count429 && cap == 0 {
 			cap = 2 * time.Minute
 		}
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -283,7 +283,7 @@ func DoRetryForStatusCodesWithCap(attempts int, backoff, cap time.Duration, code
 func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempts int, backoff, cap time.Duration, codes ...int) (resp *http.Response, err error) {
 	rr := NewRetriableRequest(r)
 	// Increment to add the first call (attempts denotes number of retries)
-	for attempt := 0; attempt < attempts+1; {
+	for attempt, delayCount := 0, 0; attempt < attempts+1; {
 		err = rr.Prepare()
 		if err != nil {
 			return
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
CC @tombuildsstuff @katbyte @feiskyer 

Note that by default there is no cap on exponential back-off, so I'm concerned that we'll encounter unreasonably long polling delays.  I'm thinking that for the short-term I hard code a 2-minute cap on the delay, and long-term make the cap configurable via a property on the client (the latter requires changes to the code generator thus is much more involved).

At any rate this change should prevent excessive polling on 429 that don't send a retry-after header.

feiskyer:
@jhendrixMSFT we have added a cache for retry-after values in k8s cloud provider with our own back-off retries, hence it could avoid sending more requests on 429.

jhendrixMSFT:
@feiskyer per the internal email thread I got the impression that the handling of retries in the SDK (specifically around 429) is still problematic and that k8s is looking for a long-term fix in go-autorest (this change is likely just a small part of that).  Is this not the case?

feiskyer:
@jhendrixMSFT yep, in the long-term, we hope the changes from k8s could be adopted into both go-autorest and Go SDK.

tombuildsstuff:
@jhendrixMSFT I'd agree a hard-cap would be practical, however 2m between polls seems kinda excessive, as such I'm wondering if the hard-cap wants to be 1m rather than 2m? But otherwise sounds good to me 👍

tombuildsstuff:
@jhendrixMSFT from our side we're planning to do another release next week, so I'm assuming we can bundle this into that - or is this delay causing larger issues/need to go out before then?

jhendrixMSFT:
@tombuildsstuff I think next week should be fine.  I picked 2 minutes as it seems to be a reasonable middle-ground (when being throttled I've seen the delays be much larger) and fewer requests while being throttled is better.  And really the cap is only there if no `retry-after` header was specified which should be the exception.

autorest/sender.go
```diff
@@ -295,14 +295,22 @@ func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempt
 			return resp, err
 		}
 		delayed := DelayWithRetryAfter(resp, r.Context().Done())
-		if !delayed && !DelayForBackoffWithCap(backoff, cap, attempt, r.Context().Done()) {
+		// enforce a 2 minute cap when not counting 429 if there isn't one.
```
catalinaperalta: I feel like this comment is a bit confusing. If I understand what we're trying to say here, I would phrase it like: enforce a 2 minute cap between requests when 429 status codes are not going to be considered and when the cap is 0.

I know it's more verbose, but it was a bit confusing at first.

autorest/sender.go
```diff
@@ -295,14 +295,22 @@ func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempt
...
 			return resp, r.Context().Err()
 		}
 		// when count429 == false don't count a 429 against the number
 		// of attempts so that we continue to retry until it succeeds
 		if count429 || (resp == nil || resp.StatusCode != http.StatusTooManyRequests) {
 			attempt++
 		}
+		// delay count is tracked separately from attempts to
+		// ensure that 429 participates in exponential back-off
+		delayCount++
```
catalinaperalta: Is the TestDelayWithRetryAfterWithSuccess the only test to step through for both scenarios?

jhendrixMSFT: There is also `TestDoRetryForStatusCodes_Cancel429` which tests the infinite retries on 429.

autorest/sender.go
jhendrixMSFT: updated comment

autorest/sender.go
```diff
@@ -283,7 +283,7 @@ func DoRetryForStatusCodesWithCap(attempts int, backoff, cap time.Duration, code
 func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempts int, backoff, cap time.Duration, codes ...int) (resp *http.Response, err error) {
 	rr := NewRetriableRequest(r)
 	// Increment to add the first call (attempts denotes number of retries)
-	for attempt := 0; attempt < attempts+1; {
+	for attempt, delayCount := 0, 0; attempt < attempts+1; {
```
Petermarcu: What is the default or common value for attempts?

autorest/sender.go
```diff
@@ -295,14 +295,23 @@ func doRetryForStatusCodesImpl(s Sender, r *http.Request, count429 bool, attempt
 			return resp, err
 		}
 		delayed := DelayWithRetryAfter(resp, r.Context().Done())
-		if !delayed && !DelayForBackoffWithCap(backoff, cap, attempt, r.Context().Done()) {
+		// enforce a 2 minute cap between requests when 429 status codes are
+		// not going to be counted as an attempt and when the cap is 0.
+		// this should only happen in the absense of a retry-after header.
+		if !count429 && cap == 0 {
```
Petermarcu: Would this be more readable and also do less in the delayed case if you had a !delayed case and then did this in there?

autorest/sender.go
jhendrixMSFT: The default is 3, specified here https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L39 however it can be configured per client.

autorest/sender.go
jhendrixMSFT: I had thought about that, I think the amount of work done in either case is negligible.  I omitted it as I didn't want to make the condition too complicated.  Not against adding it thought if we think it's an improvement.

PR #506, opened 2020-02-14
Enable specifying SendDecorators on a client

jhendrixMSFT:
Added field SendDecorators to autorest.Client; this value will be
preferred over the default SendDecorators.
Added method Client.Send() which includes logic for using custom
SendDecorator chains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## v13.4.0
+
+## New Features
+
+- Added field `SendDecorators` to the `Client` type.  This can be used to specify a custom chain of SendDecorators per client.
+- Added method `Client.Send()` which includes logic for selecting the preferred chain of SendDecorators.
+
 ## v13.3.3
 
 ### Bug Fixes
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -179,6 +179,11 @@ type Client struct {
 
 	// Set to true to skip attempted registration of resource providers (false by default).
 	SkipResourceProviderRegistration bool
+
+	// SendDecorators can be used to override the default chain of SendDecorators.
+	// This can be used to specify things like a custom retry SendDecorator.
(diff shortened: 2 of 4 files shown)


jhendrixMSFT:
@feiskyer PTAL, will wait for your approval before merging.

jhendrixMSFT:
Accompanying codegen change: https://github.com/Azure/autorest.go/pull/261

feiskyer:
This looks great. 

@jhendrixMSFT could you share an example how could it be used for Azure Go SDK clients?

jhendrixMSFT:
@feiskyer here's how you would set a custom retry policy on a client.  There are similar examples in the test added in the PR.

```go
func MyCustomRetryPolicy(attempts int, backoff time.Duration, codes ...int) autorest.SendDecorator {
    return func(s autorest.Sender) autorest.Sender {
        return autorest.SenderFunc(func(r *http.Request) (*http.Response, error) {
            // do work before sending the request to the next SendDecorator here ...
 
            // pass the request to the next SendDecorator
            resp, err := s.Do(r)
 
            // do something with the response
            if resp.StatusCode == http.StatusTooManyRequests {
                // being throttled, handle as required....
            }
 
            // return response/error to my caller
            return resp, err
        })
    }
}
 
func main() {
    a, err := auth.NewAuthorizerFromEnvironment()
    if err != nil {
        panic(err)
    }
    client := redis.NewClient(subscriptionID)
    client.SendDecorators = []autorest.SendDecorator{MyCustomRetryPolicy(3, 5*time.Second, http.StatusInternalServerError)}
    client.Authorizer = a
 
    // sync example
    result, err := client.CheckNameAvailability(context.Background(), redis.CheckNameAvailabilityParameters{})
    if err != nil && result.IsHTTPStatus(http.StatusTooManyRequests) {
        // being throttled ...
    }
}
```

feiskyer:
@jhendrixMSFT cool, thanks for the excellent example

PR #507, opened 2020-02-19
Bugfixes v11.1

jhendrixMSFT:

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -194,6 +194,7 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
 				if err != nil {
 					return resp, err
 				}
+				drainResponseBody(resp)
 				resp, err = s.Do(rr.Request())
 				if err == nil {
 					return resp, err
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
@feiskyer PTAL

autorest/sender.go
```diff
@@ -233,30 +241,48 @@ func DoRetryForStatusCodes(attempts int, backoff time.Duration, codes ...int) Se
...
 	if resp == nil {
 		return false
 	}
-	retryAfter, _ := strconv.Atoi(resp.Header.Get("Retry-After"))
-	if resp.StatusCode == http.StatusTooManyRequests && retryAfter > 0 {
+	var dur time.Duration
+	ra := resp.Header.Get("Retry-After")
+	if retryAfter, _ := strconv.Atoi(ra); retryAfter > 0 {
+		dur = time.Duration(retryAfter) * time.Second
+	} else if t, err := time.Parse(time.RFC1123, ra); err == nil {
```
catalinaperalta: IF err != nil  then we return false and there will be no delay?

jhendrixMSFT: The exponential back-off delay will be used instead.

PR #512, opened 2020-03-10
auth/storage: support for the SharedKey implementation used by the Storage Accounts Data Plane API

tombuildsstuff:
Well this is a fun one.

After a bunch of debugging it turns out API's in the root of a Storage Data Plane API use a slightly different, undocumented implementation of SharedKey authorization. Specifically [this API](https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-service-properties) and [this API](https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob-service-properties) - but presumably the others in the root too.

Whilst this is unfortunate - attempting to update this API to [match the definition](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key) at this point in time would be a breaking change and thus is a less ideal solution.

Instead this PR introduces a new mode to the SharedKey Authorizer which supports the varied authorization method used by this endpoint.

---

Before this PR, when attempting to hit the Account Endpoint using a SharedKey or SharedKeyLite authorizer with this Request using the SharedKey authorizer:

```
GET /?comp=properties&restype=service HTTP/1.1
Host: unlikely23exst2acctazcnp.blob.core.windows.net
User-Agent: Go/go1.13.5 (amd64-darwin) go-autorest/v13.3.0 tombuildsstuff/giovanni/v0.8.0 storage/2018-11-09
Authorization: SharedKey unlikely23exst2acctazcnp:{removed}
X-Ms-Date: Tue, 10 Mar 2020 10:04:41 GMT
X-Ms-Version: 2018-11-09
Accept-Encoding: gzip
```

returns the following Response..

```
HTTP/1.1 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
Content-Length: 687
Content-Type: application/xml
Server: Microsoft-HTTPAPI/2.0
x-ms-request-id: a3addf0a-b01e-002e-09c3-f6cf21000000
x-ms-error-code: AuthenticationFailed
Date: Tue, 10 Mar 2020 10:04:40 GMT
Connection: keep-alive

<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:a3addf0a-b01e-002e-09c3-f6cf21000000
Time:2020-03-10T10:04:41.2097467Z</Message><AuthenticationErrorDetail>The MAC signature found in the HTTP request '{removed}' is not the same as any computed signature. Server used following string to sign: 'GET











x-ms-date:Tue, 10 Mar 2020 10:04:41 GMT
x-ms-version:2018-11-09
/unlikely23exst2acctazcnp/
comp:properties
restype:service'.</AuthenticationErrorDetail></Error>
```

---

With these changes, and this new Authorizer - the following Request:

```
PUT /?comp=properties&restype=service HTTP/1.1
Host: unlikely23exst2acctggtll.blob.core.windows.net
User-Agent: Go/go1.13.5 (amd64-darwin) go-autorest/v13.3.0 tombuildsstuff/giovanni/v0.8.0 storage/2018-11-09
Content-Length: 146
Authorization: SharedKey unlikely23exst2acctggtll:{removed}
X-Ms-Date: Tue, 10 Mar 2020 10:17:00 GMT
X-Ms-Version: 2018-11-09
Accept-Encoding: gzip

<?xml version="1.0" encoding="UTF-8"?>
<StorageServiceProperties><StaticWebsite><Enabled>true</Enabled></StaticWebsite></StorageServiceProperties>
```

.. returns the following Response:

```
HTTP/1.1 202 Accepted
Content-Length: 0
Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: aca9eb77-301e-007a-67c5-f6b5af000000
x-ms-version: 2018-11-09
Date: Tue, 10 Mar 2020 10:17:02 GMT
Connection: keep-alive
```

### DIFF
--- a/autorest/authorization_storage.go
+++ b/autorest/authorization_storage.go
@@ -35,9 +35,6 @@ const (
 	// SharedKey is used to authorize against blobs, files and queues services.
 	SharedKey SharedKeyType = "sharedKey"
 
-	// SharedKey is used to authorize against the account.
-	SharedKeyForAccount SharedKeyType = "sharedKeyAccount"
-
 	// SharedKeyForTable is used to authorize against the table service.
 	SharedKeyForTable SharedKeyType = "sharedKeyTable"
 
@@ -130,14 +127,6 @@ func buildSharedKey(accName string, accKey []byte, req *http.Request, keyType Sh
 		date := time.Now().UTC().Format(http.TimeFormat)
 		req.Header.Set(headerXMSDate, date)
 	}
-
-	if keyType == SharedKeyForAccount {
-		// ensure a content length is set if appropriate
-		if req.Header.Get(headerContentLength) == "" {
-			req.Header.Set("Content-Length", fmt.Sprintf("%d", int(req.ContentLength)))
-		}
-	}
-
 	canString, err := buildCanonicalizedString(req.Method, req.Header, canRes, keyType)
 	if err != nil {
 		return "", err
@@ -156,9 +145,6 @@ func buildCanonicalizedResource(accountName, uri string, keyType SharedKeyType) 
 	if accountName != storageEmulatorAccountName {
 		cr.WriteString("/")
 		cr.WriteString(getCanonicalizedAccountName(accountName))
-		if keyType == SharedKeyForAccount {
-			cr.WriteString("/")
-		}
 	}
 
 	if len(u.Path) > 0 {
@@ -166,6 +152,9 @@ func buildCanonicalizedResource(accountName, uri string, keyType SharedKeyType) 
 		// the resource's URI should be encoded exactly as it is in the URI.
 		// -- https://msdn.microsoft.com/en-gb/library/azure/dd179428.aspx
 		cr.WriteString(u.EscapedPath())
+	} else {
+		// a slash is required to indicate the root path
+		cr.WriteString("/")
 	}
 
 	params, err := url.ParseQuery(u.RawQuery)
@@ -174,7 +163,7 @@ func buildCanonicalizedResource(accountName, uri string, keyType SharedKeyType) 
 	}
 
 	// See https://github.com/Azure/azure-storage-net/blob/master/Lib/Common/Core/Util/AuthenticationUtility.cs#L277
-	if keyType == SharedKey || keyType == SharedKeyForAccount {
+	if keyType == SharedKey {
 		if len(params) > 0 {
 			cr.WriteString("\n")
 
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
Worth saying this ends up fixing the associated test on our end:

```
$ TF_ACC=1 go test -v ./azurerm/internal/services/storage/tests/ -timeout=60m -run=TestAccAzureRMStorageAccount_staticWebsiteEnabled
=== RUN   TestAccAzureRMStorageAccount_staticWebsiteEnabled
=== PAUSE TestAccAzureRMStorageAccount_staticWebsiteEnabled
=== CONT  TestAccAzureRMStorageAccount_staticWebsiteEnabled
--- PASS: TestAccAzureRMStorageAccount_staticWebsiteEnabled (214.03s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/storage/tests	214.097s
```

tombuildsstuff:
ping @jhendrixMSFT :)

jhendrixMSFT:
Sorry for the delay.

I think this is a bug in the string-to-sign implementation.  We shouldn't have to special-case this based on the root, just if `content-length` is zero.  Take a look at the [implementation](https://github.com/Azure/azure-sdk-for-go/blob/master/storage/authorization.go#L144) in the legacy storage package as it works as expected.

autorest/azure/auth/go.mod
```diff
@@ -4,8 +4,10 @@ go 1.12
...
 	github.com/Azure/go-autorest/autorest v0.9.2
-	github.com/Azure/go-autorest/autorest/adal v0.7.0
+	github.com/Azure/go-autorest/autorest/adal v0.8.0
 	github.com/Azure/go-autorest/autorest/azure/cli v0.3.0
 	github.com/dimchansky/utfbom v1.1.0
-	golang.org/x/crypto v0.0.0-20190418165655-df01cb2cc480
+	golang.org/x/crypto v0.0.0-20191206172530-e9b2fee46413
 )
+
+replace github.com/Azure/go-autorest/autorest => github.com/tombuildsstuff/go-autorest/autorest v0.9.3-hashi-auth
```
jhendrixMSFT, line 13: Remove `replace` directive.

autorest/authorization_storage.go
```diff
@@ -35,6 +35,9 @@ const (
 	// SharedKey is used to authorize against blobs, files and queues services.
 	SharedKey SharedKeyType = "sharedKey"
 
+	// SharedKey is used to authorize against the account.
+	SharedKeyForAccount SharedKeyType = "sharedKeyAccount"
```
jhendrixMSFT, line 39: I think this can be removed, just add content length in the string-to-sign input if it's not equal to zero.

jhendrixMSFT:
@tombuildsstuff terribly sorry for the delay on this.  I dug into it further, the canonicalized resource requires that the root path is never empty.  I adapted the fix from [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/zc_credential_shared_key.go#L179).  It's handled slightly different in the legacy storage package but the end result is the same.  I've verified with `giovanni` that this resolves the issue.

PR #518, opened 2020-05-11
Fix check for App Service Environment

jhendrixMSFT:
The presence of the ASE envrionment vars does not indicate that the
caller is actually authenticating in this way.  Instead, look at the
token endpoint URL to see if it matches the ASE endpoint.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -650,6 +650,8 @@ func GetMSIVMEndpoint() (string, error) {
 	return msiEndpoint, nil
 }
 
+// NOTE: this only indicates if the ASE environment credentials have been set
+// which does not necessarily mean that the caller is authenticating via ASE!
 func isAppService() bool {
 	_, asMSIEndpointEnvExists := os.LookupEnv(asMSIEndpointEnv)
 	_, asMSISecretEnvExists := os.LookupEnv(asMSISecretEnv)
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Hmm, I read the docs for `resp.Body` and it will never be nil so I don't think this is the correct fix.

ArcturusZhang:
Related issue: https://github.com/terraform-providers/terraform-provider-azurerm/issues/6773

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/547

autorest/adal/token.go
```diff
@@ -850,11 +852,28 @@ func (spt *ServicePrincipalToken) getGrantType() string {
 }
 
 func isIMDS(u url.URL) bool {
-	imds, err := url.Parse(msiEndpoint)
+	return isMSIEndpoint(u) == true || isASEEndpoint(u) == true
```
catalinaperalta, line 855: I think the requests for IMDS and ASE differ, don't they?

autorest/adal/token.go
```diff
@@ -650,6 +650,8 @@ func GetMSIVMEndpoint() (string, error) {
 	return msiEndpoint, nil
 }
 
+// NOTE: this only indicates if the ASE environment credentials have been set
+// which does not necessarily mean that the caller is authenticating via ASE!
```
catalinaperalta, line 654: In track 2 we do make this assumption across languages I believe.

autorest/adal/token.go
jhendrixMSFT: The endpoints differ yes, or did you mean something else?

autorest/adal/token.go
jhendrixMSFT: Good to know, we'll need to ensure this scenario isn't broken in track 2.

autorest/adal/token.go
jhendrixMSFT: And really, this function is poorly named.  It was created before we added support for ASE.

autorest/adal/token.go
catalinaperalta: Yes, I'll bring this up with the identity crew. Did someone complain about this? Like users had these environment vars set and did not need App Service authentication?

catalinaperalta: Nevermind, I'm reading through the issue.

autorest/adal/token.go
catalinaperalta: Yes the function name is misleading. And yes I was wondering if those two MSI cases should be used interchangeably.

catalinaperalta: I checked track2, maybe it's not essential for the requests in these cases but the headers do differ between IMDS and ASE scenarios.

autorest/adal/token.go
```diff
@@ -354,6 +354,7 @@ type ServicePrincipalToken struct {
 	customRefreshFunc TokenRefresh
 	refreshCallbacks  []TokenRefreshCallback
 	// MaxMSIRefreshAttempts is the maximum number of attempts to refresh an MSI token.
+	// Settings this to a value less than 1 will use the default value.
```
jhendrixMSFT, line 357: The other option would be to return an error during token acquisition.  While this approach is more forgiving, it could mask other bugs (e.g. the original issue would have been much harder to diagnose).

jhendrixMSFT: However the Go idiom is typically to have a meaningful zero-value which usually translates to "give me the default behavior".

catalinaperalta: I'm more inclined to returning the error exactly to avoid masking other bugs, but for now I think it's fine to leave it as is with the comment.

PR #524, opened 2020-05-29
Fix go modules import issues

### LINKED ISSUE #481: Versioning of Go-AutoRest
👋🏻

As a consumer of this repository (via both [the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) and [Giovanni](https://github.com/tombuildsstuff/giovanni)) whilst I can understand the desire to version each package in this repository independently I think it's causing more complexity than it's solving since there's dependencies between packages.

From our side we've lost a substantial amount of time (days) over the past few months trying to diagnose versioning issues with go-autorest and the Azure SDK for Go. Ultimately we've had to roll back to using an older version of the Azure SDK for Go / go-autorest, since switching to the versioning per package breaks vendoring Terraform Core into downstream codebases, unless the downstream codebases also include a `replace` statement in their `go.mod` files - which unfortunately isn't a viable solution for us.

My concern is that these issues will be exacerbated when the Azure SDK for Go moves to a Go Module per Service Package (e.g. `./services/foobar/mgmt/2019-01-01`); where different service packages/modules depend on different versions of this package.

Given these issues are caused by there being two different tagging approaches in this repository (which can't be removed at this point since they'll be cached via the Go Module Cache) - I'm wondering if it'd be worth taking a different approach instead.

Since this library depends on itself, I believe that versioning per package may be causing more issues than it's solving. In particular whilst I understand grouping everything into a single module means the SemVer for this library may need to be bumped ([which @jhendrixMSFT previously outlined here](https://github.com/Azure/go-autorest/issues/414#issuecomment-522063283)) I'm wondering if bumping the major version is a concern for this library in practice?

Given the problems above, I'm wondering if it'd be worth essentially duplicating this repository (preferably with, but it might be less hassle without the history) but in either case without the existing tags. This'd allow this repository to be versioned starting from (say) 1.0 but critically as single unit (in the root).

There's two affects on this on downstream consumers; downstream SDK's, such as:

* [the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go)
* [Giovanni](https://github.com/tombuildsstuff/giovanni)
* [hashicorp/go-azure-helpers](https://github.com/hashicorp/go-azure-helpers)
* (others)

which (technically) would require a major version bump.

Software proudcts on the other hand would only require a patch release, including:

* Kubernetes
* Prometheius
* Terraform Core & downstream Providers that vendor it
* (others)

To ease the transition for slower moving dependencies it could be beneficial to use Type Aliases in this repository to the types in the new repository (in the same way Profiles do in the Azure SDK for Go), but I don't think that's a blocker (and if it is it's easy enough to retroactively add it).

---

Whilst I appreciate this is a big ask/change - since the versioning issues has affected multiple downstream projects (including [Terraform](https://github.com/hashicorp/terraform/pull/22532) and Kubernetes which in turn has affected both [CoreDNS](https://github.com/coredns/coredns/pull/3258) and [Rancher](https://github.com/rancher/rio/pull/451) (and I believe Prometheus too, albeit I can't find the link now?)) - I'm wondering if splitting this out into it's own repository provides the simplest path forward.

Thoughts?

AdamSLevy:
- Remove all nested go.mod and go.sum files.
- Add single top level go.mod and go.sum files with module import path
  "github.com/Azure/go-autorest/v15"
- Replace all internal import "github.com/Azure/go-autorest/..." paths
  with "github.com/Azure/go-autorest/v15/..."
- Run `go test && go mod tidy`

Fix #481 

 - [ x ] I've tested my changes, adding unit tests if applicable.
 - [ x ] I've added Apache 2.0 Headers to the top of any new source files.
 - [ x ] I'm submitting this PR to the `dev` branch, except in the case of urgent bug fixes warranting their own release.


jhendrixMSFT:
Having separate modules it to allow user-code (e.g. `adal` package) to version independently from code used purely by the code generator (e.g. `validation` package).  I would prefer to fix what we have if possible, and if we conclude it cannot be fixed then we can turn this into one big module.

AdamSLevy:
I highly encourage you to read up on Golang modules then. From everything I read in the issue I strongly feel that you are trying to fit a square peg in a round hole on this effort. This is further evidenced by the numerous "go_mod_tidy_hack.go" files. Go modules works very well when you work with it, but you are going against the grain. go modules has no way to actually know what to do with these mixed go.mod files that have not been properly kept up to date.

Code that is developed together, is interdependent, and is effectively distributed and released together, needs to be versioned together. A module is intended to version a set of interdependent and co-developed packages together. 

But regardless of your opinion on where the API boundaries should lay, the current approach of putting a go.mod file in each sub-package is completely untenable. It is causing rather large headaches downstream, and the intention is not clear to outside developers. 

Why would I ever want to use two different, possibly incompatible (how would I even know?), versions of the adal code and the validation code? Moreover, why does a user need to import the validation code at all? If they don't, then just move that package to an internal folder, and then make whatever changes you need to it without worrying about changing its API for consumers.

If you aren't ready to have a stable API, that's ok too, but you should re-release this repo under a new import path and stick with a major version of v0, which does not promise any API stability to users. 

Additionally version tags allow you to freely develop with breaking API changes on master. go modules will only pull the latest tagged version. 

One important rule that you are breaking, is that major version changes require a different import path. This is solved by adding `/vXX/` right after the module import path and is based on the repo version tag. But with each package its own module, each not declaring their major version in their own import path in the go.mod file, `go mod` has no way to know what the version of each sub package should be, or which combinations of interdependent versions are acceptable.

I hope you reconsider changing this repo to be inline with the rules of go modules. At the moment this package is very difficult to use confidently.

jhendrixMSFT:
> Code that is developed together, is interdependent, and is effectively distributed and released together, needs to be versioned together. A module is intended to version a set of interdependent and co-developed packages together.

I agree with this.  What I was stating earlier is that not all of the packages in this repo are interdependent and should be allowed to version at their own cadence.

> But regardless of your opinion on where the API boundaries should lay, the current approach of putting a go.mod file in each sub-package is completely untenable. It is causing rather large headaches downstream, and the intention is not clear to outside developers.

This is a good point.  If we continue to keep multiple modules I believe this can be solved with documentation and examples.

> Legitimately asking: Why would I ever want to use two different, possibly incompatible (how would I even know?), versions of the adal code and the validation code? Moreover, why does a user need to import the validation code at all? If they don't, then just move that package to an internal folder, and then make whatever changes you need to it without worrying about changing its API for consumers.

The `validation` package is used by methods created by our code generator, see [here](https://github.com/Azure/azure-sdk-for-go/blob/master/services/alertsmanagement/mgmt/2019-03-01/alertsmanagement/smartdetectoralertrules.go#L62) for an example; there's nothing in that package that's useful for SDK consumers.  Mixing it (let's call this "non-user code") in with user-consumable code (e.g. `auth` and `to` packages are the most common) means that user-code has to pay the tax of a major version bump when a breaking change is introduced in non-user code.  It would be ideal to allow user and non-user code to independently version.  (Really, the non-user code should have been internal packages in the SDK repo which likely would have avoided this).

> One important rule that you are breaking, is that major version changes require a different import path. This is solved by adding /vXX/ right after the module import path and is based on the repo version tag. But with each package its own module, each not declaring their major version in their own import path in the go.mod file, go mod has no way to know what the version of each sub package should be, or which combinations of interdependent versions are acceptable.

All of the modules in this repo are v0 which precludes the need for a major version suffix.  This was an intentional decision to test out the current strategy (without introducing any breaking changes due to semantic import versioning), providing a bit of latitude if we need to refactor the module boundaries.

Multi-module repositories are supported, and there is real value in having user and non-user code independently version.  So before we just abandon this split I think it's worth due diligence to ensure if it will/won't reliably work.

jhendrixMSFT:
This has been fixed via https://github.com/Azure/go-autorest/pull/528 which is now in master.

Issue #414, opened 2019-06-27
Azure go-autorest causing ambiguous import


27149chen:
Got an ambiguous import error when run go get ./... under a go module enabled project:

```
build github.com/.../cmd/manager: cannot load github.com/Azure/go-autorest/autorest: ambiguous import: found github.com/Azure/go-autorest/autorest in multiple modules:
	github.com/Azure/go-autorest v11.2.8+incompatible (/go/pkg/mod/github.com/!azure/go-autorest@v11.2.8+incompatible/autorest)
	github.com/Azure/go-autorest/autorest v0.3.0 (/go/pkg/mod/github.com/!azure/go-autorest/autorest@v0.3.0)
```
go-autorest is introduced by client-go. 
here is the line in my go.mod:
```
github.com/Azure/go-autorest/autorest v0.3.0 // indirect
```
but in the go.sum, there are many lines:
```
github.com/Azure/go-autorest v11.2.8+incompatible h1:Q2feRPMlcfVcqz3pF87PJzkm5lZrL+x6BDtzhODzNJM=
github.com/Azure/go-autorest v11.2.8+incompatible/go.mod h1:r+4oMnoxhatjLLJ6zxSWATqVooLgysK6ZNox3g/xq24=
github.com/Azure/go-autorest/autorest v0.3.0 h1:yOmXNB2qa2Kx40wMZB19YyafzjCHacXPk8u0neqa+M0=
github.com/Azure/go-autorest/autorest v0.3.0/go.mod h1:AKyIcETwSUFxIcs/Wnq/C+kwCtlEYGUVd7FPNb2slmg=
github.com/Azure/go-autorest/autorest/adal v0.1.0 h1:RSw/7EAullliqwkZvgIGDYZWQm1PGKXI8c4aY/87yuU=
github.com/Azure/go-autorest/autorest/adal v0.1.0/go.mod h1:MeS4XhScH55IST095THyTxElntu7WqB7pNbZo8Q5G3E=
github.com/Azure/go-autorest/autorest/date v0.1.0/go.mod h1:plvfp3oPSKwf2DNjlBjWF/7vwR+cUD/ELuzDCXwHUVA=
github.com/Azure/go-autorest/autorest/mocks v0.1.0/go.mod h1:OTyCOPRA2IgIlWxVYxBee2F5Gr4kF2zd2J5cFRaIDN0=
[...]

Issue #413, opened 2019-06-25
Cannot bump dependencies in Kubernetes projects due to census-instrumentation/opencensus-proto


justaugustus:
I've tried several combinations of versions of `azure-sdk-for-go` and `go-autorest` but have been unsuccessful in updating dependencies for the `cluster-api-provider-azure` Kubernetes project.

Here are the output logs for the current PR I have open to test (https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/216):

```
$ time make test
GOPATH=/home/augustus/go bazel run //:generate 
Starting local Bazel server and connecting to it...
INFO: Analysed target //:generate (17 packages loaded, 168 targets configured).
INFO: Found 1 target...
Target //:generate up-to-date:
  bazel-bin/generate_script
  bazel-bin/generate
INFO: Elapsed time: 21.646s, Critical Path: 0.14s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
make dep-ensure
make[1]: Entering directory '/home/augustus/go/src/sigs.k8s.io/cluster-api-provider-azure'
Repository is set up correctly
INFO: Analysed target //:dep (57 packages loaded, 7728 targets configured).
INFO: Found 1 target...
Target //:dep up-to-date:
  bazel-bin/dep_script
  bazel-bin/dep
INFO: Elapsed time: 5.101s, Critical Path: 0.28s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
make gazelle
make[2]: Entering directory '/home/augustus/go/src/sigs.k8s.io/cluster-api-provider-azure'
bazel run //:gazelle 
INFO: Analysed target //:gazelle (25 packages loaded, 124 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 2.352s, Critical Path: 0.33s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
gazelle: open /home/augustus/go/src/sigs.k8s.io/cluster-api-provider-azure/vendor/k8s.io/kubernetes/BUILD.bazel: no such file or directory
make[2]: Leaving directory '/home/augustus/go/src/sigs.k8s.io/cluster-api-provider-azure'
[...]

PR #528, opened 2020-06-17
Fix Go module ambiguous import errors

miquella:
 - [X] I'm submitting this PR to the `dev` branch, except in the case of urgent bug fixes warranting their own release.

----

This is an extension of the mitigations introduced in #455. Unfortunately, the original mitigations didn't address the primary cause of ambiguous import errors: the github.com/Azure/go-autorest module.

The issue stems from the fact that old versions of the root module (github.com/Azure/go-autorest) provide the same packages as the newer submodules.

To correct this situation, the _root module_ needs to be upgraded to a version that no longer provides those packages (a version where the submodules are present). Fortunately, the submodules can be leveraged to provide the necessary version bump.

See also: https://github.com/Azure/go-autorest/issues/414#issuecomment-521751306

----

Caveat: in order for this to work, an importable version of the root package needs to be referenceable.

PR #527 makes the root package importable. It was submitted separately to allow it to be released before this PR is merged.

The `go.mod` files assume that this importable version will be referenceable as `v14.2.0`. If the version where the importable package is available ends up being different, these files will need to be updated.

See also: #395, #413, #414, #455, #481, #524


miquella:
Upgrading to a version of a submodule that includes this fix should resolve any latent "ambiguous import" errors.

miquella:
@jhendrixMSFT: I believe this will resolve the "ambiguous import" errors that keep getting reported :crossed_fingers:

jhendrixMSFT:
Thanks for putting this together, I believe you're right.  Given that this is difficult to test, I'm going over docs and trying to find other examples to convince myself this is correct and won't make things worse (which is my bigger fear).

miquella:
I very much agree that I don't want to make things worse! :sweat_smile:

In light of that, I was actually very careful to not modify the `go.mod` files directly. Instead, I updated the `go_mod_tidy_hack.go` files to reference the root module (and added a few missing `go_mod_tidy_hack.go` files), then did the following in each submodule directory:

1. `go mod edit -require github.com/Azure/go-autorest@v14.2.0+incompatible`
2. `go mod edit -replace github.com/Azure/go-autorest=$HOME/src/github.com/Azure/go-autorest` (since version `v14.2.0` doesn't exist yet, we have to use a temporary replacement)
3. `go mod tidy -v`
4. `go mod edit -dropreplace github.com/Azure/go-autorest`

This was modeled directly after the procedure outlined in the [commit message](https://code-review.googlesource.com/c/gocloud/+/43891) mentioned by @bcmills in #414.

----

As a small bit of background: I was chatting with @jayconrod in the Gophers Slack yesterday and something he said tipped me off (emphasis mine):

> In order to avoid this ambiguity, we recommend when splitting a project into submodules that each submodule requires the **original module** at a version after the split. That way, the author can guarantee that **each package is provided by no more than one module**.

After reading that and then re-reading the discussion in #414, I am reasonably confident in this solution.

jhendrixMSFT:
I'm working on validating this today, assuming all goes as planned should have both your PRs merged EOD.

miquella:
Sounds great! If there's anything else I can help with or explain, please let me know.

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/481

miquella:
I just wanted to follow up to confirm that upgrading to the versions of the submodules released today immediately resolved the ambiguous import errors I was running into.

Thank you, again, for helping get this through, @jhendrixMSFT!

PR #546, opened 2020-08-11
Fix decoding of certificates with intermediate CAs

jhendrixMSFT:
pkcs12.Decode will fail if a certificate contains more than two safe
bags which is the case for certificates with intermediate authorities.
Switch to using pkcs12.ToPEM, then manually decoding the certificate and
private key blocks.
In CheckForUserCompletionWithContext, return a better error message when
available, falling back to ErrDeviceGeneric.

### DIFF
--- a/autorest/adal/persist.go
+++ b/autorest/adal/persist.go
@@ -86,8 +86,9 @@ func SaveToken(path string, mode os.FileMode, token Token) error {
 }
 
 // DecodePfxCertificateData extracts the x509 certificate and RSA private key from the provided PFX data.
-// The PFX data must contain a local certificate along with a private key or an error is returned.
-// If the certificate is not password protected pass the empty string for password.
+// The PFX data must contain a private key along with a certificate whose public key matches that of the
+// private key or an error is returned.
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Part of the fix for https://github.com/Azure/azure-sdk-for-go/issues/5412

autorest/adal/persist.go
```diff
@@ -71,3 +84,50 @@ func SaveToken(path string, mode os.FileMode, token Token) error {
...
+		return nil, nil, err
+	}
+	// first extract the private key
+	var priv *rsa.PrivateKey
+	for _, block := range blocks {
+		if block.Type == "PRIVATE KEY" {
+			priv, err = x509.ParsePKCS1PrivateKey(block.Bytes)
+			if err != nil {
+				return nil, nil, err
+			}
```
catalinaperalta, line 103: Should this break here?

jhendrixMSFT: Yes I just fixed it.

jhendrixMSFT:
@schaabs to also review

Issue #542, opened 2020-07-26
It is not possible to optimistically use MSI from this SDK


colemickens:
This is related to #541 but I think warrants it's own issue.

1. Calling `msiConfig.Authorizer()` doesn't ensure you have a functioning authorizer, so you can't rely on `err=nil` to trust you have a good client. This seems like a general API issue across all auth methods.

2. One might then write extra code to call `msiConfig.ServicePrincipalToken()` *first*, ahead of time, to make sure MSI is working, *then* getting the `Authorizer()` afterward. (Or, if the call to SPT fails, go ahead and fall straight through to the trying the CLI authentication).

**First Issue**: This API doesn't make a lot of sense. I waste an entire network call on a refresh token operation that ends up getting thrown away, just to see if the Authorizer is valid.

**Second (more pressing) Issue**: This doesn't even work! The current MSI refresh token behavior means *that this will block application startup for 5 minutes while it's retried, until it gives up, if you execute this code outside of Azure.*

For me to get Sops to support Azure as seamlessly as GCP, I need:
* at least a `IsMSIAvailable()` that returns a true/false quickly so I can re-implement a full, working `GetAuthorizer` that tries MSI and then moves-on to CLI
* much more preferably, #541 is addressed by:
  * fixing the `Authorizer()` interface and implementations to ensure they return valid token sources
  * MSI's RefreshToken doesn't hang for 5 minutes outside Azure
  * `GetAuthorizer` tries conventional auth mechanisms based on the environment, as it does currently, then falls to MSI and then falls to CLI. Then everyone can just use this function everywhere and be done with this.

This blocks: https://github.com/mozilla/sops/pull/710

jhendrixMSFT:
Agreed that the `Authorizer()` pattern here stinks.  Unfortunately it would be a pretty build overhaul to fix this in a non-breaking way.  FWIW this is much better in our track 2 `azidentity` package.

I think the best we can do here is fail early, avoiding retries in the case where MSI authentication isn't available.  I can also make a `MSIAvailable()` func in the `adal` package available if you want to build on top of that.

colemickens:
Makes sense. I'm definitely intrigued about `azidentity`! 

>I think the best we can do here is fail early, avoiding retries in the case where MSI authentication isn't available. I can also make a MSIAvailable() func in the adal package available if you want to build on top of that.

[...]

PR #548, opened 2020-08-13
Fail-fast if the MSI endpoint is not available

jhendrixMSFT:
Added adal.MSIAvailable to probe if the MSI endpoint is available.
Return a TokenRefreshError when attempting to authenticate via MSI when
the endpoint isn't available, this will prevent retries from happening.

### DIFF
--- a/autorest/adal/token_test.go
+++ b/autorest/adal/token_test.go
@@ -237,16 +237,24 @@ func TestServicePrincipalTokenFromMSIRefreshZeroRetry(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to get MSI SPT: %v", err)
 	}
-	spt.MaxMSIRefreshAttempts = 0
+	spt.MaxMSIRefreshAttempts = 1
 
 	body := mocks.NewBody(newTokenJSON("12345", "test"))
 	resp := mocks.NewResponseWithBodyAndStatus(body, http.StatusOK, "OK")
 
 	c := mocks.NewSender()
+	// the call to MSIAvailable() means the sender will be invoked twice.
+	reqCount := 0
 	s := DecorateSender(c,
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/542

autorest/adal/token_test.go
```diff
@@ -1083,6 +1083,12 @@ func TestNewMultiTenantServicePrincipalToken(t *testing.T) {
 	}
 }
 
+func TestMSIAvailable(t *testing.T) {
+	if MSIAvailable(context.Background(), http.DefaultClient) {
```
jhendrixMSFT, line 1087: Can we predicate the expected result based on some other well-known factor like an environment var?

catalinaperalta: To see if you're in a MSI env, you can check if the MSI_ENDPOINT is set and, if not, you can check the IMDS endpoint availability. There are other variables that come into play to see which particular MSI env we're running in, but those two are the main ways to check for a MSI env.

jhendrixMSFT: Let me clarify.  MSIAvailable() already checks the if the IMDS endpoint is available.  Running this test on your dev box, MSIAvailable() will return false, while running it in CI (Azure DevOps VMs) it returns true.

Given that this test is somewhat non-deterministic, maybe it's not a great candidate to run in CI.  I know we hav similar functionality in track 2, how do we test it there?

jhendrixMSFT: We just need to mock the responses here, unclear why I overlooked that :p.

jhendrixMSFT:
Test fix depends on https://github.com/Azure/go-autorest/pull/550

Issue #555, opened 2020-09-02
Fallback for http.NewRequestWithContext for older Go versions


wrosenuance:
A [recent commit](https://github.com/Azure/go-autorest/commit/cbf46deb11b50a1bc9ac059f430c647493829459) introduced a reference to `http.NewRequestWithContext`, which was added in [Go 1.13](https://golang.org/doc/go1.13).

Adding this has meant that the library no longer builds on Go 1.8, which we are trying to support in the [Go MS SQL driver](https://github.com/denisenkom/go-mssqldb), and I am trying to have Azure AD authentication merged there using autorest/adal.

Is it possible to replace this with the previous API of `http.NewRequest` combined with the `WithContext` method? Or to introduce a conditional build option that does so for older Go versions?

jhendrixMSFT:
This has been fixed in autorest/adal/v0.9.2

jhendrixMSFT:
There was a bug found, please use v0.9.4 instead.

PR #556, opened 2020-09-04
Add support for MSIAvailable on older versions of Go

jhendrixMSFT:
http.NewRequestWithContext() was added in Go 1.13.  This change uses
build constraints to support older versions.
Note that Go 1.7 is the minimum version due to reliance on the context
package.

### DIFF
--- a/autorest/adal/token_1.13.go
+++ b/autorest/adal/token_1.13.go
@@ -1,0 +1,38 @@
+// +build go1.13
+
+// Copyright 2017 Microsoft Corporation
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package adal
+
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+// MSIAvailable returns true if the MSI endpoint is available for authentication.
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/555

autorest/adal/token_1.13.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+// MSIAvailable returns true if the MSI endpoint is available for authentication.
+func MSIAvailable(ctx context.Context, sender Sender) bool {
+	// this cannot fail, the return sig is due to legacy reasons
+	msiEndpoint, _ := GetMSIVMEndpoint()
```
catalinaperalta, line 28: Should the error be checked here to "fail" (return false) earlier if the endpoint isnt available?

autorest/adal/token_legacy.go
```diff
@@ -0,0 +1,32 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package adal
+
+// MSIAvailable returns true if the MSI endpoint is available for authentication.
+func MSIAvailable(ctx context.Context, sender Sender) bool {
+	// this cannot fail, the return sig is due to legacy reasons
+	msiEndpoint, _ := GetMSIVMEndpoint()
```
catalinaperalta, line 22: Same question as above

autorest/adal/token_1.13.go
jhendrixMSFT: `GetMSIVMEndpoint()` just returns the endpoint itself (i.e. the URL), it doesn't probe it.

catalinaperalta: Ah I see, ok then 👍

catalinaperalta (approved):
LGTM

wrosenuance:
Thank you! This is really helpful!

PR #558, opened 2020-09-04
fix: return error for failed MSI endpoint check

alexeldeib:
[This commit](https://github.com/Azure/go-autorest/commit/cbf46deb11b50a1bc9ac059f430c647493829459) allows MSI token acquisition to fail fast. However this flow [swallows the real error](https://github.com/Azure/go-autorest/commit/cbf46deb11b50a1bc9ac059f430c647493829459#diff-104007a705afedbaf9fd03870136bd96R946) unlike the other code paths. This PR bubbles the error up for users.

As `MSIAvailable` is publicly exported, I've avoided changing the external signature and relied on the existing tests. Happy to add some unit tests if you like.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -947,6 +947,7 @@ func (spt *ServicePrincipalToken) refreshInternal(ctx context.Context, resource 
 			// return a TokenRefreshError here so that we don't keep retrying
 			return newTokenRefreshError(fmt.Sprintf("the MSI endpoint is not available. Failed HTTP request to MSI endpoint: %v", err), nil)
 		}
+		resp.Body.Close()
 	}
 	if isIMDS(spt.inner.OauthConfig.TokenEndpoint) {
 		resp, err = retryForIMDS(spt.sender, req, spt.MaxMSIRefreshAttempts)
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -941,9 +941,12 @@ func (spt *ServicePrincipalToken) refreshInternal(ctx context.Context, resource 
 	}
 
 	var resp *http.Response
-	if isMSIEndpoint(spt.inner.OauthConfig.TokenEndpoint) && !MSIAvailable(ctx, spt.sender) {
-		// return a TokenRefreshError here so that we don't keep retrying
-		return newTokenRefreshError("the MSI endpoint is not available", nil)
+	if isMSIEndpoint(spt.inner.OauthConfig.TokenEndpoint) {
+		resp, err = getMSIEndpoint(ctx, spt.sender)
+		if err != nil {
+			// return a TokenRefreshError here so that we don't keep retrying
+			return newTokenRefreshError(fmt.Sprintf("the MSI endpoint is not available. Status Code = '%d'. Failed HTTP request to MSI endpoint: %v", resp.StatusCode, err), resp)
+		}
 	}
 	if isIMDS(spt.inner.OauthConfig.TokenEndpoint) {
 		resp, err = retryForIMDS(spt.sender, req, spt.MaxMSIRefreshAttempts)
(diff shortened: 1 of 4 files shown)


alexeldeib:
cc @marwanad

autorest/adal/token.go
```diff
@@ -941,9 +941,12 @@ func (spt *ServicePrincipalToken) refreshInternal(ctx context.Context, resource
...
 
 	var resp *http.Response
-	if isMSIEndpoint(spt.inner.OauthConfig.TokenEndpoint) && !MSIAvailable(ctx, spt.sender) {
-		// return a TokenRefreshError here so that we don't keep retrying
-		return newTokenRefreshError("the MSI endpoint is not available", nil)
+	if isMSIEndpoint(spt.inner.OauthConfig.TokenEndpoint) {
+		resp, err = getMSIEndpoint(ctx, spt.sender)
+		if err != nil {
+			// return a TokenRefreshError here so that we don't keep retrying
+			return newTokenRefreshError(fmt.Sprintf("the MSI endpoint is not available. Status Code = '%d'. Failed HTTP request to MSI endpoint: %v", resp.StatusCode, err), resp)
```
jhendrixMSFT, line 948: If `err` is not `nil` it's not safe to dereference `resp` here (there very well might be no response).

jhendrixMSFT (changes requested):

autorest/adal/token.go
alexeldeib: better? added a unit test

jhendrixMSFT: it definitely fixes the panic.  does it add anything of value though?  from the docs it looks like there's only a very specific case where `Do()` will return a non-nil response and error
```go
// On error, any Response can be ignored. A non-nil Response with a
// non-nil error only occurs when CheckRedirect fails, and even then
// the returned Response.Body is already closed.
```
I'm fine with leaving the change as-is but wanted to point this out.

Out of curiosity, what prompted this change?

alexeldeib: > Out of curiosity, what prompted this change?

we find our service (AKS) hitting this error path frequently after switching over to heavy use of MSI. I suspect the responses will be 429s, which appear to be failing nil check I'm enriching. I assumed the full response would be attached too in this case, apologies for not being more familiar with the sender semantics. Maybe you could suggest a better behavior?

alexeldeib: I guess I should probably flip this -- we should return the error if not nil, and check the http status code otherwise, and do logic similar to inside `retryForIMDS`. But the intent of this code is to fail fast, so maybe that's not really worthwhile?

As an end user, I'll probably end up wrapping the refresh logic myself with backoff to avoid this shortcircuit logic.

jhendrixMSFT: Per the docs, a non-2xx status code doesn't return an error.  The only time I'd expect `Do()` to return an error is if the underlying transport can't connect to the MSI endpoint.  Ideally this should only happen if the MSI endpoint truly isn't available (either it's not enabled on your VM or you're running on a dev box).  If you're seeing this return `false` when it should return `true` I'd like to fix this if possible.

alexeldeib: (I guess that also means we weren't getting throttled, but had some transient connectivity failures? interesting to know)

alexeldeib: > is if the underlying transport can't connect to the MSI endpoint. Ideally this should only happen if the MSI endpoint truly isn't available

we see on the order of 10k successful calls and single digits of failure over long enough time periods (hours). MSI is definitely there so after perusing the docs I agree it's probably network connectivity failures. It'd be nice to get the errors, though 😃 

let me make a few additional tweaks, perhaps.

alexeldeib: maybe more like that?

jhendrixMSFT: This works, and should help add context to the failure so we can make the implementation more robust.

jhendrixMSFT (approved):

PR #567, opened 2020-09-29
Add Content-Length header to XML preparer

mbfrahry:
The `Content-Length` header is missing from the XML preparer which causes mismatch issues:

```
The MAC signature found in the HTTP request 'encodedauthorizationtoken' is not the same as any computed signature
```

This PR adds that in and we don't see the above error message anymore.


jhendrixMSFT (changes requested):
Can you please fix the affected tests?

mbfrahry:
@jhendrixMSFT the test failures have been fixed

jhendrixMSFT (approved):

Issue #583, opened 2020-11-25
Multi-tenant / Cross-tenant authorization not working


bensussman:
I am attempting to get "Cross Tenant Image Sharing" to work by following this guide: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/share-images-across-tenants I am able to follow the guide verbatim, and spin up a VM successfully in "Tenant 2" using the CLI as described in the final "Create a VM using Azure CLI" step in the documentation.

When I attempt to create a VM using the golang CLI, the only way that I see to be "authorized" as a service principal with TWO separate tenants simultaneously is via the `AuxTenants` field on the `ClientCredentialsConfig` like so:

```
clientCreds := auth.NewClientCredentialsConfig(
	sp.ClientID,
	sp.ClientSecret,
	tenantTwoID,
)
clientCreds.AuxTenants = append(clientCreds.AuxTenants, tenantOneID)
multiTenantAuthorizer, e := clientCreds.Authorizer()
...
multiTenantVMClient := compute.NewVirtualMachinesClient(tenantTwoSubscriptionID)
multiTenantVMClient.Authorizer = multiTenantAuthorizer
multiTenantVMClient.Sender = httpClient
...
vmFuture, err := multiTenantVMClient.CreateOrUpdate(ctx, ....)
```

The behavior I see is:

1. `multiTenantVMClient.CreateOrUpdate` hangs for a very very very long time. (about TEN MINUTES)
2. `multiTenantVMClient.CreateOrUpdate` after hanging errors with:

```
Azure CreateOrUpdate() error: azure.multiTenantSPTAuthorizer#WithAuthorization: Failed to refresh one or more Tokens for request to https://management.azure.com/subscriptions/<tenant 2 subscription>/resourceGroups/<rg-name>/providers/Microsoft.Compute/virtualMachines/vm-12345?api-version=2019-03-01: StatusCode=0 -- Original Error: failed to refresh auxiliary token: adal: Refresh request failed. Status Code = '401'. Response body: {"error":"invalid_client","error_description":"AADSTS7000215: Invalid client secret is provided.\r\nTrace ID: 2f...f00
Correlation ID: 5e....b
Timestamp: 2020-11-25 19:03:50Z","error_codes":[7000215],"timestamp":"2020-11-25 19:03:50Z","trace_id":"2f9...0","correlation_id":"5e...b","error_uri":"https://login.microsoftonline.com/error?code=7000215"}
```

[...]

PR #587, opened 2020-12-03
Walk error chain when checking for TokenRefreshError

jhendrixMSFT:
The refresh methods on MultiTenantServicePrincipalToken didn't properly
wrap the inner error, so as a result the check for TokenRefreshError
would fail leading to retries on a credential that would always fail.
Updated the methods to use %w to wrap the error.
Added Unwrap() method to DetailedError
Updated IsTokenRefreshError() to use errors.As() to walk the error
chain.
Legacy versions for consumers on earlier versions of Go have been
provided.  In these implementations the TokenRefreshError isn't wrapped
to avoid the hang at a loss of error information.

### DIFF
--- a/autorest/adal/token_1.13.go
+++ b/autorest/adal/token_1.13.go
@@ -18,6 +18,7 @@ package adal
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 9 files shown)


jhendrixMSFT:
Fixes the hang as mentioned in https://github.com/Azure/go-autorest/issues/583

autorest/adal/token_1.13.go
```diff
@@ -34,3 +35,43 @@ func getMSIEndpoint(ctx context.Context, sender Sender) (*http.Response, error)
 	req.URL.RawQuery = q.Encode()
 	return sender.Do(req)
 }
+
+// EnsureFreshWithContext will refresh the token if it will expire within the refresh window (as set by
```
catalinaperalta, line 39: Just for my own knowledge..why are these methods repeated in both token_1.13.go and token_legacy.go?

jhendrixMSFT: If you look at the top of the file you'll notice a build tag.

`// +build go1.13`

This file will only compile for Go versions 1.13 and above.

In the legacy file, the build tag is `// +build !go1.13` meaning it will only compile for Go versions before 1.13.

With this combination we provide different implementations for all versions of Go.

jhendrixMSFT: Note that in the 1.13 impl we use the `%w` format specifier which was added in Go 1.13.

catalinaperalta: Ah got it, that's good to know. Thanks for the explanation :)

catalinaperalta (approved):
LGTM

Issue #593, opened 2020-12-08
Function App MSI token failed to parse expires_on date format


btHolert:
When generating a ServicePrincipalTokenFromMSI in Azure Function App and then calling EnsureFresh() this error happens:
`parsing time "12/09/2020 16:12:09 +00:00" as "1/2/2006 15:04:05 PM +00:00": cannot parse "0:00" as "PM"`

Code:
```
msiEndpoint, err := adal.GetMSIEndpoint()
if err != nil {
	return nil, err
}

callback := func(t adal.Token) error {
	return nil
}

token, err = adal.NewServicePrincipalTokenFromMSI(msiEndpoint, resourceId, callback)
if err != nil {
	return nil, err
}

err = token.EnsureFresh()
if err != nil {
	return nil, err
}
```

This error happens at https://github.com/Azure/go-autorest/blob/6afd852870d16ce46bf25dfc4c995cc036314afd/autorest/adal/token.go#L1030

It seems like the expiresOnDateFormat wrong there ?

Go version: 1.15.4
go-autorest/autorest version: v0.11.12 // indirect
go-autorest/autorest/adal version: v0.9.8

jhendrixMSFT:
Fixed in `autorest/adal v0.9.9`

Thanks for the bug report.

PR #594, opened 2020-12-08
Fix parsing of expires_on

jhendrixMSFT:
Handle expires_on time stamps that omit an AM/PM designation.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -81,8 +81,11 @@ const (
 	// secret header used when authenticating against app service MSI endpoint
 	secretHeader = "Secret"
 
-	// the format for expires_on in UTC (applicable to MSI via legacy ASE only)
-	expiresOnDateFormat = "1/2/2006 15:04:05 PM +00:00"
+	// the format for expires_on in UTC with AM/PM
+	expiresOnDateFormatPM = "1/2/2006 15:04:05 PM +00:00"
+
+	// the format for expires_on in UTC without AM/PM
+	expiresOnDateFormat = "1/2/2006 15:04:05 +00:00"
 )
 
 // OAuthTokenProvider is an interface which should be implemented by an access token retriever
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/593

catalinaperalta (approved):
lgtm

autorest/adal/token.go
```diff
@@ -1024,13 +1027,18 @@ func (spt *ServicePrincipalToken) refreshInternal(ctx context.Context, resource
 
 // converts expires_on to the number of seconds
 func parseExpiresOn(s string) (json.Number, error) {
+	// convert the expiration date to the number of seconds from now
+	timeToDuration := func(t time.Time) json.Number {
+		dur := t.Sub(time.Now().UTC())
+		return json.Number(strconv.FormatInt(int64(dur.Round(time.Second).Seconds()), 10))
+	}
```
mdgreenfield, line 1034: @jhendrixMSFT, I'm a little confused why this calculates the difference between `t time.Time` (the passed in and parsed `s string`) and `time.Now().UTC()`. That difference I believe is effectively the `expires_in` value (the TTL of the token). 

However, when the `strconv.ParseInt(s, 10, 64)` logic is successful this `parseExpiresOn` function would be passing through the unix timestamp as far as I can tell (i.e. the actual time when the token expires and not the `expires_in` value).

From the IMDS docs - https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http

> `expires_on` - The timespan when the access token expires. The date is represented as the number of seconds from "1970-01-01T0:0:0Z UTC" (corresponds to the token's `exp` claim).

mdgreenfield: I've moved this question to https://github.com/Azure/go-autorest/issues/638. Apologies for the noise here.

Issue #515, opened 2020-04-16
Add custom unmarshalling for ServiceError.InnerError field


jhendrixMSFT:
NRP sends back InnerError as a string while OData-v4 requires it to be an object.  Update the unmarshaller for ServiceError to handle this discrepancy.

Issue #608, opened 2021-01-14
Error unmarshalling an error message to serviceError2


mikhailshilkov:
While creating an Azure SQL Database, I get an error (because I specified some properties incorrectly) but the library fails with `json: cannot unmarshal array into Go struct field serviceError2.details of type map[string]interface {}` which is not helpful at all.

My call that fails:

```go
err = autorest.Respond(
		resp,
		k.client.ByInspecting(),
		azure.WithErrorUnlessStatusCode(http.StatusOK, http.StatusCreated),
		autorest.ByUnmarshallingJSON(&outputs),
		autorest.ByClosing())
```

I logged the body that I get:

```
{"code":"40645","message":"Servername \"someServer20210108\" cannot be empty or null. It can only be made up 
of lowercase letters \u0027a\u0027-\u0027z\u0027, the numbers 0-9 and the hyphen. The hyphen may not lead or 
trail in the name.","target":null,"details":[{"code":"40645","message":"Servername \"someServer20210108\" 
cannot be empty or null. It can only be made up of lowercase letters \u0027a\u0027-\u0027z\u0027, the 
numbers 0-9 and the hyphen. The hyphen may not lead or trail in the name.","target":null,"severity":"16"}],
"innererror":[]}
```

Does the service behave incorrectly here? Could `go-autorest` fail gracefully and maybe return an error which would contain the raw response body?

jhendrixMSFT:
The problem here is `innererror` being an array.  OData v4 spec says it should be a single object.

I have a fix for this, both to handle discrepancies in the error format as well as including the raw JSON in the error message for an unmarshalling failure.

jhendrixMSFT:
Fixed in `autorest/v0.11.17`

PR #609, opened 2021-01-14
Handle other ServiceError corner-cases

jhendrixMSFT:
Improve unmarshalling support for errors that are not OData v4
compliant.
Include raw JSON body when failing to unmarshall into ServiceError.

### DIFF
--- a/autorest/azure/azure.go
+++ b/autorest/azure/azure.go
@@ -92,54 +92,85 @@ func (se ServiceError) Error() string {
 
 // UnmarshalJSON implements the json.Unmarshaler interface for the ServiceError type.
 func (se *ServiceError) UnmarshalJSON(b []byte) error {
-	// per the OData v4 spec the details field must be an array of JSON objects.
-	// unfortunately not all services adhear to the spec and just return a single
-	// object instead of an array with one object.  so we have to perform some
-	// shenanigans to accommodate both cases.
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/515 https://github.com/Azure/go-autorest/issues/608

autorest/azure/azure.go
```diff
@@ -92,54 +92,85 @@ func (se ServiceError) Error() string {
...
+		if len(c) == 1 {
+			se.InnerError = c[0]
+		} else {
+			// multiple errors, stuff them into the value
+			se.InnerError = map[string]interface{}{
+				"multi": c,
+			}
+		}
+	} else if c, ok := sei.InnerError.(string); ok {
+		se.InnerError = map[string]interface{}{c: ""}
```
catalinaperalta, line 166: Why is the string in the key and not the value?

catalinaperalta (approved):
LGTM. Sad there are so many variations. Only left one comment but I'd say it could be chalked up to preference. I do think a key like "error" would be nicer but everything else looks great.

autorest/azure/azure.go
jhendrixMSFT: Oh good idea on an "error" key, will fix.

Issue #612, opened 2021-01-25
Latest version breaks CloudShell authentication


tombuildsstuff:
The recent changes to support MSI in App Service and Function Apps have broken CloudShell authentication, which also specifies the MSI Environment Variable [used in this function to check if `go-autorest` is being run inside an AppService/Function App](https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L885-L896):

<img width="477" alt="Screenshot 2021-01-25 at 07 23 40" src="https://user-images.githubusercontent.com/666005/105668745-40eda980-5ede-11eb-9436-1d405c837a77.png">

Which means attempting to load the MSI token leads to:

```
azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/f6d69ee2-34d5-4ca8-a143-7a2fc1aeca55/providers?api-version=2016-02-01: StatusCode=400 -- Original Error: adal: Refresh request failed. Status Code = '400'. Response body: {"error":{"code":"bad_request_102","message":"Required metadata header not specified"}} Endpoint http://localhost:50342/oauth2/token?api-version=2017-09-01&resource=https%3A%2F%2Fmanagement.azure.com%2F
```

This happens because the `metadata` header is conditionally set based on whether this is/isn't an ASE:

https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L952-L954

Upstream issue: https://github.com/terraform-providers/terraform-provider-azurerm/issues/10292

jhendrixMSFT:
Our support for cloud shell in `go-autorest` is shaky at best, the fact that it happened to work is purely coincidental.  I will look at getting this fixed.

On a side-note, perhaps as a very first step towards track 2, you can evaluate [`azidentity`](https://github.com/Azure/azure-sdk-for-go/tree/master/sdk/azidentity) for authentication in TF?  It has first-class support for cloud shell and other forms of authentication (see the README for the complete list).  It will work with track 1 via the [`azidext`](https://github.com/jongio/azidext/tree/master/go/azidext) adapter.

jhendrixMSFT:
Fixed in module `github.com/Azure/go-autorest/autorest/adal@v0.9.11`

jhendrixMSFT:
`github.com/Azure/go-autorest/autorest/azure/auth/@v0.5.7` now uses the fixed `adal` module.

PR #613, opened 2021-02-03
Refactor managed identity detection logic

jhendrixMSFT:
This back-ports a portion of the managed identity environment detection logic from `azidentity`.  It does not add support for other managed identity environments that weren't already present in this package.

### DIFF
--- a/autorest/adal/token.go
+++ b/autorest/adal/token.go
@@ -708,12 +708,14 @@ func getMSIType() (msiType, string, error) {
 }
 
 // GetMSIVMEndpoint gets the MSI endpoint on Virtual Machines.
+// NOTE: this always returns the IMDS endpoint, it does not work for app services or cloud shell.
 // Deprecated: NewServicePrincipalTokenFromMSI() and variants will automatically detect the endpoint.
 func GetMSIVMEndpoint() (string, error) {
 	return msiEndpoint, nil
 }
 
-// GetMSIAppServiceEndpoint get the MSI endpoint for App Service and Functions
+// GetMSIAppServiceEndpoint get the MSI endpoint for App Service and Functions.
+// It will return an error when not running in an app service/functions environment.
 // Deprecated: NewServicePrincipalTokenFromMSI() and variants will automatically detect the endpoint.
 func GetMSIAppServiceEndpoint() (string, error) {
 	msiType, endpoint, err := getMSIType()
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/612

autorest/adal/token.go
```diff
@@ -932,9 +932,9 @@ func (spt *ServicePrincipalToken) refreshInternal(ctx context.Context, resource
 	req = req.WithContext(ctx)
 	var resp *http.Response
 	if msiSecret, ok := spt.inner.Secret.(*ServicePrincipalMSISecret); ok {
-		req.Method = http.MethodGet
 		switch msiSecret.msiType {
 		case msiTypeAppServiceV20170901:
+			req.Method = http.MethodGet
 			req.Header.Set("secret", os.Getenv(msiSecretEnv))
 			break
```
catalinaperalta: Does this env not add the clientid or am I missing it somewhere?

jhendrixMSFT: That happens in `newServicePrincipalTokenFromMSI()`.  In that function the endpoint is created.  Then in `refreshInternal()` the HTTP request is created.  So the complete request creation is split over two places :(

catalinaperalta: Ah I see...that threw me off for a second there. So refreshInternal is called even for the first request?

catalinaperalta (approved):
My question isnt related to the MSI changes so the fixes LGTM.

autorest/adal/token.go
jhendrixMSFT: Correct.

autorest/adal/token.go
```diff
@@ -662,118 +664,161 @@ func NewServicePrincipalTokenFromAuthorizationCode(oauthConfig OAuthConfig, clie
...
-		return GetMSIAppServiceEndpoint()
-	}
-	return GetMSIVMEndpoint()
+	_, endpoint, err := getMSIType()
+	return endpoint, err
 }
 
 // NewServicePrincipalTokenFromMSI creates a ServicePrincipalToken via the MSI VM Extension.
 // It will use the system assigned identity when creating the token.
+// NOTE: the msiEndpoint parameter is deprecated and will be ignored.
```
jhendrixMSFT, line 740: @tombuildsstuff I don't see why this would break anything but please let me know if this will negatively impact TF.

tombuildsstuff: iirc there's some cases where some folks use Custom URI's for this (I vaguely recall talk of a custom proxy/implementation being used) - so we expose that field from our side, which'd be a breaking change here

Would it be worth making CloudShell, App Service and MSI as separate authorizers - having this "combined MSI" as some kind of factory for those for backwards compatibility purposes? This'd also allow explicitly picking which ones to/not to support?

jhendrixMSFT: Ok, we can support this param.  My thoughts are pass an empty string to get the default value, and if non-empty we'll use that value instead.  We'll still have to go through our detection logic though to determine if this is IMDS, cloud shell, or app services as the endpoint isn't enough to tell.  Will also add a new authorizer that doesn't have this `msiEndpoint` arg.  Will this work for you?
Also, it would be great to get more info on this scenario so that we can properly support it.  Is this for stack?

autorest/adal/token.go
```diff
@@ -737,32 +739,54 @@ func GetMSIEndpoint() (string, error) {
...
+	// ClientID is the user-assigned identity to use during authentication.
+	// It is mutually exclusive with IdentityResourceID.
+	ClientID string
+
+	// IdentityResourceID is the resource ID of the user-assigned identity to use during authentication.
+	// It is mutually exclusive with ClientID.
+	IdentityResourceID string
+}
+
+//NewServicePrincipalTokenFromManagedIdentity creates a ServicePrincipalToken using a managed identity.
```
catalinaperalta, line 781: Should this comment explain what MSI envs it will check for?

jhendrixMSFT: Good idea, will add.

jhendrixMSFT: Fixed

PR #621, opened 2021-03-08
Changing default polling interval from 60 to 30

madewithsmiles:
Proposing this change following email discussion on making it less conservative. What do you think of this change?

Now it makes me think, can we update the `DefaultRetryDuration` from 30 seconds to 15 seconds if the change above is fine?

---

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -31,7 +31,7 @@ import (
 
 const (
 	// DefaultPollingDelay is a reasonable delay between polling requests.
-	DefaultPollingDelay = 60 * time.Second
+	DefaultPollingDelay = 30 * time.Second
 
 	// DefaultPollingDuration is a reasonable total polling duration.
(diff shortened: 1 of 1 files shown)


autorest/client.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	// DefaultPollingDelay is a reasonable delay between polling requests.
-	DefaultPollingDelay = 60 * time.Second
+	DefaultPollingDelay = 30 * time.Second
```
jhendrixMSFT, line 34: This could end up doubling the number of GET calls during polling which might impact large-scale apps and introduce throttling.  This needs to be investigated before we could merge this as it affects everybody.

jhendrixMSFT:
Closing for now.  If we get data that confirms we can safely make this change please ping back.

jhendrixMSFT (approved):

PR #622, opened 2021-03-09
Brfole/add retry headers

bfoley13:
In an attempt to better understand request failures and monitor for potential networking issues, I've added a retry attempt header to RetriableRequests that addes the retry attempt autorest is currently sending!

Testing succeeds and I was also trying to add verification to sender_tests, but that required editing the mock sender to verify header attempt is equivalent to mock sender attempts and it seems to be its own module.  Would that require separate PR's?

### DIFF
--- a/autorest/azure/rp.go
+++ b/autorest/azure/rp.go
@@ -32,7 +32,7 @@ func DoRetryWithRegistration(client autorest.Client) autorest.SendDecorator {
 		return autorest.SenderFunc(func(r *http.Request) (resp *http.Response, err error) {
 			rr := autorest.NewRetriableRequest(r)
 			for currentAttempt := 0; currentAttempt < client.RetryAttempts; currentAttempt++ {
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
You are correct that the mock server would need to be updated first and tagged so this PR can consume it.

Can you tell me a bit more about this feature, do the RPs consume/log this header?  Is it applicable to all RPs?

bfoley13:
Its's less about the rp change and more about monitoring retries.  I did make changes to the azure/rp.go sender func as well because I thought it would be useful, but those can be reverted!  For our use case, we want to log this header on the client and server side to have more in depth knowledge about client requests, when they make it to the server, why they might fail with retry-able errors, and how many retry attempts have occurred.

Let me know your thoughts!

jhendrixMSFT:
Unfortunately this change would impact all clients which we don't want to do.

PR #625, opened 2021-03-31
[DO NOT MERGE] Cache last Retry-After value when polling on LRO

jhendrixMSFT:
Some operations return a retry-after value on the first response but not
subsequent responses.  This causes the polling logic to fall back to the
default polling value, causing some LROs to take longer than required.
Cache the last retry-after value and use it when no retry-after value is
provided by the service.


jhendrixMSFT:
Still need to test this.

jhendrixMSFT:
CC @tombuildsstuff @katbyte to assess impact on Terraform scenarios.

jhendrixMSFT:
Will change the default polling interval to 30s instead.

PR #635, opened 2021-05-27
Add ossRDBMSResourceId to Environment configuration.

rbranche:

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -45,6 +45,7 @@ type ResourceIdentifier struct {
 	Datalake            string `json:"datalake"`
 	Batch               string `json:"batch"`
 	OperationalInsights string `json:"operationalInsights"`
+	OSSRDBMS            string `json:"ossRDBMS"`
 	Storage             string `json:"storage"`
 	Synapse             string `json:"synapse"`
 	ServiceBus          string `json:"serviceBus"`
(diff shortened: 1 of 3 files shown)


rbranche:
@jhendrixMSFT This is similar to the change I just made for the DNS Suffixes (#631). This is another area where the identifier is different between the different clouds so the environment should include the ResourceId.

autorest/azure/environments.go
```diff
@@ -116,6 +117,7 @@ var (
 			Datalake:            "https://datalake.azure.net/",
 			Batch:               "https://batch.core.windows.net/",
 			OperationalInsights: "https://api.loganalytics.io",
+			OSSRDBMS:            "https://ossrdbms-aad.database.windows.net",
```
matthchr, line 120: FWIW, I am not sure `OSSRDBMS` is the right terminology here. This is really the active directory resource(?) URL to provide to AAD when requesting a token with access to MySQL. (Note that I am not 100% sure the terminology I used is correct either...)

See [connecting to azure database for mysql using azure AD](https://docs.microsoft.com/en-gb/azure/mysql/howto-configure-sign-in-azure-ad-authentication#connecting-to-azure-database-for-mysql-using-azure-ad).

It might be better to just say `MySQL` here?

rbranche: This ResourceID is also used for Postgres, Azure SQL, etc.

e.g. https://docs.microsoft.com/en-us/azure/postgresql/howto-connect-with-managed-identity#retrieving-the-access-token-from-azure-instance-metadata-service

The Azure CLI also uses this terminology:

```
lekonia:azure-service-operator.git{cloud-centric-environment-config} branche$ az cloud show
{
  "endpoints": {
...
    "ossrdbmsResourceId": "https://ossrdbms-aad.database.windows.net",
...
```

matthchr: This sounds good to me then! Thanks for clarifying.

jhendrixMSFT (approved):

PR #637, opened 2021-06-04
Create go.yml

smooth80:
- name: Setup Go environment
  uses: actions/setup-go@v2.1.3
  with:
    # The Go version to download (if necessary) and use. Supports semver spec and ranges.
    go-version: # optional
    # Whether to download only stable versions
    stable: # optional, default is true
    # Used to pull node distributions from go-versions.  Since there's a default, this is typically not supplied by the user.
    token: # optional, default is ${{ github.token }}

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,25 @@
+name: Go
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Hello!  Thanks for the PR.  We use Azure DevOps for our CI in this repo.

https://github.com/Azure/go-autorest/blob/master/azure-pipelines.yml

Is the current CI missing something that a Github flow provides?

jhendrixMSFT:
Closing.  Please ping if you have concerns.

PR #641, opened 2021-07-27
Print error to info log before backing off

Nirusu:

-----------------------------------------------------------------

This is a small change I made that should help users of the Azure SDK debugging networking issues whenever DelayForBackoff(WithCap) is called. Normally you will have to wait until the cap is reached to actually see the error. With this change, it should be printed to the info log (assuming `AZURE_GO_SDK_LOG_LEVEL` is set at least to `Info`) before DelayForBackoff(WithCap) is called, so that potential errors can be detected quicker.

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -272,7 +272,7 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
 				if err == nil {
 					return resp, err
 				}
-				logger.Instance.Writef(logger.LogInfo, "DoRetryForAttempts: received error for attempt %d: %v\n", attempt+1, err)
+				logger.Instance.Writef(logger.LogError, "DoRetryForAttempts: received error for attempt %d: %v\n", attempt+1, err)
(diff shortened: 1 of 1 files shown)


autorest/sender.go
```diff
@@ -272,6 +272,7 @@ func DoRetryForAttempts(attempts int, backoff time.Duration) SendDecorator {
 				if err == nil {
 					return resp, err
 				}
+				logger.Instance.Writef(logger.LogInfo, "DoRetryForAttempts: received error for attempt %d: %v\n", attempt+1, err)
```
jhendrixMSFT, line 275: These should be `logger.LogError`

jhendrixMSFT (changes requested):
Sorry for the delay.

autorest/sender.go
Nirusu: Done

jhendrixMSFT (approved):

Nirusu:
Sorry for the force-pushes, actually changed the "sleeping for" message by accident too; removed that change.

Yup, no problem, signed the CLA.

Issue #642, opened 2021-07-29
Replace github.com/form3tech-oss/jwt-go with https://github.com/golang-jwt/jwt?


brackendawson:
The maintained fork of https://github.com/dgrijalva/jwt-go seems now to be https://github.com/golang-jwt/jwt, can [go-autorest/autorest/adal](https://github.com/Azure/go-autorest/blob/76c7860e9b602854c8598676b7550c715c48a607/autorest/adal/go.mod#L11) move to this fork now?

This request stems from distribution/distribution#3361, we import this repository via https://github.com/Azure/azure-sdk-for-go and we would like to settle on the most likely to be maintained fork.

dims:
+1 from me. I see the same request in https://github.com/Azure/go-autorest/pull/570#issuecomment-857445942 as well

brackendawson:
So a good reason to do this is that v3.2.4 of form3tech-oss/jwt-go that contains a vulnerability fix is not importable, it was given an invalid go.mod file which should have had a version in its module line.

thaJeztah:
Quick attempt at unifying them again; https://github.com/golang-jwt/jwt/pull/40

PR #645, opened 2021-08-13
autorest/adal: update modules, and switch to github.com/golang-jwt/jwt v4.0.0

thaJeztah:
fixes https://github.com/Azure/go-autorest/issues/642
fixes https://github.com/Azure/go-autorest/issues/648

- autorest: update go.mod files to latest patch releases
- autorest/adal: switch to github.com/golang-jwt/jwt v3.2.2

The github.com/form3tech-oss/jwt-go fork has been archived, and all
development is moved to the github.com/golang-jwt/jwt project, which
is now the canonical/official fork of github.com/dgrijalva/jwt-go.

The latest version of github.com/golang-jwt/jwt (v4.0.0) also adds
go module support, but the main github.com/Azure/go-autorest package
currently uses "go dep", which is not compatible with go modules that
use versioned import paths (github.com/golang-jwt/jwt/v4) for modules
that do not use a `v<version>` directory.


thaJeztah:
- Relates to https://github.com/distribution/distribution/pull/3459
- Relates to https://github.com/form3tech-oss/jwt-go/pull/17#issuecomment-895259465
- Relates to https://github.com/Azure/go-autorest/pull/570

@brackendawson @jhendrixMSFT  PTAL (hope I did the `go dep` changes correctly)

thaJeztah:
oh! this probably requires Go to be updated; see https://github.com/distribution/distribution/pull/3472 https://github.com/Azure/go-autorest/blob/778ae565bde392f58e04c3c8ff07c2ac506d00c9/azure-pipelines.yml#L10-L16

thaJeztah:
marking as draft until https://github.com/Azure/go-autorest/pull/646 is merged (included in this PR to validate in CI, but I'll rebase once https://github.com/Azure/go-autorest/pull/646 is merged)

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/648

autorest/adal/go.mod
```diff
@@ -8,6 +8,9 @@ require (
 	github.com/Azure/go-autorest/autorest/mocks v0.4.1
 	github.com/Azure/go-autorest/logger v0.2.1
 	github.com/Azure/go-autorest/tracing v0.6.0
-	github.com/form3tech-oss/jwt-go v3.2.2+incompatible
+	// NOTE: cannot move to github.com/golang-jwt/jwt/v4 yet, because the main
+	//       github.com/Azure/go-autorest code uses go dep, which is not
+	//       compatible with v2+ go modules that do not use a v<version> directory.
+	github.com/golang-jwt/jwt v3.2.2+incompatible
```
jhendrixMSFT, line 14: I think we can rip the band-aid off now and just delete the gopkg.* files, then update to v4.

thaJeztah: That'd be great! I wasn't sure if someone / some team internal at Microsoft was depending on this, which is why I tried to keep things as-is.

I can give it a quick try later to create a suitable `go.mod` (and 🤞 hoping `go mod` is cooperating 😂)

thaJeztah: Oh, hmm... didn't fully realise the "main" module is empty. I wonder if it's better to _add_ a module or to just remove the GoDep files.

jhendrixMSFT: It should be ok to remove them.  The only reason I left them was to keep back-compat during the transition to modules.  Now that we're several releases in and the world has moved on from `dep` I think it's ok to remove them.

thaJeztah: Adding a `go.mod` would also mean it has to be renamed to `github.com/Azure/go-autorest/v15`, so perhaps we should just remove GoDep, and consider the repository to be a "collection of modules" without a "main" module 🤔

jhendrixMSFT: Yes exactly, and is the reason I didn't add one during the initial migration (although, if I had to do it over again, I'd make `/autorest` one module instead of several).

azure-pipelines.yml
```diff
@@ -51,25 +49,32 @@ jobs:
         displayName: 'Install Dependencies'
 
       - script: |
-          go vet ./autorest/...
-          go vet ./logger/...
-          go vet ./tracing/...
+          modules=$(find . -name go.mod)
+          for module in $modules; do
+            pushd "$(dirname $module)" && go vet ./... && popd;
+          done
```
thaJeztah, line 55: Each module is now it's own "thing", so looks like we need to iterate over all modules and run vet, lint, test separately. `gofmt` seems to be the exception (it didn't like iterating 🤔)

azure-pipelines.yml
```diff
@@ -100,11 +106,11 @@ jobs:
...
-          testResultsFiles: $(sdkPath)/report.xml
+          testResultsFiles: $(sdkPath)/**/report.xml
           failTaskOnFailedTests: true
 
       - task: PublishCodeCoverageResults@1
         inputs:
           codeCoverageTool: Cobertura 
           summaryFileLocation: $(sdkPath)/coverage.xml
-          additionalCodeCoverageFiles: $(sdkPath)/coverage.html
+          additionalCodeCoverageFiles: $(sdkPath)/**/coverage.html
```
thaJeztah, line 116: Not sure if this one will work as expected; each module will now produce an `xml` and a `html`, but this is only uploading the `html`, and a `coverage.xml` of the "main" module (which doesn't exist, so nothing will be stored there)

Perhaps you're able to check if the uploaded results produce something useful now.

azure-pipelines.yml
```diff
@@ -40,8 +40,6 @@ jobs:
 
       - script: |
           set -e
-          curl -sSL https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-          dep ensure -v
```
thaJeztah, line 44: This is no longer needed; go will automatically download modules for each submodule during `test` / `vet`, etc.

azure-pipelines.yml
jhendrixMSFT: I made a slight change to merge all the coverage.txt files together (we do the same thing in azure-sdk-for-go).

thaJeztah: Ah thanks! It was getting late here, and I didn't dig yet to see if that was possible, so thanks for that!

thaJeztah: arf 😢 looks like it's not happy with that change;

```
error: can't find "config.go": cannot find module providing package github.com/Azure/go-autorest/autorest/adal: working directory is not part of a module
##[error]Bash exited with code '1'.
Finishing: Run Tests
```

thaJeztah: I guess because the root `/` is not a module, so those tools won't work outside of a module 😞 

Perhaps doing a merge within the loop could work (using `$(sdkPath)` for the target directory to write to

jhendrixMSFT: Yes, apparently `gocov` is required to be run from the module directory.

thaJeztah: Let me give that a quick try

jhendrixMSFT: I wonder if it can be _any_ module directory or if it will have to be per each, which then defeats joining them together.

jhendrixMSFT: Hmm no code coverage info now.  I'm gonna try something, one sec.

thaJeztah: Ah, yes, I'm stupid. It just pipes the output, and doesn't take the input of "what's already there". Let me fix (hoping it will accept that).

azure-pipelines.yml
```diff
@@ -40,34 +40,42 @@ jobs:
...
-          go test -race -v -coverprofile=coverage.txt -covermode atomic ./autorest/... ./logger/... ./tracing/... 2>&1 | go-junit-report > report.xml
-          gocov convert coverage.txt > coverage.json
+          modules=$(find . -name go.mod)
+          for module in $modules; do
+            pushd "$(dirname $module)";
+              go test -race -v -coverprofile=coverage.txt -covermode atomic ./... 2>&1 | go-junit-report > report.xml;
+              gocovmerge "$(sdkPath)/mergedCoverage.txt" coverage.txt > "$(sdkPath)/mergedCoverage.txt"
+            popd;
+          done
+          gocov convert mergedCoverage.txt > coverage.json
```
jhendrixMSFT, line 78: This is the command that fails when not run from a (any?) module directory.

azure-pipelines.yml
thaJeztah: If that doesn't work, I'm wondering if `gocovmerge` works with `GO111MODULE=off` (or `auto`); were go modules required to make it work?

Because if it can work without, we could take your approach, but then keep go modules disabled for that step 🤔

jhendrixMSFT: My idea was to merge after the loop, then `cd autorest` and run `gocov` from there.

thaJeztah: So, the good new is "CI is still green" with the current approach, but I don't think I have access to the published results, so I'm not sure the merge worked.

Feel free to amend my last commit to try your approach if it didn't work. I'm curious why it needs a module though; I hope running it from within a non-related module won't cause issues 🤔

jhendrixMSFT (approved):

jhendrixMSFT:
Thanks for your help on this.

thaJeztah:
You're welcome! Thanks for review and merge!

PR #646, opened 2021-08-13
update go versions (go1.15, go1.16, go1.17)

thaJeztah:
see https://github.com/Azure/go-autorest/pull/645#issuecomment-898440791

### DIFF
--- a/autorest/adal/go.mod
+++ b/autorest/adal/go.mod
@@ -1,6 +1,6 @@
 module github.com/Azure/go-autorest/autorest/adal
 
-go 1.12
+go 1.15
 
 require (
 	github.com/Azure/go-autorest v14.2.0+incompatible
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,5 @@
 variables:
+  GO111MODULE: 'auto'
   GOPATH: '$(system.defaultWorkingDirectory)/work'
   sdkPath: '$(GOPATH)/src/github.com/$(build.repository.name)'
 
@@ -8,12 +9,16 @@ jobs:
 
     strategy:
       matrix:
-        Linux_Go113:
+        Linux_Go115:
           vm.image: 'ubuntu-18.04'
-          go.version: '1.13'
-        Linux_Go114:
+          go.version: '1.15'
+        Linux_Go116:
           vm.image: 'ubuntu-18.04'
-          go.version: '1.14'
+          go.version: '1.16'
+        Linux_Go117:
+          vm.image: 'ubuntu-18.04'
+          go.version: '1.17'
+          current_go: 'true'
 
     pool:
       vmImage: '$(vm.image)'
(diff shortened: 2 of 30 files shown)


thaJeztah:
Hmm... probably outdated linter;

```
/home/vsts/work/_temp/dc578aa8-fc5c-4e34-bd4c-3fb0a4f9ff39.sh: line 1: golint: command not found
/home/vsts/work/_temp/dc578aa8-fc5c-4e34-bd4c-3fb0a4f9ff39.sh: line 2: golint: command not found
```

thaJeztah:
@jhendrixMSFT PTAL

autorest/adal/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/go-autorest/autorest/adal
 
-go 1.12
+go 1.15
```
jhendrixMSFT, line 3: This sets the minimum version to 1.15.  While we only officially support Go versions N and N-1, we try to keep things working on older versions where possible.

jhendrixMSFT (changes requested):
The CI changes are good but we can't take the `go.mod` changes.

autorest/adal/go.mod
thaJeztah: Unfortunately, golang-jwt no longer with versions before 1.15; https://github.com/Azure/go-autorest/pull/645#issuecomment-898440791

thaJeztah: Note that (currently) the version in `go.mod` is only used as a "recommended' version; it's not (yet?) enforcing users to use this version, so depending on what code they consume, it may still work.

thaJeztah: @jhendrixMSFT ptal 🤗

azure-pipelines.yml
```diff
@@ -8,12 +9,12 @@ jobs:
 
     strategy:
       matrix:
-        Linux_Go113:
+        Linux_Go115:
           vm.image: 'ubuntu-18.04'
-          go.version: '1.13'
-        Linux_Go114:
+          go.version: '1.15'
```
jhendrixMSFT, line 14: Now that 1.17 has shipped, can we change these versions to 1.16.7 and 1.17 respectively?

thaJeztah: Yes, that should be possible; do you want to fully "drop" 1.15, or keep 1.15 running in CI as well for some time as minimum version (while people upgrade)? Looks like CI is fairly short, and they run in parallel, so probably should be ok.

jhendrixMSFT: I'm ok to drop it given that Google only supports the latest two versions.  However, CI is short as you pointed out, so if you have a compelling reason to keep it I'm fine with that.

jhendrixMSFT: We can leave the `go.mod` changes at `1.15` though.

thaJeztah: If `go.mod` sticks at 1.15, it's probably good to also test on that version (just incase). Go 1.17 is pretty "fresh", and some project may linger a bit. Let me add 1.17 to the matrix 👍

azure-pipelines.yml
```diff
@@ -81,7 +86,7 @@ jobs:
         workingDirectory: '$(sdkPath)'
         displayName: 'Format Check'
         failOnStderr: true
-        condition: succeededOrFailed()
+        condition: and(succeededOrFailed(), eq(variables['current_go'], 'true'))
```
thaJeztah, line 89: I had to update this condition, because go1.17 has a new `gofmt` output (adding `//go:build` comments, instead of the "old" `// +build` ones).

thaJeztah:
Ah, cool, my `condition:` worked (thanks StackOverflow 😂), and it's green again 👍

jhendrixMSFT (approved):

jhendrixMSFT:
I will work on releasing the updated modules a bit later today.

thaJeztah:
Let me rebase my other PR now that this one's merged

Issue #652, opened 2021-09-03
GetTokenFromCLI returns an empty error message, if invoke cli invoke failed


StiviiK:
`GetTokenFromCLI` returns an empty error message if the cli could not be invoked for some reason, as the returned error only contains information from `stderr` if the command fails, but none if the command could not be even executed in the first place.

jhendrixMSFT:
Fixed in `autorest/azure/cli/v0.4.4`

PR #653, opened 2021-09-03
GetTokenFromCLIWithParams fixed empty error message if command invoke failed.

StiviiK:

Fixes the Issue: #652

### DIFF
--- a/autorest/azure/cli/token.go
+++ b/autorest/azure/cli/token.go
@@ -165,12 +165,13 @@ func GetTokenFromCLIWithParams(params GetAccessTokenParams) (*Token, error) {
 		cliCmd.Args = append(cliCmd.Args, "--tenant", params.Tenant)
 	}
 
-	var stderr bytes.Buffer
-	cliCmd.Stderr = &stderr
-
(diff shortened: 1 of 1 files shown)


autorest/azure/cli/token.go
```diff
@@ -170,7 +170,11 @@ func GetTokenFromCLIWithParams(params GetAccessTokenParams) (*Token, error) {
 
 	output, err := cliCmd.Output()
 	if err != nil {
-		return nil, fmt.Errorf("Invoking Azure CLI failed with the following error: %s", stderr.String())
+		if _, ok := err.(*exec.ExitError); ok {
```
jhendrixMSFT, line 173: Better to use `errors.As()` as that's the preferred pattern.

Another option, maybe it's simplest to check if `stderr.String()` for the empty string?

StiviiK: Yeah, you are right a check if stderr is empty is the simplest. Updated the PR accordingly.

jhendrixMSFT (approved):

Issue #629, opened 2021-05-08
panic: interface conversion: http.RoundTripper not *http.Transport


uvw:
The library panics at https://github.com/Azure/go-autorest/blob/master/autorest/adal/sender.go#L76 when the caller had `http.DefaultTransport` replaced with a custom RoundTripper, which is quite common for global tracing/debugging purposes. 

Usually, it is not expected from a library to panic unless explicitly documented (e.g., `Must*` functions). Would it be possible to handle this case gracefully? At least preventing the panic, for instance, fall back to new `http.Transport` and emit a warning if there is a type mistmatch.

jhendrixMSFT:
Fixed in `adal/v0.9.16`

PR #654, opened 2021-09-09
Create a default transport

jhendrixMSFT:
Type-asserting the http.DefaultTransport can fail if it was replaced
with a custom RoundTripper.  Instead, create a new http.Transport with
the settings copied from http.DefaultTransport.

### DIFF
--- a/autorest/adal/sender.go
+++ b/autorest/adal/sender.go
@@ -16,9 +16,11 @@ package adal
 
 import (
 	"crypto/tls"
+	"net"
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/629

jhendrixMSFT:
We'll need to forward-port this to `azcore`.

seankane-msft (approved):
LFTM

PR #662, opened 2021-11-15
Don't raise an error from Client.AddToUserAgent

aaron-trout:
This is essentially just string concatenation, there is no real reason
to return an error here. This causes clients to write dozens of lines
of pointless error handling or just not handle the err.

Thinking about the contract of this function, if I've asked for it to
append "" (empty string) to the user agent, and it does nothing, then
it has been a success. Raising an error in that case is a bit weird.

`"foo" + ""` does not result in an error, which is essentially what
this is doing.

---

 - [x] ~I've added Apache 2.0 Headers to the top of any new source files.~ Not applicable

### DIFF
--- a/autorest/client.go
+++ b/autorest/client.go
@@ -222,12 +222,10 @@ func newClient(ua string, renegotiation tls.RenegotiationSupport) Client {
 }
 
 // AddToUserAgent adds an extension to the current user agent
-func (c *Client) AddToUserAgent(extension string) error {
+func (c *Client) AddToUserAgent(extension string) {
 	if extension != "" {
 		c.UserAgent = fmt.Sprintf("%s %s", c.UserAgent, extension)
-		return nil
 	}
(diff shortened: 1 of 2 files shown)


jhendrixMSFT (changes requested):
While I agree that the error is useless, this would be a breaking change and the public surface area at this point is frozen.  If you want to update the doc comment explaining that this error can be ignored that would be acceptable.

aaron-trout:
@jhendrixMSFT thanks for the reply. Do you have any versioning strategy for this repo which could accommodate this change? i.e. could this be earmarked for a future "v2.0" release (or rather, v1 looking at the current version numbers).

In the mean time I will go ahead and update the docs regarding the error.

aaron-trout:
Oh interesting, so we should be updating our code to move to that new set of SDK packages I guess. Thanks for the info!

jhendrixMSFT:
Closing as stale.  Please reopen if still needed.

Issue #671, opened 2021-12-22
Add support for federated authentication


chrisst:
The recently added federated identity workflow will allow users to pass in an externally generated JWT to exchange for a Azure oauth token. Ref: https://docs.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation

Can support for this workflow be added so that an service principal token can be constructed from a JWT ?

jhendrixMSFT:
Fixed in `autorest/adal@v0.9.19`

PR #680, opened 2022-01-26
Added federated OIDC JWT support.

JustinJudd:
In response to issue #671, I took a pass at adding in capability of using OIDC JWTs. Completed the CLA.

> Thank you for your contribution to Go-AutoRest! We will triage and review it as soon as we can.
>
> As part of submitting, please make sure you can make the following assertions:
>  - [x] I've tested my changes, adding unit tests if applicable.
>  - [ ] *N/A - No new files* I've added Apache 2.0 Headers to the top of any new source files.
>


JustinJudd:
Friendly ping if a reviewer or maintainer can take a look at this.

autorest/adal/token.go
```diff
@@ -365,6 +365,25 @@ func (secret ServicePrincipalAuthorizationCodeSecret) MarshalJSON() ([]byte, err
 	})
 }
 
+// ServicePrincipalFederatedSecret implements ServicePrincipalSecret for Federated JWTs.
+type ServicePrincipalFederatedSecret struct {
+	Jwt string
```
jhendrixMSFT, line 370: I don't think this needs to be exported.

chlowell: Maybe this is intended as a (goroutine unsafe) way to replace JWTs as they expire? Does `adal` have a different pattern for that sort of thing?

jhendrixMSFT: Good point.  There's no established pattern for this in `adal` today, so if the `Jwt` needs to be updated, then you'd have to export it like at present (it's racy like you said).  Also, if this is the case, then the field should be named `JWT` for consistency.

@chlowell do we need to add support for this in `azidentity`?

jhendrixMSFT: Or perhaps our current abstractions make it possible for customers to write their own implementation.

chlowell: For `azidentity` this is part of https://github.com/Azure/azure-sdk-for-go/issues/15615. Planning to take a callback for updating the assertion.

jhendrixMSFT (changes requested):
Need clarification in the usage of the `Jwt` field.

JustinJudd:
Yeah, I don't see any strong reason that the JWT would need to be exported in this instance. I updated the PR to make it internal.

jhendrixMSFT (approved):

autorest/adal/token.go
chlowell: Do we want applications to create a new `ServicePrincipalFederatedSecret` after the JWT expires?

karlschriek:
Anything still blocking this? I opened up two downstream issues where I would love to implement workload identities as soon as this is merged.

jhendrixMSFT:
There's an open question about how to handle the JWT expiring.  @karlschriek how is your app handling this?

karlschriek:
We use this for "Azure Workload Identities" on Kubernetes. Essentially we annotate a Pod's ServiceAccount, and that leads to the following vars being injected into the Pod, as follows (note the token expiry annotation).

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: external-dns
  namespace: external-dns
  labels:
    azure.workload.identity/use: 'true'
  annotations:
    azure.workload.identity/client-id: bbbbbbbb-xxx-yyyyy-mmmm-rrrrrrrrrr
    azure.workload.identity/service-account-token-expiration: '86400' # expiry after one day
    azure.workload.identity/tenant-id: yyyyyyy-xxx-zzzz-xxxx-rrrrrrrrrrrrr
```

Starting a Pod with the above service account results in the following ENV vars being mounted.

```
AZURE_AUTHORITY_HOST: https://login.microsoftonline.com/
AZURE_CLIENT_ID: bbbbbbbb-xxx-yyyyy-mmmm-rrrrrrrrrr
AZURE_FEDERATED_TOKEN_FILE: /var/run/secrets/azure/tokens/azure-identity-token
AZURE_TENANT_ID: yyyyyyy-xxx-zzzz-xxxx-rrrrrrrrrrrrr
```

In addition, a JWT token is mounted at the file location `/var/run/secrets/azure/tokens/azure-identity-token`.

My understanding is that after `86400` seconds the token at that location will be refreshed, i.e. the contents of the file is replaced (I cannot 100% confirm this, but that is what I assume happens). I noticed that the `adal` module has token refresh callbacks, though it isn't clear to me how they work. Am I to take from the comments above that these cannot be used to automatically refresh the token? I our case "refreshing" the token in-memory would simply mean reading it from the same location on disk again.

karlschriek:
By far the most convenient on client side would be to be able to pass for example `adal.NewServicePrincipalTokenFromFederatedToken(*oauthConfig, awiClientId, jwtDateTime, expirySeconds, env.ResourceManagerEndpoint)` instead have adal just look up the token again after `expirySeconds` has passed. Is there any feasible way to imlement this?

JustinJudd:
Since the JWT has it's own expiry, there isn't a way to get refresh tokens using the JWT. Rather we are wrapping around this call to get a new JWT and then a new Azure access token when needed. Getting the JWT will be highly dependent on the Identity Provider in use.

Issue #684, opened 2022-02-22
"MSI not available" for app services


scott-the-programmer:
# Issue

Currently the pre-auth check introduced in https://github.com/Azure/go-autorest/pull/670 only checks against http://169.254.169.254/metadata/identity/oauth2/token. This will always fail for app services as the endpoint is defined in the env var MSI_ENDPOINT  

## Steps to reproduce

Run go azure app service and call NewAuthorizerFromEnvironment() with MSI enabled

```go
	authorizer, err := kvauth.NewAuthorizerFromEnvironment() 
	if err != nil {
		return nil, fmt.Errorf("key vault authorizer error: %w", err)
	}
```

This will return the error "MSI not available" 

## Workaround

Rollback to an earlier version such as v58

PR #685, opened 2022-02-24
do not poll for availability if MSI type is not IMDS

scott-the-programmer:

# Description

This PR tries to solve https://github.com/Azure/go-autorest/issues/684 by leveraging `getMSIType()` to ensure that we're not pinging for availability for non-IMDS endpoints (as this is should instead manifest itself as an auth error)

In an app service environment, the endpoint will be set to $MSI_ENDPOINT. If $MSI_ENDPOINT does not exist, it will assume that the msi type is IMDS and behave as it did previously

## Validation in Azure

Validated in an Azure App Service by running

```go
	authorizer, err := kvauth.NewAuthorizerFromEnvironment() 
	if err != nil {
		return nil, fmt.Errorf("key vault authorizer error: %w", err)
	}
```


autorest/adal/token.go
```diff
@@ -1317,12 +1317,28 @@ func NewMultiTenantServicePrincipalTokenFromCertificate(multiTenantCfg MultiTena
 
 // MSIAvailable returns true if the MSI endpoint is available for authentication.
 func MSIAvailable(ctx context.Context, s Sender) bool {
+	_, _, err := getMSIAvailable(ctx, s)
```
chlowell, line 1320: With the caveat that I'm not familiar with the usage patterns around this function and may be missing some context, I would have this function simply return `true` when MSI_ENDPOINT is set, in order to:
1. trust the platform to configure the environment
2. surface App Service problems as authentication errors rather than swallowing them here
3. probe only the IMDS endpoint, if we probe at all, because it's the only one with no availability indicator

scott-the-programmer: Sounds good to me! I've just updated the PR now 👍🏼

chlowell (approved):

jhendrixMSFT (approved):

PR #688, opened 2022-03-17
Add support for auth with a client resource ID

plombardi89:

This PR adds support for using a Managed Identity Resource ID in the MSI auth flow. This is a supported feature of the IMDS that is not currently supported by go-autorest.

### DIFF
--- a/autorest/azure/auth/auth.go
+++ b/autorest/azure/auth/auth.go
@@ -741,9 +741,15 @@ type MSIConfig struct {
 
 // ServicePrincipalToken creates a ServicePrincipalToken from MSI.
 func (mc MSIConfig) ServicePrincipalToken() (*adal.ServicePrincipalToken, error) {
-	spToken, err := adal.NewServicePrincipalTokenFromManagedIdentity(mc.Resource, &adal.ManagedIdentityOptions{
-		ClientID: mc.ClientID,
-	})
+	mio := adal.ManagedIdentityOptions{}
+
+	if strings.Contains(mc.ClientID, "/Microsoft.ManagedIdentity/userAssignedIdentities/") {
+		mio.IdentityResourceID = mc.ClientID
+	} else {
+		mio.ClientID = mc.ClientID
+	}
+
+	spToken, err := adal.NewServicePrincipalTokenFromManagedIdentity(mc.Resource, &mio)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get oauth token from MSI: %v", err)
 	}


autorest/azure/auth/auth.go
```diff
@@ -741,9 +741,15 @@ type MSIConfig struct {
 
 // ServicePrincipalToken creates a ServicePrincipalToken from MSI.
 func (mc MSIConfig) ServicePrincipalToken() (*adal.ServicePrincipalToken, error) {
-	spToken, err := adal.NewServicePrincipalTokenFromManagedIdentity(mc.Resource, &adal.ManagedIdentityOptions{
-		ClientID: mc.ClientID,
-	})
+	mio := adal.ManagedIdentityOptions{}
+
+	if strings.Contains(mc.ClientID, "/Microsoft.ManagedIdentity/userAssignedIdentities/") {
```
jhendrixMSFT, line 746: @chlowell do we need to support this in `azidentity`?

chlowell: `azidentity` supports specifying a user-assigned identity by its resource ID but doesn't support passing a resource ID as a client ID. I don't want to support that ☺️

jhendrixMSFT: True.  I was more wondering about support client and resource IDs, then I remembered that we [do](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/managed_identity_credential.go#L21-L22) in a much cleaner way.

jhendrixMSFT: So we do support client resource IDs at the [adal](https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L766-L774) level.  Apparently we never exposed this in the `auth` package.  Do you think this is a safe workaround for that?  The other option would be to add an `IdentityResourceID` field to the `MSIConfig` type.

chlowell: I would expose the field. This approach is probably safe, but I think we should maintain the distinction between client and resource IDs. Users generally know the difference and I expect they would find it unintuitive to assign a resource ID to a `ClientID` field--that looks like a bug.

jhendrixMSFT (changes requested):
Please expose the `IdentityResourceID` field as per discussed.

In addition, you can use `azidentity` module for authentication which already supports this (and other) scenarios.

jhendrixMSFT:
Closing as stale.  If there's still interest in this, please address feedback and reopen.

Issue #687, opened 2022-03-09
Support Synapse in USGovernmentCloud


lucashby:
Please update the environments to support Synapse in USGovernmentCloud for Azure.

We are using Synapse in USGovernmentCloud and really need to be able to create resources under our Synapse resource through Terraform, but are unable to do so because the environments say it is not supported even though we are currently using it.

Issue from Terraform provider:  https://github.com/hashicorp/terraform-provider-azurerm/issues/15443

jhendrixMSFT:
Fixed in `autorest@v0.11.26`

lucashby:
> Fixed in `autorest@v0.11.26`

Can this also be fixed in the v14.x.x releases?  I don't understand the difference between the 0.x.x releases and 14.x.x releases, but I do see the downstream provider depends on 0.x.x but ends up with an indirect dependency on 14.x.x.

jhendrixMSFT:
The `v14` releases are from pre-Go modules when we used `dep` for dependency management.

The indirect dependency you're seeing is a result of "carving out" the `autorest` (and other) modules from the monolithic release.  It's not actually used for anything, so moving to `autorest@v0.11.26` should be sufficient.

Issue #585, opened 2020-12-02
Environment: MicrosoftGraphEndpoints


miosmian:
Currently the `Environment.GraphEndpoint` corresponds to `"https://graph.windows.net/",` (other cloud endpoints follow this format) but the new microsoft graph unified api endpoint is not in this struct. Are there plans to add this? It would be,

`Environment.GraphEndpoint = "https://graph.microsoft.com/",`  (public cloud)
`Environment.GraphEndpoint = "https://graph.microsoft-ppe.com/",` (ppe)
 and others for Gov..etc clouds?

jhendrixMSFT:
The current graph endpoints are for AAD graph which is different from Microsoft Graph.  We can add them, but the existing ones will remain the same.

jhendrixMSFT:
Fixed in `autorest@v0.11.26`

Issue #312, opened 2018-08-31
Environments should expose Data Lake URI's


tombuildsstuff:
👋 

It'd be really handy for the Data Lake URI to be exposed as part of the Environments struct:

https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L37-L58

An example URI for Azure Public is `adl://tomdevdls1.azuredatalakestore.net/test/example.txt`, so I'd expect the domain `azuredatalakestore.net` to be exposed as part of the Environments struct. Would it be possible to expose this information?

Thanks!

tombuildsstuff:
I've just noticed this is currently exposed on the Client within the SDK, which we should be able to use in the interim: https://github.com/Azure/azure-sdk-for-go/blob/master/services/datalake/store/2016-11-01/filesystem/client.go#L29

jhendrixMSFT:
@joshgav do you know if there are endpoints for the sovereign clouds?

marstr:
Putting this back in the triage bucket now that I've been re-orged.

jhendrixMSFT:
Fixed in `autorest@v0.11.26`

PR #694, opened 2022-04-15
Update some Azure environment endpoints and resource IDs

jhendrixMSFT:
Adapted from https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/cloud.py

Fixes
- https://github.com/Azure/go-autorest/issues/312
- https://github.com/Azure/go-autorest/issues/585
- https://github.com/Azure/go-autorest/issues/687

### DIFF
--- a/autorest/azure/environments.go
+++ b/autorest/azure/environments.go
@@ -52,6 +52,7 @@ type ResourceIdentifier struct {
 	SQLDatabase         string `json:"sqlDatabase"`
 	CosmosDB            string `json:"cosmosDB"`
 	ManagedHSM          string `json:"managedHSM"`
+	MicrosoftGraph      string `json:"microsoftGraph"`
 }
 
 // Environment represents a set of endpoints for each of Azure's Clouds.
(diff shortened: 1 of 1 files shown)


autorest/azure/environments.go
```diff
@@ -163,7 +170,8 @@ var (
 		ContainerRegistryDNSSuffix:   "azurecr.us",
 		TokenAudience:                "https://management.usgovcloudapi.net/",
 		APIManagementHostNameSuffix:  "azure-api.us",
-		SynapseEndpointSuffix:        NotAvailable,
+		SynapseEndpointSuffix:        ".dev.azuresynapse.usgovcloudapi.net",
```
chlowell, line 173: Should this have a leading . ?
```suggestion
		SynapseEndpointSuffix:        "dev.azuresynapse.usgovcloudapi.net",
```

chlowell (approved):

autorest/azure/environments.go
jhendrixMSFT: Good catch, fixed.

Issue #696, opened 2022-05-17
adal.token.EnsureFresh threw unmarshal error on `expires_on` field


northtyphoon:
The following code tried to decode an int field (expires_on) to string and threw error.

https://github.com/Azure/go-autorest/blob/49906996bd99c5b2dfa96a0ff9ed4756c649dfa6/autorest/adal/token.go#L1108

'json: cannot unmarshal number into Go struct field .expires_on of type string' JSON = '{"access_token":"","refresh_token":"","expires_in":86399,"expires_on":1652886678,"not_before":1652799978,"resource":"[https://management.azure.com/","token_type":"Bearer"}'](https://management.azure.com/%22,%22token_type%22:%22Bearer%22%7D%27)

jhendrixMSFT:
Fixed in `adal@v0.9.20`

PR #698, opened 2022-05-19
Handle expires_on in int format

jhendrixMSFT:
Unmarshal the value into an interface{} and perform the proper
conversion depending on the underlying type.

### DIFF
--- a/autorest/adal/token_test.go
+++ b/autorest/adal/token_test.go
@@ -89,7 +89,7 @@ func TestTokenWillExpireIn(t *testing.T) {
 
 func TestParseExpiresOn(t *testing.T) {
 	// get current time, round to nearest second, and add one hour
-	n := time.Now().UTC().Round(time.Second).Add(time.Hour)
+	n := time.Now().UTC()
 	amPM := "AM"
 	if n.Hour() >= 12 {
 		amPM = "PM"
@@ -107,12 +107,12 @@ func TestParseExpiresOn(t *testing.T) {
 		{
 			Name:   "timestamp with AM/PM",
 			String: fmt.Sprintf("%d/%d/%d %d:%02d:%02d %s +00:00", n.Month(), n.Day(), n.Year(), n.Hour(), n.Minute(), n.Second(), amPM),
-			Value:  3600,
+			Value:  n.Unix(),
 		},
 		{
 			Name:   "timestamp without AM/PM",
 			String: fmt.Sprintf("%d/%d/%d %d:%02d:%02d +00:00", n.Month(), n.Day(), n.Year(), n.Hour(), n.Minute(), n.Second()),
-			Value:  3600,
+			Value:  n.Unix(),
 		},
 	}
 	for _, tc := range testcases {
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/696

autorest/adal/token_test.go
```diff
@@ -1461,6 +1489,23 @@ func newTokenJSON(expiresIn, expiresOn, resource string) string {
 		expiresIn, expiresOn, nb, resource)
 }
 
+func newTokenJSONIntExpiresOn(expiresIn string, expiresOn int, resource string) string {
+	/*nb, err := parseExpiresOn(expiresOn)
```
chlowell, line 1493: Got some cruft here

chlowell (dismissed):

jhendrixMSFT:
Looking at the related issue, it looks to me that `expires_on`, at least in that example, isn't the number of seconds from now but probably from the Unix epoch.  I need to take a closer look.

chlowell:
Everywhere except App Service, `expires_on` is epoch seconds, either as a number or a string.

jhendrixMSFT:
And App Service is as a time-stamp correct?

jhendrixMSFT:
OK I did a little digging.  [`Token.Expires()`](https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L158-L167) already treats `ExpiresOn` as Unix time.  It does mean though that our handling of `expires_on` in date-time format is incorrect at present.

autorest/adal/token_test.go
```diff
@@ -107,12 +107,12 @@ func TestParseExpiresOn(t *testing.T) {
 		{
 			Name:   "timestamp with AM/PM",
 			String: fmt.Sprintf("%d/%d/%d %d:%02d:%02d %s +00:00", n.Month(), n.Day(), n.Year(), n.Hour(), n.Minute(), n.Second(), amPM),
-			Value:  3600,
+			Value:  n.Unix(),
 		},
 		{
 			Name:   "timestamp without AM/PM",
 			String: fmt.Sprintf("%d/%d/%d %d:%02d:%02d +00:00", n.Month(), n.Day(), n.Year(), n.Hour(), n.Minute(), n.Second()),
```
chlowell, line 114: I don't know whether it matters here but the formats also differ in padding 🙄 
```suggestion
			String: fmt.Sprintf("%02d/%02d/%02d %02d:%02d:%02d +00:00", n.Month(), n.Day(), n.Year(), n.Hour(), n.Minute(), n.Second()),
```

autorest/adal/token_test.go
```diff
@@ -89,7 +89,7 @@ func TestTokenWillExpireIn(t *testing.T) {
 
 func TestParseExpiresOn(t *testing.T) {
 	// get current time, round to nearest second, and add one hour
-	n := time.Now().UTC().Round(time.Second).Add(time.Hour)
+	n := time.Now().UTC()
```
chlowell, line 92: Comment is incorrect now

chlowell (approved):

autorest/adal/token_test.go
jhendrixMSFT: Good catch.  Looks like the result is the same but I changed the padding to match.

autorest/adal/token_test.go
jhendrixMSFT: Removed

PR #704, opened 2022-06-24
Fix validation in resource id parsing

karataliu:

Before fix, 'ANYSTRING+valid id' can also be parsed as valid id, that's not correct.


karataliu:
@jhendrixMSFT 
That's in description, without the fix:
`ABCDEF/subscription/xxx/resourcegroup/xxx/...` is considered a valid id, but actually not.

jhendrixMSFT (approved):

PR #705, opened 2022-06-30
LRO update status per `status` in response also applies to Location Poll

magodo:
This PR extends the LRO pt status update based on the response `status` filed to location polling type.

This is to fix issues like: https://github.com/Azure/azure-rest-api-specs/issues/18844

### DIFF
--- a/autorest/azure/async.go
+++ b/autorest/azure/async.go
@@ -561,7 +561,7 @@ Default:
 }
 
 func (pt *pollingTrackerBase) updatePollingState(provStateApl bool) error {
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Per the [ARM RPC spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#azure-asyncoperation-resource-format), `Location` header polling doesn't use `status`.  Is the service returning the wrong LRO polling header?

jhendrixMSFT:
Closing as it looks like the service team needs to make changes on their side.

PR #706, opened 2022-07-19
set MaxIdleConnsPerHost to 100 to reuse connections when there is large volume of requests

MartinForReal:

### DIFF
--- a/autorest/sender.go
+++ b/autorest/sender.go
@@ -139,6 +139,7 @@ func sender(renengotiation tls.RenegotiationSupport) Sender {
 			}).DialContext,
 			ForceAttemptHTTP2:     true,
 			MaxIdleConns:          100,
+			MaxIdleConnsPerHost:   100,
 			IdleConnTimeout:       90 * time.Second,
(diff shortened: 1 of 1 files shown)


MartinForReal:
@jhendrixMSFT  Could you please review the code? Thanks!

jhendrixMSFT (changes requested):
This is a significant increase from the default value of 2.  Is there any reason why you can't use a custom HTTP client, configured as required, and assign it to [Client.Sender](https://github.com/Azure/go-autorest/blob/main/autorest/client.go#L158)?

MartinForReal:
related: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/2059

MartinForReal:
I agree. But most of the connections are sent to the same endpoint. And when there are a lot of clients reusing the same transport, the transport will only keep 2 idle connections for this endpoint. It is insufficient when compared with 100 idle connections available.

jhendrixMSFT:
Looks like this was fixed on the kubernetes side so I'm closing this.  Please ping back if you disagree.

PR #715, opened 2022-12-08
Update environments.go

codecae:
While researching an issue for an application that uses autorest, we stumbled upon an incorrect constant defined in environments.go.

We found that one graph endpoint URL is defined as `.net` and another graph endpoint url in the same object is defined as `.us`.

These values should both reflect `.us` endpoints to ensure go applications implementing go-autorest do not run into endpoint issues.


jhendrixMSFT:
The entries we have are for AAD graph.  I believe that `graph.windows.us` is for Microsoft Graph which this SDK doesn't support.  Please reopen if you disagree.

There is a separate MSGraph SDK [here](https://github.com/microsoftgraph/msgraph-sdk-go) if that's what you need.

PR #719, opened 2023-02-26
fix: enable update jwt via callback for workloadidentity

cvvz:

---
### Why make this change

When Using federated workload identity, we need to request Oauth bearer token by JWT provided by external platform(e.g. k8s) at first. Related code snippets: 
https://github.com/Azure/go-autorest/blob/main/autorest/authorization.go#L120-L125
https://github.com/Azure/go-autorest/blob/main/autorest/adal/token.go#L1045-L1062
https://github.com/Azure/go-autorest/blob/main/autorest/adal/token.go#L375-L379

When this request happens after the external JWT expired, we will get 401 error: `adal: Refresh request failed. Status Code = '401'. Response body: {"error":"invalid_client","error_description":"AADSTS700024: Client assertion is not within its valid time range. `

We should provide a callback to get the JWT dynamically, instead of providing it statically.


cvvz:
@JustinJudd Could you also help to take a look? Since it's you that bring workload identity to adal ^^

cvvz:
@jhendrixMSFT Could you please kindly take a look?

autorest/adal/token.go
```diff
@@ -686,8 +693,8 @@ func NewServicePrincipalTokenFromAuthorizationCode(oauthConfig OAuthConfig, clie
 	)
 }
 
-// NewServicePrincipalTokenFromFederatedToken creates a ServicePrincipalToken from the supplied federated OIDC JWT.
-func NewServicePrincipalTokenFromFederatedToken(oauthConfig OAuthConfig, clientID string, jwt string, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
+// NewServicePrincipalTokenFromFederatedToken creates a ServicePrincipalToken from the supplied federated OIDC JWTCallback.
+func NewServicePrincipalTokenFromFederatedToken(oauthConfig OAuthConfig, clientID string, jwtCallback JWTCallback, resource string, callbacks ...TokenRefreshCallback) (*ServicePrincipalToken, error) {
```
jhendrixMSFT, line 697: This is a breaking change.

jhendrixMSFT (changes requested):
We cannot accept breaking changes in `go-autorest` so this will have to be refactored.

In addition, the modules in this repo are reaching end-of-life.  Our new authentication module [`azidentity`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) is the path forward.  You can find the migration guide [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/MIGRATION.md).

cvvz:
> We cannot accept breaking changes in go-autorest so this will have to be refactored.

Sure I can make it without breaking changes. However, the reason why I didn't do that is that I think users will encounter building error once they upgrade adal to the new version, which will enforce them to fix this bug. If there is no breaking changes, then users may not fix it actively until they get into real trouble.

jhendrixMSFT:
The real fix is to use `azidentity`.  Have you looked at that?  We're not really taking features in `go-autorest` anymore, only critical bug fixes.

cvvz:
similar fix in msal-for-go: https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/319

autorest/adal/token.go
andyzhangx: I doubt whether there is any user would call this `NewServicePrincipalTokenFromFederatedToken` since the token would expire after 24 hours, that's more like a bug

cvvz: This bug actually only happens in such scenario:

For applications running on AKS, they need to use the JWT, which is generated by apiserver and mount to the Pod by kubelet , to exchange OAuth token from Azure identity platform when there is any request to arm **at the first time**. Azure will not only return OAuth token for authenticating, it will also return a `refresh_token` for the application to refresh Oauth token. ADAL will use the JWT to request Oauth token and `refresh_token` when there is a request to arm at the first time, and then it will use `refresh_token` to refresh the OAuth token if it is expired and the JWT should be useless.

However, application may have no request to arm for a very long time after it is up (or after master-slave switch). If the JWT is expired, apiserver will refresh the JWT and write to file mounted in the Pod. Under current implementation, the jwt we used to exchange token from Azure identity platform cannot be refreshed since we cannot read the JWT file from the file dynamically. So after the JWT is expired, request to exchange token would be rejected.

jhendrixMSFT (approved):

Issue #722, opened 2023-04-26
Remove unspecified behavior in return statement


go101:
https://github.com/Azure/go-autorest/blob/main/autorest/utility.go#L63-L66

For more details, please read https://github.com/golang/go/issues/58233#issuecomment-1499471060

jhendrixMSFT:
Fixed in `autorest/v0.11.29`, thanks for the bug report.

PR #723, opened 2023-04-27
Remove undefined behavior during return

jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/722

### DIFF
--- a/autorest/utility.go
+++ b/autorest/utility.go
@@ -60,9 +60,9 @@ func NewDecoder(encodedAs EncodedAs, r io.Reader) Decoder {
 // is especially useful if there is a chance the data will fail to decode.
 // encodedAs specifies the expected encoding, r provides the io.Reader to the data, and v
 // is the decoding destination.
-func CopyAndDecode(encodedAs EncodedAs, r io.Reader, v interface{}) (bytes.Buffer, error) {
-	b := bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


autorest/utility.go
```diff
@@ -60,9 +60,9 @@ func NewDecoder(encodedAs EncodedAs, r io.Reader) Decoder {
 // is especially useful if there is a chance the data will fail to decode.
 // encodedAs specifies the expected encoding, r provides the io.Reader to the data, and v
 // is the decoding destination.
-func CopyAndDecode(encodedAs EncodedAs, r io.Reader, v interface{}) (bytes.Buffer, error) {
-	b := bytes.Buffer{}
-	return b, NewDecoder(encodedAs, io.TeeReader(r, &b)).Decode(v)
+func CopyAndDecode(encodedAs EncodedAs, r io.Reader, v interface{}) (b bytes.Buffer, err error) {
+	err = NewDecoder(encodedAs, io.TeeReader(r, &b)).Decode(v)
```
RickWinter, line 64: Does b get auto initialized to an empty buffer before it gets dereferenced?

RickWinter (approved):

autorest/utility.go
jhendrixMSFT: Yes.  I'm not usually a fan of [named return values](https://go.dev/tour/basics/7) but in very small methods like this one it can simplify the code.

